PR #32, opened 2025-09-08
Add limit on number of episodes returned by DeclarativeMemory

edwinyyyu:
### Purpose of the change

This change replaces request for number of episodes to a strict limit on number of episodes returned `DeclarativeMemory`.

Previously, when a caller searches the `DeclarativeMemory`, the caller is required to do any strict limiting themself. Because the episodes are returned in a list sorted by timestamp, the caller is not aware of the ranking of the episodes with respect to the relevance to the query. Thus, any limiting heuristic could unfortunately discard highly relevant search results.

By implementing a limit at a lower level, callers need not be aware of the internals of how the relevant search results are obtained.

### Description

This change adds a `num_episodes_limit` argument to `DeclarativeMemory.search()`, so that callers will not need to do limiting themselves.

`num_episodes_limit` replaces `num_episodes_request`. The `num_episodes_request` parameter is removed to simplify the interface.

`LongTermMemory` is updated accordingly to use the new argument rather than a keep-most-recent heuristic.

### Type of change

-   [x] Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?

Ran LoCoMo benchmark on episodic memory, resulting in improved score by LLM judge.

### Checklist:

-   [x] I have made corresponding changes to the documentation

### Maintainer Checklist

### DIFF
--- a/src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
+++ b/src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
@@ -481,7 +481,7 @@ class DeclarativeMemory:
     async def search(
         self,
         query: str,
-        num_episodes_request: int = 20,
+        num_episodes_limit: int = 20,
         isolation_properties: dict[str, IsolationPropertyValue] = {},
     ) -> list[Episode]:
         """
@@ -490,8 +490,8 @@ class DeclarativeMemory:
         Args:
             query (str):
                 The search query.
-            num_episodes_request (int, optional):
-                The requested minimum number
+            num_episodes_limit (int, optional):
+                The maximum number
                 of episodes to return (default: 20).
             isolation_properties (
                 dict[str, IsolationPropertyValue], optional
@@ -623,18 +623,30 @@ class DeclarativeMemory:
         )
 
         # Rerank contexts.
-        reranked_episode_node_contexts = (
-            await self._rerank_episode_node_contexts(
-                query, episode_node_contexts
-            )
+        episode_node_context_scores = await self._score_episode_node_contexts(
+            query, episode_node_contexts
         )
 
+        reranked_anchored_episode_node_contexts = [
+            (nuclear_episode_node, episode_node_context)
+            for _, nuclear_episode_node, episode_node_context in sorted(
+                zip(
+                    episode_node_context_scores,
+                    nuclear_episode_nodes,
+                    episode_node_contexts,
+                ),
+                key=lambda pair: pair[0],
+                reverse=True,
+            )
+        ]
+
         # Unify contexts.
-        unified_episode_node_context: set[Node] = set()
-        for episode_node_context in reranked_episode_node_contexts:
-            if len(unified_episode_node_context) >= num_episodes_request:
-                break
-            unified_episode_node_context.update(episode_node_context)
+        unified_episode_node_context = (
+            DeclarativeMemory._unify_anchored_episode_node_contexts(
+                reranked_anchored_episode_node_contexts,
+                num_episodes_limit=num_episodes_limit,
+            )
+        )
 
         # Return episodes sorted by timestamp.
         episodes = [
@@ -711,11 +723,11 @@ class DeclarativeMemory:
 
         return retrieved_context
 
-    async def _rerank_episode_node_contexts(
+    async def _score_episode_node_contexts(
         self, query: str, episode_node_contexts: list[set[Node]]
-    ) -> list[set[Node]]:
+    ) -> list[float]:
         """
-        Rerank episode node contexts
+        Score episode node contexts
         based on their relevance to the query.
         """
         episode_node_context_contents = [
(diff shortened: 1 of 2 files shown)


src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
```diff
@@ -482,6 +482,7 @@ async def search(
         self,
         query: str,
         num_episodes_request: int = 20,
+        num_episodes_limit: int | None = None,
```
malatewang: Why do the extra parameter is required? The num_episodes_request is supposed to be the limit.

edwinyyyu: Context is higher-quality if there is no limit.

With `num_episodes_request`, `DeclarativeMemory.search()` tries to get the least number of episode contexts that provides at least that many episodes, and ensures that there is surrounding context provided for every episode i.e. It will not return an isolated episode with no context, which can confuse the model. A contrived example that can confuse the model would be:

```
A: Do you like pizza? (episode context 1)
B: No. (episode context 1)
B: Yes. (episode context 2) -- the limit has caused the question to be discarded
```

With a limit, the surrounding context of an episode can be discarded.

src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
```diff
@@ -753,6 +761,27 @@ async def _rerank_episode_node_contexts(
...
+        for episode_node_context in episode_node_contexts:
+            if len(unified_episode_node_context) >= num_episodes_request:
+                return unified_episode_node_context
+            for episode_node in episode_node_context:
+                if (
+                    num_episodes_limit is not None
+                    and len(unified_episode_node_context) >= num_episodes_limit
+                ):
+                    return unified_episode_node_context
+                unified_episode_node_context.add(episode_node)
```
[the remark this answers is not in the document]
edwinyyyu: This won't work if the limit is 0 or negative.

src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
```diff
@@ -753,6 +761,27 @@ async def _rerank_episode_node_contexts(
...
+        unified_episode_node_context: set[Node] = set()
+
+        for episode_node_context in episode_node_contexts:
+            if len(unified_episode_node_context) >= num_episodes_request:
+                return unified_episode_node_context
+            for episode_node in episode_node_context:
+                if (
+                    num_episodes_limit is not None
+                    and len(unified_episode_node_context) >= num_episodes_limit
+                ):
```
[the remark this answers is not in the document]
edwinyyyu: The request is not meant to be a limit. We want to return the fewest number of episode ndoes such that a whole number of contexts is provided, unless the limit forces us not to provide a whole context.

src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
edwinyyyu: On second thought it may be okay to add an extra episode or two even without context. I will just have a limit then.

edwinyyyu:
> Need to understand why the extra parameter is necessary

I have removed the `num_episodes_request` parameter in favor of a single `num_episodes_limit` parameter.

src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
```diff
@@ -736,22 +748,50 @@ async def _rerank_episode_node_contexts(
...
+                    node.properties.get("timestamp", datetime.min),
                 ),
-                key=lambda pair: pair[0],
-                reverse=True,
             )
-        ]
-        return reranked_episode_node_contexts
+
+            # Order context episodes by proximity to nucleus.
+            nucleus_index = chronological_context.index(nucleus)
```
malatewang: Sort is relatively expensive. This sort can be handled by a simpler method by just manipulating the index.

src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
```diff
@@ -736,22 +748,50 @@ async def _rerank_episode_node_contexts(
...
+            nuclear_context = sorted(
+                chronological_context,
+                key=lambda node: abs(
+                    chronological_context.index(node) - nucleus_index
+                ),
+            )
+
+            # Add episodes to unified context until limit is reached.
+            for episode_node in nuclear_context:
```
malatewang: Given the neighbors of nucleus are retrieved, it is possible that there could be overlap between the "context". It is better to check and remove the duplications here.

edwinyyyu: Duplicates are handled by the set addition.

src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
edwinyyyu: I do not expect there to be millions of episodes sorted.

Because we are using a set here, the logic to add a desired number of episodes would be more complex than this, and I think it would be less readable, which I think is the bigger concern as compared to performance, at least at present.

I will add some more comments with reasoning for this.

edwinyyyu:
I have added a check that ensures the sort is only performed if there is a possibility that adding the context exceeds the limit.

This reduces the cost when the number of episodes is large. I have kept the second sort which I think is more readable than doing some index magic. Let me know if I should change it to be more efficient.

malatewang (approved):

sscargal (approved):
LGTM

PR #36, opened 2025-09-09
Add MCP server

malatewang:
### Purpose of the change

Add MCP server endpoints

### Description

This PR adds MCP server implementation. The MCP server supports both stateless and stateful memory access APIs.

Fixes # (issue)

### Type of change

### How Has This Been Tested?

-   [x] Integration Test

### Checklist:

-   [x] Any dependent changes have been merged and published in downstream modules

### Maintainer Checklist

### Screenshots/Gifs

[If applicable, add screenshots or GIFs that show the changes in action. This is especially helpful for API responses.]

### Further comments

[Add any other relevant information here, such as potential side effects, future considerations, or any specific questions for the reviewer.]

### DIFF
--- a/src/server/app.py
+++ b/src/server/app.py
@@ -55,23 +55,6 @@ class SessionData(BaseModel):
     session_id: str
 
 
-class EpisodeData(BaseModel):
-    """Request model for episode data to create new memories."""
-    producer: str
-    produced_for: str
-    episode_content: str | list[float]
-    episode_type: str
-    metadata: dict[str, Any] | None
-
-
-class SearchRequest(BaseModel):
-    """Request model for searching memories by MCP."""
-    query: str
-    user_id: str | None = None
-    filter: dict[str, Any] | None = None
-    limit: int | None = None
-
-
 # === Request Models ===
 class NewEpisode(BaseModel):
     """Request model for adding a new memory episode."""
@@ -249,97 +232,6 @@ app.add_route("/metrics", make_asgi_app())
 
 
 @mcp.tool()
-async def mcp_open_memory_store(sess: SessionData, ctx: Context):
-    """MCP tool to create and open a memory store.
-
-    This tool establishes a session context for subsequent MCP tool calls.
-    The session data is stored in the MCP context state.
-
-    Args:
-        sess: The session data to establish the context.
-        ctx: The MCP context.
-
-    Returns:
-        No return value.
-    """
-    old_sess = ctx.get_state("session_data")
-    if old_sess is not None:
-        await episodic_memory.close_episodic_memory_instance(
-            group_id=old_sess.group_id,
-            agent_id=old_sess.agent_id,
-            user_id=old_sess.user_id,
-            session_id=old_sess.session_id,
-        )
-    ctx.set_state("session_data", sess)
-
-
-@mcp.tool()
-async def mcp_close_memory_store(ctx: Context):
-    """MCP tool to close the currently open memory store.
-
-    This tool retrieves the session data from the MCP context, closes the
-    corresponding episodic memory instance, and clears the session from the
-    context.
-
-    Args:
-        ctx: The MCP context.
-
-    Returns:
-        No return value.
-    """
-    sess = ctx.get_state("session_data")
-    if sess is None:
-        return
-    await episodic_memory.close_episodic_memory_instance(
-        group_id=sess.group_id,
-        agent_id=sess.agent_id,
-        user_id=sess.user_id,
-        session_id=sess.session_id,
-    )
-    ctx.set_state("session_data", None)
-
-
-@mcp.tool()
-async def mcp_add_memory(
-    episode: EpisodeData,
-    ctx: Context
-) -> dict[str, Any]:
-    """MCP tool to add a memory episode to the current session.
-
-    This tool requires an open memory session. It adds a new memory episode
-    to the episodic and profile memories associated with the current session.
-
-    Args:
-        episode: The episode data to be added.
-        ctx: The MCP context.
-
-    Returns:
-        Status 0 if the memory was added successfully, -1 otherwise with
-        error message.
-    """
-    sess = ctx.get_state("session_data")
-    if sess is None:
-        return {"status": -1, "error_msg": "No session open"}
-    data = NewEpisode(
-        session=sess,
-        producer=episode.producer,
-        produced_for=episode.produced_for,
-        episode_content=episode.episode_content,
-        episode_type=episode.episode_type,
-        metadata=episode.metadata,
-    )
-    try:
-        await add_memory(data)
-    except HTTPException as e:
-        session_name = f"""{sess.group_id}-{sess.agent_id}-
-                           {sess.user_id}-{sess.session_id}"""
-        logger.error("Failed to add memory episode for %s", session_name)
-        logger.error(e)
-        return {"status": -1, "error_msg": str(e)}
-    return {"status": 0, "error_msg": ""}
-
-
-@mcp.tool()
 async def mcp_add_session_memory(episode: NewEpisode) -> dict[str, Any]:
     """MCP tool to add a memory episode for a specific session.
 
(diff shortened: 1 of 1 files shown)


src/server/app.py
```diff
@@ -184,10 +221,245 @@ async def http_app_lifespan(application: FastAPI):
...
+        return False
+    data = NewEpisode(
+        session=sess,
+        producer=episode.producer,
+        produced_for=episode.produced_for,
+        episode_content=episode.episode_content,
+        episode_type=episode.episode_type,
+        metadata=episode.metadata,
+    )
+    try:
```
sscargal: Rather than the tools returning only TRUE or FALSE, we can improve error handling for MCP clients by returning an object. This allows the clients to handle errors more gracefully.

For example:

```python
from typing import Dict, Any

@ mcp.tool()
async def mcp_add_memory(episode: EpisodeData, ctx: Context) -> Dict[str, Any]:
    try:
        await add_memory(data)
        return {"success": True}
    except HTTPException as e:
        logger.error(e)
        return {
            "success": False,
            "error": str(e.detail) if hasattr(e, "detail") else str(e),
            "code": e.status_code if hasattr(e, "status_code") else None,
            "details": {
                "episode": episode.dict(),
                "exception_type": type(e).__name__,
            }
        }
 ```

We can use the same pattern in other tools, where applicable.

File: src/server/app.py [hunk 2]
```diff
@@ -106,6 +127,37 @@ class DeleteDataRequest(BaseModel):
...
+
+def get_db_config(yaml_config) -> DBConfig:
+    """Helper function to retrieve database configuration."""
+    db_host = os.getenv("POSTGRES_HOST")
+    db_port = os.getenv("POSTGRES_PORT")
+    db_user = os.getenv("POSTGRES_USER")
+    db_pass = os.getenv("POSTGRES_PASS")
+    db_name = os.getenv("POSTGRES_DB")
+    # get DB config from configuration file is available
+    profile_config = yaml_config.get("profile_memory", {})
```
sscargal, line 133: For security purposes, we should move credentials to environment variables or a secrets manager only and avoid plain text YAML unless absolutely necessary. Storing usernames, passwords, and database information in plain text is a security concern.

src/server/app.py
```diff
@@ -184,10 +221,245 @@ async def http_app_lifespan(application: FastAPI):
...
+        producer=episode.producer,
+        produced_for=episode.produced_for,
+        episode_content=episode.episode_content,
+        episode_type=episode.episode_type,
+        metadata=episode.metadata,
+    )
+    try:
+        await add_memory(data)
+    except HTTPException as e:
+        logger.error(e)
```
sscargal: [Optional Enhancement] It may be helpful to include additional information, such as session details, in the log event to provide more context. Don't log sensitive information such as user passwords.

src/server/app.py
```diff
@@ -184,10 +221,245 @@ async def http_app_lifespan(application: FastAPI):
...
+
+    Returns:
+        True if the session was successfully opened.
+    """
+    ctx.set_state("session_data", sess)
+    return True
+
+
+@mcp.tool()
+async def mcp_close_memory(ctx: Context):
```
edwinyyyu: Will the server leak resources if the client never calls this?

malatewang: > Will the server leak resources if the client never calls this?

The episodic memory manager will keep the instance open until the server shutdown if the client does not call this. It will just waste some memory. The problem is the fastmcp does not provide API to monitor the session state change so there is no easy way to do automatic cleanup.

src/server/app.py
```diff
@@ -184,10 +221,245 @@ async def http_app_lifespan(application: FastAPI):
...
+        async with mcp_app.lifespan(application):
+            yield
+
+app = FastAPI(lifespan=mcp_http_lifespan)
+app.mount("/mcp", mcp_app)
 app.add_route("/metrics", make_asgi_app())
 
 
+@mcp.tool()
+async def mcp_open_memory(sess: SessionData, ctx: Context):
```
edwinyyyu: I feel like the term "memory" is overloaded here. There is a difference between the opened/closed memory and the added memory (episode).

Also, I believe all of these tools will be named prefixed with "mcp", which is somewhat awkward outside of the Python file. Could all of these functions be organized into another file or data structure, so that the prefix can be removed?

edwinyyyu (changes requested):
Please tick applicable boxes in PR description.
like: [x]

src/server/app.py
malatewang: > For security purposes, we should move credentials to environment variables or a secrets manager only and avoid plain text YAML unless absolutely necessary. Storing usernames, passwords, and database information in plain text is a security concern.

The credentials are retrieved from the environmental variables. It does allow them to be stored in yaml file.

src/server/app.py
malatewang: Good suggestion. I changed to return status and error message in json format.

src/server/app.py
malatewang: Added to the log

src/server/app.py
malatewang: I renamed the memory to memory store. I agree that it is better to separate the MCP and http. But it requires much more efforts now. I prefer to postpone it.

edwinyyyu:
Are we allowing both merge and rebase?

Applicable if rebasing:
I would prefer if the commits are squashed into a single "Add MCP server" commit or cleaned up. In particular, the "Apply limit . . . " and "Remove files . . . " commits can be dropped.

I also think it may be helpful if the commit messages provided more context so that it is easier to tell at a glance what the commit is about without looking into this PR.

src/server/app.py
edwinyyyu: How much downside is there to closing the memory at the end of every request? If anything is RAM-only, is there much cost to persisting on disk?

I think this behavior needs to change before it is ready for a real use case.

edwinyyyu: That said, I can approve as urgent.

EDIT: I'm checking out this PR to use it currently. No need to merge immediately.

src/server/app.py
```diff
@@ -49,6 +55,23 @@ class SessionData(BaseModel):
...
+class SearchRequest(BaseModel):
+    """Request model for searching memories by MCP."""
+    query: str
+    user_id: str | None = None
+    filter: dict[str, Any] | None = None
+    limit: int | None = None
+
+
 # === Request Models ===
 class NewEpisode(BaseModel):
```
edwinyyyu: Fields should not be duplicated between NewEpisode and EpisodeData.

src/server/app.py
```diff
@@ -184,10 +223,290 @@ async def http_app_lifespan(application: FastAPI):
...
+    Args:
+        sess: The session data to establish the context.
+        ctx: The MCP context.
+
+    Returns:
+        No return value.
+    """
+    old_sess = ctx.get_state("session_data")
+    if old_sess is not None:
+        await episodic_memory.close_episodic_memory_instance(
```
edwinyyyu: Why does mcp_open_memory_store call close?

malatewang: If the client open a new episodic without closing existing one, we need to close the opened memory store

malatewang:
> Are we allowing both merge and rebase?
> 
> Applicable if rebasing: I would prefer if the commits are squashed into a single "Add MCP server" commit or cleaned up. In particular, the "Apply limit . . . " and "Remove files . . . " commits can be dropped.
> 
> I also think it may be helpful if the commit messages provided more context so that it is easier to tell at a glance what the commit is about without looking into this PR.

I think usually, we keep all the histories.

src/server/app.py
malatewang: Yes, we could use EpisodeData in the NewEpisode. But it means the http API change. Probably we do not touch API change now.

src/server/app.py
```diff
@@ -184,10 +223,290 @@ async def http_app_lifespan(application: FastAPI):
...
+            group_id=old_sess.group_id,
+            agent_id=old_sess.agent_id,
+            user_id=old_sess.user_id,
+            session_id=old_sess.session_id,
+        )
+    ctx.set_state("session_data", sess)
```
edwinyyyu: You never open a new memory instance.

edwinyyyu:
I think this design doesn't work when there are multiple concurrent users.
One client can close another client's memory.

src/server/app.py
malatewang: If an instance is actually opened and session is broken, there is no way to cleanup.

malatewang:
> I think this design doesn't work when there are multiple concurrent users. One client can close another client's memory.

The instance is reference counted.

sscargal:
> I think this design doesn't work when there are multiple concurrent users. One client can close another client's memory.

Initially, I thought that too. However, I believe FastAPI creates a new `ctx` for each user/connection, without explicitly requiring the creation of a new context in this code. It needs to be verified, though.

edwinyyyu:
> > I think this design doesn't work when there are multiple concurrent users. One client can close another client's memory.
> 
> Initially, I thought that too. However, I believe FastAPI creates a new `ctx` for each user/connection, without explicitly requiring the creation of a new context in this code. It needs to be verified, though.

On the server-side, they are all using the same memory manager so they share the same set of memory instances.

edwinyyyu:
> > I think this design doesn't work when there are multiple concurrent users. One client can close another client's memory.
> 
> The instance is reference counted.

In another comment:
> If an instance is actually opened and session is broken, there is no way to cleanup.

Possible issues:
- If a client calls close multiple times, it can close another client's memory.
- If a client opens, does nothing, then closes, it can close another client's memory (because open doesn't actually open).

PR #37, opened 2025-09-10
Format episodes before sending to reranker in DeclarativeMemory

edwinyyyu:
### Purpose of the change

This change allows the user of DeclarativeMemory to specify a template to populate with episode metadata, so that metadata may be present in the episode contents provided to the reranker.

If using LongTermMemory or another above-layer, then this change is transparent to the user. In this case, the DeclarativeMemory shares the template with the metadata derivative deriver.

This improves returned episode quality by providing more context to the reranker.

### Description
This change adds a new config key `episode_metadata_template` to DeclarativeMemory, which accepts a template string supporting $-substitutions.

Supported $-substitution keys include the Episode metadata, isolation properties, and user metadata. User metadata takes priority over isolation properties, which takes priority over base Episode metadata, in the case of naming conflicts, as works for the metadata derivative mutator.

### Type of change

### How Has This Been Tested?
Ran LoCoMo benchmark on episodic memory, resulting in improved score by LLM judge. Seems to be 1-2% improvement from a single full run.

### Checklist:

### Maintainer Checklist

### DIFF
--- a/src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
+++ b/src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
@@ -8,6 +8,7 @@ import functools
 import json
 from collections.abc import Awaitable, Callable
 from datetime import datetime
+from string import Template
 from typing import Any, Self, cast
 from uuid import uuid4
 
@@ -74,6 +75,9 @@ class DeclarativeMemory:
                             - derivative_mutator:
                               DerivativeMutator instance
                               used for mutating derivatives.
+                - episode_metadata_template:
+                    Template string supporting $-substitutions
+                    (default: "[$timestamp] $content").
         """
 
         self._vector_graph_store: VectorGraphStore = config[
@@ -148,6 +152,10 @@ class DeclarativeMemory:
                 "derivation_workflows"
             ].items()
         }
+
+        self._episode_metadata_template = Template(
+            config.get("episode_metadata_template", "[$timestamp] $content")
+        )
 
     class Workflow:
         def __init__(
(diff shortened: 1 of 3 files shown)


edwinyyyu:
I would like feedback on the name conflict resolution. I chose this solution because the user has to learn the least about how to use the system.

One potential problem if the user uses a conflicting name is that when the user metadata are populated for some Episodes and are not populated for other Episodes, the internal metadata may leak into the reranking.

A similar issue may occur with the metadata derivative formatter, where the internal metadata may leak into the embedding.

The alternatives I have come up with are:
- to reserve certain metadata keys
  - we can keep the current implementation, except maybe with some checks
    - maybe harder to add new metadata keys later
  - we can have all of our metadata keys to use a prefix or something
- to mangle the user metadata (and require the user to do mangling or bracket/dot-notation in the template)

src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
```diff
@@ -730,26 +718,53 @@ async def _score_episode_node_contexts(
...
+            DeclarativeMemory._episodes_from_episode_nodes(
+                list(episode_node_context)
+            )
+            for episode_node_context in episode_node_contexts
+        ]
+
+        def get_formatted_episode_content(episode: Episode) -> str:
+            # Format episode content for reranker using metadata.
+            return self._episode_metadata_template.safe_substitute(
+                {
```
malatewang, line 731: I think most of the Episode members should not be included in the content: episode type, isolation_properties, content_type

edwinyyyu: The config determines what is used. I agree that in a typical conversational use case that several of these metadata fields are not useful.

Allowing it to be configurable reduces the amount of logic and documentation required.

This is also adaptable. Suppose a future or reranker is smart enough to make use of the episode type or isolation properties.

edwinyyyu:
> Some unnecessary data is used in rerank.

The config determines what is used.

Issue #100, opened 2025-09-19
[Bug]: memmachine-compose.sh on Mac OS X (15.6.1) returns "timeout: command not found"


sscargal:
### Describe the bug

When executing `memmachine-compose.sh` on Mac OS X, we see an error "./memmachine-compose.sh: line 160: timeout: command not found".

Here is a snippet of the terminal output

```
[SUCCESS] Services started successfully!
[INFO] Waiting for services to be healthy...
NAME                  IMAGE                    COMMAND                  SERVICE             CREATED             STATUS                                     PORTS
memmachine-app        memmachine/memmachine    "sh -c 'memmachine-s…"   memmachine          33 seconds ago      Up Less than a second (health: starting)   0.0.0.0:8080->8080/tcp
memmachine-neo4j      neo4j:5.23-community     "tini -g -- /startup…"   neo4j               33 seconds ago      Up 32 seconds (healthy)                    0.0.0.0:7473-7474->7473-7474/tcp, 0.0.0.0:7687->7687/tcp
memmachine-postgres   pgvector/pgvector:pg16   "docker-entrypoint.s…"   postgres            33 seconds ago      Up 32 seconds (healthy)                    0.0.0.0:5432->5432/tcp
[INFO] Checking service health...
[INFO] Waiting for PostgreSQL to be ready...
./memmachine-compose.sh: line 160: timeout: command not found
[ERROR] PostgreSQL failed to become ready in 120 seconds. Check container logs and configuration.
```

### Steps to reproduce

Run `memmachine-compose.sh` on Mac OS X with a service that doesn't exist or fails.

The cause is that the `timeout` command isn't available on Mac OS X

Ubuntu Linux 24.04: 
```
$ which timeout
/usr/bin/timeout
```

Mac OS X 15.6.1 (24G90):
```
which timeout
timeout not found
```

### Expected behavior

`memmachine-compose.sh` should work on Linux and macOS X. Native Windows support can be added in the future.

### Environment

- OS: Mac OS X
- MemMachine v0.1.1

ChristianKniep:
I still get an error that the health endpoint is not reachable.

```
$ docker logs 7d31bb4eff82
Syncing schema to {'host': 'postgres', 'port': '5432', 'user': 'memmachine', 'password': '****', 'database': 'memmachine'}
Re-initializing ...
INFO:     Started server process [14]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
[...]

PR #102, opened 2025-09-19
fixing timeout in #100

ChristianKniep:
PR to fix the `timeout` problem on MacOSX (zsh). #100

### DIFF
--- a/memmachine-compose.sh
+++ b/memmachine-compose.sh
@@ -14,30 +14,31 @@ NC='\033[0m' # No Color
 
 ## Function to run a command with a timeout
 timeout() {
-  local duration=$1
-  shift
+    local duration=$1
+    shift
 
-  # Run the command in the background
-  "$@" &
-  local cmd_pid=$!
+    # Run the command in the background
+    "$@" &
+    local cmd_pid=$!
 
-  # Start a background sleep that will kill the command
-  (
-    sleep "$duration"
-    kill -0 "$cmd_pid" 2>/dev/null && kill -TERM "$cmd_pid"
-  ) &
+    # Start a background sleep that will kill the command
+    (
+        sleep "$duration"
+        kill -0 "$cmd_pid" 2>/dev/null && kill -TERM "$cmd_pid"
+    ) &
 
-  local watchdog_pid=$!
+    local watchdog_pid=$!
 
-  # Wait for the command to finish
-  wait "$cmd_pid"
-  local status=$?
+    # Wait for the command to finish
+    wait "$cmd_pid"
+    local status=$?
 
-  # Clean up watchdog if command finished early
-  kill -TERM "$watchdog_pid" 2>/dev/null
+    # Clean up watchdog if command finished early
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -94,7 +94,7 @@ services:
       neo4j:
         condition: service_healthy
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8080/health", "||", "exit", "1"]
+      test: ["CMD", "curl", "-f", "http://localhost:8080/docs", "||", "exit", "1"]
```
ChristianKniep: I thought that helps with the remaining `/health` endpoint issue...

sscargal: The use of `/docs` should only be required in the v0.1.0 release. The `/health` endpoint was introduced in v01.1. I recommend we stick with `/health` for now.

memmachine-compose.sh
```diff
@@ -12,6 +12,31 @@ YELLOW='\033[1;33m'
 BLUE='\033[0;34m'
 NC='\033[0m' # No Color
 
+## Function to run a command with a timeout
+timeout() {
+    local duration="$1"
+    shift
+    perl -e '
```
sscargal: Using `perl` adds yet another dependency to the MemMachine product. Perl is available on all distros, but we shouldn't expect it to be installed by default. I can't think of a universal solution that works across all Linux distributions and Mac OS X, but I think there may be another approach.

ChristianKniep:
@sscargal something missing or do you reckon we can merge that one?

svetly-todorov:
I tested this out on MacOS and Linux, and it seems to work well... let's roll back the /health change and merge!

svetly-todorov:
#100 was closed but this fix was not actually included. My PR did not include a fix for `timeout`. @sscargal @ChristianKniep Let's roll back the health endpoint change and merge this into `memmachine-compose.sh`.

docker-compose.yml
svetly-todorov:
```suggestion
      test: ["CMD", "curl", "-f", "http://localhost:8080/health", "||", "exit", "1"]
```

svetly-todorov (commented):
@ChristianKniep I think if we switch this endpoint from docs back to health, this PR is ready to go. @sscargal, what do you think?

docker-compose.yml
sscargal: Agreed. `/health` should be the correct testing endpoint.

svetly-todorov: Great! @ChristianKniep let's just commit this change and merge the timeout function!

sscargal: The correct test syntax is

```
test: ["CMD", "curl", "--fail", "--silent", "http://localhost:8080/health"]
```

This is currently in the `docker-compose.yml` file, and is the cause of the conflict.

Remove ` test: ["CMD", "curl", "-f", "http://localhost:8080/docs", "||", "exit", "1"]`

ChristianKniep:
Ok, I rebased the PR so it only deals with the timeout stuff.

PR #128, opened 2025-09-23
[do not merge] add tracing

ChristianKniep:
I hacked in some tracing to tip my toe into performance profiling.


ChristianKniep:
aged out

PR #152, opened 2025-09-26
Neo4j Vector Indexing

edwinyyyu:
### Purpose of the change

Current vector similarity search is exact KNN and linear in the number of stored vectors.
We want to use ANN to reduce the time complexity in order to support searching large amounts of data.

### Description

Add SimilarityMetric type for Embedder to give a recommendation on which similarity metric (currently includes cosine and euclidean) consumers should use. This is necessary for vector index to be built.

Add model_id, dimensions, and similarity_metric properties to Embedder.
- dimension tells the consumer what to expect without having to actually generate an embedding. This is necessary for some vector stores.
- model_id, dimensions enable generating unique names for fields containing vectors produced by different embedding models -- allows storing different embeddings in one database.
- similarity_metric allows the embedder to recommend a similarity metric to use.

Add dimensions field to OpenAIEmbedder config to support different embedding sizes for models that support it. Implemented while implementing dimensions property.

Disable exact similarity search in Neo4j by default -- use ANN using vector index instead.

Add embedding_property_name and similarity_metric to similarity search parameters. This allows the VectorGraphStore to support multiple different embeddings.

Update some docstrings.

Make interface for VectorGraphStore more generic in accepting collection or mapping types (instead of just accepting set or dict). This helps with unit testing and allows use without converting data structures.

Revamped configuration to use a Pydantic BaseModel instead of a dict.

Added integration tests to new directory. Moved unit tests to separate directory.

Updated evaluation scripts and evaluation config.

### Type of change

Breaks vector search behavior for consumers expecting exact number of returned vectors and the true most similar vectors.

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Refactor (does not change functionality, e.g., code style improvements, linting)
-   [ ] Documentation update
-   [ ] Project Maintenance (updates to build scripts, CI, etc., that do not affect the main project)
-   [ ] Security (improves security without changing functionality)

### How Has This Been Tested?

Added integration tests. Existing unit tests pass. Ran LoCoMo evaluation scripts after fixing evaluation scripts.

-   [ ] Unit Test
-   [x] Integration Test
-   [ ] End-to-end Test
-   [ ] Test Script (please provide)

### Checklist:

-   [ ] I have commented my code, particularly in hard-to-understand areas
-   [ ] I have checked my code and corrected any misspellings

### Maintainer Checklist

### Further comments

Neo4j `CREATE VECTOR INDEX IF NOT EXISTS` can fail with exception when executed concurrently, which is unexpected. Hence the code to create index is locked. This could be a Neo4j bug.

The new integration tests require Docker.

### DIFF
--- a/src/memmachine/common/embedder/openai_embedder.py
+++ b/src/memmachine/common/embedder/openai_embedder.py
@@ -89,7 +89,26 @@ class OpenAIEmbedder(Embedder):
             raise TypeError("Dimensions must be an integer")
 
         if dimensions <= 0:
-            raise ValueError("Dimensions must be a positive integer")
+            raise ValueError("Dimensions must be positive")
+
+        match self._model:
+            case "text-embedding-3-small":
+                if dimensions > 1536:
+                    raise ValueError(
+                        "Dimensions for text-embedding-3-small "
+                        "cannot be greater than 1536"
+                    )
+            case "text-embedding-3-large":
+                if dimensions > 3072:
+                    raise ValueError(
+                        "Dimensions for text-embedding-3-large "
+                        "cannot be greater than 3072"
+                    )
+            case "text-embedding-ada-002":
+                if dimensions != 1536:
+                    raise ValueError(
+                        "Dimensions for text-embedding-ada-002 must be exactly 1536"
+                    )
 
         self._dimensions = dimensions
 
@@ -173,8 +192,17 @@ class OpenAIEmbedder(Embedder):
                     attempt,
                     max_attempts,
                 )
-                response = await self._client.embeddings.create(
-                    input=inputs, model=self._model, dimensions=self.dimensions
+                response = (
+                    await self._client.embeddings.create(
+                        input=inputs,
+                        model=self._model,
+                        dimensions=self._dimensions,
+                    )
+                    if self._model != "text-embedding-ada-002"
+                    else await self._client.embeddings.create(
+                        input=inputs,
+                        model=self._model,
+                    )
                 )
                 break
             except (


sscargal (approved):
LGTM, this is a great improvement, thanks.

src/memmachine/common/embedder/data_types.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class SimilarityMetric(Enum):
+    COSINE = "cosine"
+    EUCLIDEAN = "euclidean"
```
malatewang, line 6: We may add dot product here to make the list complete.

src/memmachine/common/embedder/openai_embedder.py
```diff
@@ -48,11 +53,39 @@ def __init__(self, config: dict[str, Any]):
...
+            raise TypeError("Embedder API key must be a string")
+
+        model = config.get("model", "text-embedding-3-small")
+        if not isinstance(model, str):
+            raise TypeError("Model name must be a string")
+
+        self._model = model
+
+        # https://platform.openai.com/docs/guides/embeddings#embedding-models
+        dimensions = config.get("dimensions")
```
malatewang, line 73: This is Ok now. Probably we need a configuration for models. The embedding model needs the dimensions and the LLM model needs the context window size.

src/memmachine/common/vector_graph_store/vector_graph_store.py
```diff
@@ -18,33 +21,34 @@ class VectorGraphStore(ABC):
...
-            edges (list[Edge]): List of Edge objects to add.
+            edges (Collection[Edge]): Collection of Edge objects to add.
         """
         raise NotImplementedError
 
     @abstractmethod
     async def search_similar_nodes(
         self,
         query_embedding: list[float],
-        similarity_threshold: float = 0.2,
```
malatewang, line 44: Why remove the similarity threshold?

src/memmachine/common/embedder/openai_embedder.py
```diff
@@ -50,11 +55,39 @@ def __init__(self, config: dict[str, Any]):
...
+                case "text-embedding-3-large":
+                    dimensions = 3072
+                case "text-embedding-ada-002":
+                    dimensions = 1536
+                case _:
+                    raise ValueError(
+                        f"Unknown dimensions for model {model}."
+                        "Please specify dimensions in the configuration."
+                    )
+
```
malatewang, line 87: Since the dimension can be from configuration, it is better to check if the dimension value is valid here.

src/memmachine/common/embedder/openai_embedder.py
```diff
@@ -129,7 +162,7 @@ async def _embed(
                     max_attempts,
                 )
                 response = await self._client.embeddings.create(
-                    input=inputs, model=self._model
+                    input=inputs, model=self._model, dimensions=self.dimensions
```
malatewang: Should be self._dimensions?

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -127,97 +151,196 @@ async def add_edges(self, edges: list[Edge]):
...
+                for relation, edges in relation_edges_map.items()
+            )
         ]
 
         await asyncio.gather(*add_edges_tasks)
 
     async def search_similar_nodes(
         self,
         query_embedding: list[float],
-        similarity_threshold: float = 0.2,
```
malatewang, line 152: Why remove the similarity threshold? This should be useful. I think we should normalize the similarity score for upper layer to use.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -127,97 +151,196 @@ async def add_edges(self, edges: list[Edge]):
...
-        required_properties: dict[str, Property] = {},
+        required_labels: Collection[str] | None = None,
+        required_properties: Mapping[str, Property] = {},
         include_missing_properties: bool = False,
     ) -> list[Node]:
-        async with self._semaphore:
-            records, _, _ = await self._driver.execute_query(
+        exact_similarity_search = self._force_exact_similarity_search
+
+        sanitized_property = Neo4jVectorGraphStore._sanitize_name(
```
malatewang: sanitized_embedding_property_name?

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
edwinyyyu: It is not really useful if there is a hard limit set.

Also, distribution of similarity scores can vary between different embedding models. For example, for a vector with 1 dimension, 50% of vectors have similarity -1 and 50% of vectors have similarity 1, assuming a uniform distribution of vectors. This distribution does not hold for higher dimensions.

Also, different vector database implementations do different things for scoring. We would have to document what the score means and what distribution it is meant for, so that the caller can use it well. Then we have to translate that distribution to whatever distribution the underlying implementation uses before giving it in the database query. For example, Euclidean scoring can be done as just Euclidean distance, squared Euclidean distance, 1 / (1 + Euclidean distance), or 1 / (1 + squared Euclidean distance). There are other equivalents for cosine similarity.

src/memmachine/common/embedder/openai_embedder.py
edwinyyyu: Can you explain what you mean? This block is only executed if the dimensions is not provided and cannot be inferred with a reasonable default, so there is no value to check.

src/memmachine/common/embedder/data_types.py
edwinyyyu: Do you mean cosine, which is computed using the dot product? Raw dot product is not usually a good measure of similarity and vector databases generally do not support it.

src/memmachine/common/embedder/openai_embedder.py
edwinyyyu: Both do the same thing right now. dimensions from public API should be more reliable if the implementation ever changes. I don't do it for model because internally the same model can be differently named for different platforms (maybe difference in hyphens or underscores, capitalization), but dimensions is an int so it should be universal.

src/memmachine/common/embedder/data_types.py
malatewang: Since we make a list of the similarity metric methods, it is better to make a complete list.

src/memmachine/common/embedder/openai_embedder.py
malatewang: Because the dimension is provided in the configuration, the value may be invalid. The provided dimension can not be greater than what be supported by the model. If we know the supported dimension of the model, we should check if the configured value is valid.

src/memmachine/common/embedder/openai_embedder.py
malatewang: This just looks a little inconsistent

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
malatewang: Ok. Looks like most vector DB does not support similarity threshold based search.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
edwinyyyu: Updated.

src/memmachine/common/embedder/openai_embedder.py
edwinyyyu: Actually I think using internal dimensions is better because it's implementation-specific what is passed to the client.

src/memmachine/common/embedder/openai_embedder.py
edwinyyyu: Added checks.

src/memmachine/common/vector_graph_store/vector_graph_store.py
edwinyyyu: Resolving so that discussion is in other comment thread.

src/memmachine/common/embedder/data_types.py
edwinyyyu: There is no dot product similarity that I'm aware of that's different from cosine similarity.

edwinyyyu: I think we add more similarity metrics when they become supported.

Issue #99, opened 2025-09-19
[Feat]: Implement a robust and extensible unit testing framework with Pytest


sscargal:
### Is your feature request related to a problem?

The MemMachine project currently lacks a standardized and automated unit testing framework. This poses several challenges:
      1.  **Risk of Regressions:** Without an automated test suite, it's difficult to ensure that new changes don't inadvertently break existing functionality.
      2.  **Higher Barrier to Contribution:** New contributors cannot easily validate their changes or feel confident that they haven't introduced bugs.
      3.  **Code Quality:** Verifying the correctness of new features relies on manual testing, which is time-consuming, error-prone, and not scalable for a growing open-source project. This slows down the development and release cycle.

### Describe the solution you'd like

I propose implementing a comprehensive unit testing framework using **Pytest** and integrating it into our development workflow via **GitHub Actions**.

### 1. Framework and Structure

- **Adopt Pytest:** Utilize the `pytest` framework for its simple assertion syntax, powerful fixture system, and extensive plugin ecosystem, which makes writing and maintaining tests straightforward.
- **Directory Structure:** Create a dedicated `tests/` directory at the project root. The structure within `tests/` should mirror the main `memmachine/` source directory to keep tests organized and discoverable.
      
```text
memmachine/
├── src/memmachine/
│       └── ... (source code)
├── tests/
│   └── ... (tests mirroring the source structure)
└── pyproject.toml
```

### 2. CI/CD Integration

- **Automated Testing:** Create a GitHub Actions workflow (e.g., `.github/workflows/ci-tests.yml`) that automatically triggers on every pull request and push to the `main` branch.
    - **Workflow Steps:** This workflow will:
      1.  Check out the code.
      2.  Set up a Python environment.
      3.  Install project dependencies and testing requirements (like `pytest`).
      4.  Run the entire test suite using the `pytest` command.
    - **Status Checks:** Configure the repository to require that the test suite passes before a pull request can be merged. This will prevent regressions from being introduced into the main codebase.

### 3. Contributor Guidelines

- **Test Requirement:** Establish a policy that all new features, enhancements, and bug fixes **must** be accompanied by corresponding unit tests.
- **Documentation:** Update the `CONTRIBUTING.md` file with a new "Testing" section that guides developers on how to write, run, and structure their tests. This will serve as a clear example for all contributors.

### 4. Code Coverage

[...]

PR #182, opened 2025-10-01
Add unit tests for profile memory

o-love:
### Purpose of the change

Adds unit test coverage to profile memory to help ensure functioning.

### Description

Implements testing for profile memory using an in memory storage and a mock embedder and LLM.
Also includes integration tests for Profile Memory.

Todo for future PRs:
- Write unit tests to validate semantic search returns closest features.

### Type of change

### How Has This Been Tested?

- [x] Unit Test
- [ ] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

#99


src/memmachine/profile_memory/profile_memory.py
```diff
@@ -75,7 +77,12 @@ def __init__(
         self._consolidation_prompt = getattr(
             prompt_module, "CONSOLIDATION_PROMPT", ""
         )
-        self._profile_storage = AsyncPgProfileStorage(db_config)
+        if profile_storage is None:
+            assert db_config is not None, "db_config must be provided"
```
malatewang: Just raise exception should be good enough.

src/memmachine/profile_memory/profile_memory.py
```diff
@@ -313,6 +320,7 @@ async def add_persona_message(
         metadata: dict[str, str] | None = None,
         isolations: dict[str, bool | int | float | str] | None = None,
         user_id: str = "",  # TODO fully deprecate user_id parameter
+        wait_consolidate: bool = False,
```
malatewang: How can this parameter work?

src/memmachine/profile_memory/profile_memory.py
```diff
@@ -57,12 +58,13 @@ class ProfileMemory:
     """
 
     def __init__(
-        self,
+        self, *,
         model: LanguageModel,
         embeddings: Embedder,
-        db_config: dict[str, Any],
+        db_config: dict[str, Any] | None = None,
```
malatewang: db_config has default value but prompt_module does not. The db_config must be after the promp_module

src/memmachine/profile_memory/storage/asyncpg_profile.py
```diff
@@ -395,8 +415,11 @@ async def get_last_history_messages(
         self,
         user_id: str,
         k: int = 0,
-        isolations: dict[str, bool | int | float | str] = {},
+        isolations: dict[str, bool | int | float | str] = None,
```
malatewang: should be: isolations: dict[str, bool | int | float | str] | None= None

tests/memmachine/profile_memory/storage/in_memory_profile_storage.py
```diff
@@ -0,0 +1,436 @@
+from __future__ import annotations
```
malatewang, line 1: What is the purpose of the in memory storage?

malatewang (dismissed):

tests/memmachine/profile_memory/storage/in_memory_profile_storage.py
o-love: The purpose of in memory storage is to allow unit testing and other testing without needing the PostgreSQL backend.

src/memmachine/profile_memory/profile_memory.py
o-love: When adding profile memory, it doesn't necessarily wait for the memory to be processed and consolidated into the profile. 
This option forces the addition of a new message to await the ingestion into the Profile.

My other work in moving the ingestion to a background task will have this be a function,
`wait_all_ingestion`. Which will return once all current background processing is complete.

tests/memmachine/profile_memory/storage/in_memory_profile_storage.py
malatewang: If it is just used for test, I think it should be put into this location.

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    integration: integration test that requires environment
```
edwinyyyu: Can we add this to `pyproject.toml` instead?
https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml

src/memmachine/profile_memory/profile_memory.py
```diff
@@ -57,12 +58,13 @@ class ProfileMemory:
...
     def __init__(
-        self,
+        self, *,
         model: LanguageModel,
         embeddings: Embedder,
-        db_config: dict[str, Any],
         prompt_module: ModuleType,
+        db_config: dict[str, Any] | None = None,
         max_cache_size=1000,
+        profile_storage: ProfileStorageBase | None = None,
```
edwinyyyu: I think eventually the config and validation logic can be replaced by Pydantic BaseModels. No changes needed now.

See #152 for example configuration for Neo4jVectorGraphStore.

src/memmachine/profile_memory/storage/storage_base.py
edwinyyyu, line 1: Interface should have docstrings to explain what each method does.
Ignore if out of scope for this PR.

tests/memmachine/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--integration",
+        action="store_true",
+        default=False,
+        help="Run integration tests"
+    )
+
+def pytest_collection_modifyitems(config, items):
```
edwinyyyu, line 14: Is there a rule for the underscore placement?

tests/memmachine/__init__.py
edwinyyyu, line 1: Remove file duplicates.

tests/memmachine/profile_memory/test_profile_memory_integration.py
```diff
@@ -0,0 +1,137 @@
...
+        await self.ingest_question_convos("test_user",
+                                          profile_memory=profile_memory,
+                                          conversation_sessions=long_mem_convos)
+
+        eval_resp = await self.eval_answer("test_user",
+                                           profile_memory=profile_memory,
+                                           question_str=long_mem_question,
+                                           llm_model=llm_model)
+
+        assert "The relevant information is not found in the user's persona profile" not in eval_resp
```
edwinyyyu: Flaky test: I think this is highly dependent on model/prompt. Maybe such a test should be put under evaluation instead, since a poorly functioning but technically working/API-conforming memory can still produce this result.

tests/memmachine/profile_memory/test_profile_memory_integration.py
```diff
@@ -0,0 +1,137 @@
...
+        "model": config["embedding_model"]
+    })
+
+
+@pytest.fixture
+def llm_model(config):
+    return OpenAILanguageModel({
+        "api_key": config["api_key"],
+        "model": config["llm_model"]
+    })
```
edwinyyyu: Using real models is expensive and flaky. Can you mock or fake this?

Also, where is the API key coming from?

edwinyyyu (changes requested):

tests/memmachine/conftest.py
o-love: In this case it is a specific hook https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.hookspec.pytest_collection_modifyitems

tests/memmachine/__init__.py
o-love: __init__.py is needed for tests to import other modules in the test directory with each folder needing its own `__init__.py`.

tests/memmachine/profile_memory/test_profile_memory_integration.py
o-love: I agree, I modified the prompt to have it specifially return that frase. But with it being an integration test that needs to be explicitlly ran by the user I felt that it is ok.

tests/memmachine/profile_memory/test_profile_memory_integration.py
o-love: In this case it comes form the fixture some lines up.

This test is ment to be an integration test manually ran by the user.

The test won't run unless they pass the `--integration` flag.

tests/memmachine/__init__.py
edwinyyyu: Can't they all import the nltk setup code from a single file?
I don't think it's sustainable to maintain so many copies of the same file.

Why not `from memmachine import setup_nltk` since it's in the memmachine package anyway?

tests/memmachine/profile_memory/test_profile_memory_integration.py
edwinyyyu: I think we need a category for automated integration tests that should always pass (and without user adding auth information), for example for the database wrappers (e.g. #152). But it should be separated from unit tests due to performance.

What do you think about creating a flag for each category of tests, if we want to use the pytest framework for everything?

edwinyyyu (approved):
LGTM

PR #217, opened 2025-10-04
Introduce async profile ingestion

o-love:
### Purpose of the change

Instead of processing ingestion in real time. Set the framework to have ingestion for profile memory to be done in the background.

### Description

Adds a new column to the raw data store of Profile memory to indicate if a message has been igested or not.

Then a background task loads these uningested messages and processes them.

This way adding messages to profile memory never block due to the LLM call.

Next steps:
- Batch calls to the LLM. In which multiple messages of the same user get sent in a single prompt to the LLM.

### Fixes/Closes

Fixes #(issue number)

### Type of change

Breaking change:
Requires updating the profile store database.

### How Has This Been Tested?

- [ ] Unit Test
- [x] Integration Test
- [ ] End-to-end Test

Unit test for this will be part of #182 

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### DIFF
--- a/src/memmachine/profile_memory/profile_memory.py
+++ b/src/memmachine/profile_memory/profile_memory.py
@@ -7,9 +7,10 @@ information extraction and a vector database for semantic search capabilities.
 """
 
 import asyncio
+import datetime
 import json
 import logging
-from itertools import accumulate, tee
+from itertools import accumulate, groupby, tee
 from types import ModuleType
 from typing import Any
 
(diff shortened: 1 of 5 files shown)


src/memmachine/profile_memory/profile_memory.py
```diff
@@ -313,6 +326,7 @@ async def add_persona_message(
         metadata: dict[str, str] | None = None,
         isolations: dict[str, bool | int | float | str] | None = None,
         user_id: str = "",  # TODO fully deprecate user_id parameter
+        wait_consolidate: bool = False,
```
malatewang: This parameter is never used. What is its purpose?

src/memmachine/profile_memory/profile_memory.py
```diff
@@ -336,42 +352,67 @@ async def add_persona_message(
...
     ):
-        """request re-ingestion of session data to profile"""
-        if isolations is None:
-            isolations = {}
+        message_isolation_groups = await self._get_isolation_grouped_memories(user_id)
+
+        async def process_messages(messages):
+            for i in range(0, len(messages)-1):
+                message = messages[i]
+                await self._update_user_profile_think(message)
```
malatewang, line 518: The profile should be updated by the message in the arriving order. I think this can not guarantee it. I suggest to order the message by timestamp and update the profile in batch.

src/memmachine/profile_memory/profile_memory.py
```diff
@@ -336,42 +352,67 @@ async def add_persona_message(
...
+
+            await self._update_user_profile_think(messages[-1], wait_consolidate=True)
```
malatewang, line 523: Consolidation can be expensive. You should count the number of ingested messages and do consolidation only if necessary.

src/memmachine/profile_memory/profile_memory.py
```diff
@@ -336,42 +352,67 @@ async def add_persona_message(
...
+    async def _get_isolation_grouped_memories(self, user_id: str):
+        rows = await self._profile_storage.get_ingested_history_messages(
+            user_id=user_id, k=1_000_000, is_ingested=False,
         )
-        if wait_consolidate:
-            await fut
-        return wait_consolidate
 
-    async def _reconsolidate_memory(
+        def key_fn(r):
```
malatewang, line 497: It is better to do the normalization before save the message to the database

src/memmachine/profile_memory/profile_memory.py
o-love: It gets ordered at the storage level with the sql query.

Maybe explicitly make it a requirement (documentation) for the storage layer to return it ordered.

malatewang: Even if the messages are ordered, I think the asyncio can not guarantee the execution order.

o-love: Correct. 
Each profile has a single thread, with it awaiting each message before processing the next message for that profile

src/memmachine/profile_memory/profile_memory.py
o-love: Not sure what you mean?

src/memmachine/profile_memory/profile_memory.py
o-love: I'm thinking of pushing this back to the token usage/batching optimization for profile memory

src/memmachine/profile_memory/storage/asyncpg_profile.py
```diff
@@ -391,30 +411,56 @@ async def delete_history(
         async with self._pool.acquire() as conn:
             await conn.execute(stm, user_id, json.dumps(isolations))
 
-    async def get_last_history_messages(
+    async def get_ingested_history_messages(
```
edwinyyyu: Naming is confusing because this says to get ingested but then there is an `is_ingested` parameter. Perhaps you can use different terminology for whether a message has been added to database and whether a message has been processed by LLM.

src/memmachine/profile_memory/storage/storage_base.py
edwinyyyu, line 1: Can you add docstrings for new methods?

src/memmachine/profile_memory/profile_memory.py
```diff
@@ -75,9 +77,18 @@ def __init__(
...
+        if profile_storage is None:
+            if db_config is None:
+                raise ValueError("db_config must be provided if profile_storage is None")
+            self._profile_storage = AsyncPgProfileStorage(db_config)
+        else:
+            self._profile_storage = profile_storage
+
         self._update_interval = 1
-        self._msg_count: dict[str, int] = {}
+        self._dirty_users: dict[str, bool] = {}
```
edwinyyyu: Is there a reason this is not a set?

src/memmachine/profile_memory/storage/storage_base.py
jealous: done.

src/memmachine/profile_memory/storage/asyncpg_profile.py
jealous: renamed to `get_history_messages_by_ingestion_status`

src/memmachine/profile_memory/profile_memory.py
malatewang: This can be done with a simple per-user counter.

src/memmachine/profile_memory/profile_memory.py
jealous: I don't see a reason.  Updated to set.

jealous: Update it to a class to manage the lock and the dirty status.

jealous:
@o-love @malatewang Uses a manager class to manage trackers and determine whether to update
the profile based on two conditions:

- If the total count of messages exceeds a limit (default: 5)
- If the elapsed time exceeds the maximum profile update interval
  (default: 2 minutes)

src/memmachine/profile_memory/storage/asyncpg_profile.py
```diff
@@ -357,9 +405,14 @@ async def add_history(
         self,
         user_id: str,
         content: str,
-        metadata: dict[str, str] = {},
-        isolations: dict[str, bool | int | float | str] = {},
+        metadata: dict[str, str] | None = None,
+        isolations: dict[str, bool | int | float | str] | None = None,
     ) -> asyncpg.Record:
```
edwinyyyu: Public method should not return implementation-specific type.
Ensure that the typing matches the interface.

o-love: Yeah, Apparently if the subclass returns something but the parent class doesn't mypy dosn't fix it.

Did another pass between the subclass and the parent class to ensure a consistent interface.

PR #241, opened 2025-10-09
Add a README to the Health assistant example

ChristianKniep:
### Purpose of the change

Add guidance for the Hackathon.


SarahScargall:
Watching for the merge of this PR, so I can update the Docs site.

examples/health_assistant/README.md
sscargal, line 1: We should include a description of this healthcare example. What is it, what it does, etc. Perhaps include a statement that it uses a Streamlit UI.

memmachine-compose.sh
sscargal, line 1: Do we need a local copy of `memmachine-compose.sh` in this example directory? If no, please remove it. If so, we need to consider how to handle the situation, as each example requires its own copy, as it will be challenging to keep them in sync unless we use soft links to the parent script. Softlinks may not be cross-architecture or multi-OS compatible.

sscargal (changes requested):

memmachine-compose.sh
ChristianKniep: it's in the root of the repo. I added the `timeout()` function again as it's not working on my zsh without. I'll remove it.

ChristianKniep: removed the file @sscargal

Charlie-Yi-2002:
The additions look good to me, but I'm wondering if it might be weird for a user to see that we have bruno in this health assistant but not the other ones ? @sscargal @ChristianKniep @gauri-nagavkar

ChristianKniep:
I was not really aware that the streamlit frontend is also able to show the health assistant.
But on the other hand: I am all for adding bruno files everywhere - it's a good clean way of demonstration to developers and ops folks how things work.

sscargal:
Closing as the PR is 6 months old. If required, please reopen or create a new PR that uses the rebased code.

PR #245, opened 2025-10-09
Support session data in the header.

### LINKED ISSUE #234: [Feat]: Move ID fields from request body to header

I want it easier to support external gateway that does user authorization. Otherwise, the gateway has to decode the request body for every request. Not every gateway supports that. Besides, decoding request body adds significant overhead.

Move ID fields to the request header. In this way, the gateway can reject unauthorized requests even before receiving the request body.

_No response_

### Additional context

_No response_

jealous:
### Purpose of the change

Support extracting session information from request headers, improve `/v1/memories` related API tests.

### Description

* Add logic in app.py to extract session fields from HTTP request headers, handling cases where the payload is missing session
* Update test-rest-memories.py to add/improve test cases for passing session information via headers
* Fix some test assertions for session validation to ensure consistent API behavior
### Fixes/Closes

Fixes #234

### Type of change

- [x] Documentation update

### How Has This Been Tested?

* Unit Test

All related endpoints have passed automated tests, covering both scenarios where session information is provided in the payload and in the header, ensuring consistent behavior with no regressions.

- [x] Unit Test

**Test Results:** passed

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,7 @@ dev = [
 
 [tool.mypy]
 follow_untyped_imports = true
+exclude = ["venv/", ".venv/"]
 
 [tool.ruff.lint]
 select = ["E", "F", "I", "W"]
--- a/src/memmachine/server/app.py
+++ b/src/memmachine/server/app.py
@@ -16,16 +16,18 @@ import os
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from importlib import import_module
-from typing import Any, cast
+from typing import Any, Self, cast
 
 import uvicorn
 import yaml
 from dotenv import load_dotenv
-from fastapi import FastAPI, HTTPException
+from fastapi import FastAPI, Header, HTTPException
+from fastapi.exceptions import RequestValidationError
+from fastapi.params import Depends
 from fastapi.responses import Response
 from fastmcp import Context, FastMCP
 from prometheus_client import CONTENT_TYPE_LATEST, generate_latest
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from memmachine.common.embedder.openai_embedder import OpenAIEmbedder
 from memmachine.common.language_model.openai_language_model import (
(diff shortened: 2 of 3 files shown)


edwinyyyu:
@malatewang I think we had that info in the header before, but it was removed. Do you remember why?

malatewang:
> @malatewang I think we had that info in the header before, but it was removed. Do you remember why?

Yes. I remember at that time, it is not a good practice to put the session data in the header.

jealous:
It's required by the proxy I think.  @xiongzubiao

jealous:
The main reason for moving the session information to the request headers is that authentication and security will be handled by a platform-level proxy or gateway. After verifying the request, the proxy will inject user and session details into the headers so that downstream services can identify the requester. Modifying the request body at that stage would be complex and inappropriate, whereas adding or updating headers is a standard and reliable approach.

xiongzubiao:
> > @malatewang I think we had that info in the header before, but it was removed. Do you remember why?
> 
> Yes. I remember at that time, it is not a good practice to put the session data in the header.

Can you please elaborate why?

malatewang:
> > > @malatewang I think we had that info in the header before, but it was removed. Do you remember why?
> > 
> > 
> > Yes. I remember at that time, it is not a good practice to put the session data in the header.
> 
> Can you please elaborate why?

There is no special reason. We had no experience on HTTP API. We did some search and found nobody did that.

jealous:
@edwinyyyu I also update the `mypy` settings to exclude `venv` or `.venv`.

jealous:
Move logic to internal functions prefixed with `_` so they can be shared
by MCP and REST. The signature change for REST handlers will not break
MCP, which can still use the old method to get session data.

xiongzubiao:
Mem0 Indeed has user_id in request body. However, I found that in Mem0 one account can create multiple users. The API key from an account can see memories from all users under this account. That doesn't seem to align with what we want (user memory isolation).

xiongzubiao:
- I think the user id and group id should not be allowed in request body at all. Otherwise, it can be confusing. One might still specify the user_id in request body and expect it to work, but it won't in the SaaS version...

- I am not sure about session id or agent id. At this moment, they don't seem tied to user identity. @malatewang @edwinyyyu , what you do think?

jealous:
I mark them deprecated in the PR.  Maybe we can remove them in 2 or 3 release.

jealous:
@malatewang I updated the code so that it doesn't affect the MCP.

edwinyyyu:
@xiongzubiao 
The current model seems to be:
We have no concept of an account right now.
A user_id represents a user (likely human) who can produce data to feed into the memory.
We do not use agent_id really. It's basically another user_id but it's supposed to represent non-human agents.
We don't isolate at the user level. Groups are used for isolation. Groups should be uniquely defined by group_id. I think two groups with different ids can have the same lists of users and agents.
Sessions are used to separate short-term memories. There can be multiple sessions per group.


Absolutely none of these should be present in the MCP API parameters, because the expectation is that the AI agent populates all parameters in an MCP tool.

I don't know what the best practice is for REST.

PR #285, opened 2025-10-14
Add gitstar

ChristianKniep:
### Purpose of the change

To raise awareness a graph is added to show the history of our github stars.


ChristianKniep:
@SarahScargall What do you think?

SarahScargall:
> @SarahScargall What do you think?

Hi @ChristianKniep , for the docs site, we're using "https://starchart.cc/MemMachine/MemMachine".  I haven't checked out app.repohistory.com much, but it is doing the same type of chart, just with a bit softer font.  Is there a reason you prefer app.repohistory.com?  I'm not against change, but I'm not sure what it provides in benefit that is stronger than starchart.cc.

ChristianKniep:
> > @SarahScargall What do you think?
> 
> Hi @ChristianKniep , for the docs site, we're using "https://starchart.cc/MemMachine/MemMachine". I haven't checked out app.repohistory.com much, but it is doing the same type of chart, just with a bit softer font. Is there a reason you prefer app.repohistory.com? I'm not against change, but I'm not sure what it provides in benefit that is stronger than starchart.cc.

No reason... lemme change it to starchart.cc

README.md
```diff
@@ -8,6 +8,12 @@
 
 </div>
 
+## Growing Community
+
+MemMachine is a growing community of builders and developers. Please help us grow by clicking the *Star* button above.
+
+<img src="https://starchart.cc/MemMachine/MemMachine.svg?variant=light" alt="Alt text" height="300"/>
```
o-love, line 15: Only nit is if you can align it in center

PR #297, opened 2025-10-15
[CI] Run tests on all OSs

o-love:
In order to ensure our code runs on all operating systems. Add the 3 major OSs to our CI pipeline

#99


o-love:
@edwinyyyu Docker isn't supported on arm mac github actions. Specifically they don't support nested virtualization. So running a Linux VM (in order to run the docker container) isn't allowed.
See https://github.com/docker/setup-docker-action/blob/bca5e4d3e252bb47ebed85515713050e55f7eb15/README.md?plain=1#L12-L17

I'm wondering if we can explore how to setup neo4j without docker. Maybe having the github action pull it in.


Also if you can take a look at the permission issue with windows

edwinyyyu:
> @edwinyyyu Docker isn't supported on arm mac github actions. Specifically they don't support nested virtualization. So running a Linux VM (in order to run the docker container) isn't allowed. See https://github.com/docker/setup-docker-action/blob/bca5e4d3e252bb47ebed85515713050e55f7eb15/README.md?plain=1#L12-L17
> 
> I'm wondering if we can explore how to setup neo4j without docker. Maybe having the github action pull it in.
> 
> Also if you can take a look at the permission issue with windows

For Neo4j, it may not be worth setting up testing for all OS. Also, I'm thinking in the future there may be certain integrations that don't natively support all major OS anyway, so I think a solution could be to enable/disable integration tests individually for different OS.

Is the Windows permission issue documented anywhere?

o-love:
> > @edwinyyyu Docker isn't supported on arm mac github actions. Specifically they don't support nested virtualization. So running a Linux VM (in order to run the docker container) isn't allowed. See https://github.com/docker/setup-docker-action/blob/bca5e4d3e252bb47ebed85515713050e55f7eb15/README.md?plain=1#L12-L17
> > I'm wondering if we can explore how to setup neo4j without docker. Maybe having the github action pull it in.
> > Also if you can take a look at the permission issue with windows
> 
> For Neo4j, it may not be worth setting up testing for all OS. Also, I'm thinking in the future there may be certain integrations that don't natively support all major OS anyway, so I think a solution could be to enable/disable integration tests individually for different OS.
> 
> Is the Windows permission issue documented anywhere?

I agree with disabling the inregration tests on macos. Since docker is a convenient way to run these tests.

The issue with Windows is that the SessionManager isn't closing the sqlite connection before we try to delete the file. 

While in Linux this is hidden due to deletion being unlinks. Allowing files to still stay open after they have been removed from the fs.

It does look like it should be closed here
https://github.com/MemMachine/MemMachine/blob/82307ff5d9b812a6c077d62350e2a3a5a353be59/src/memmachine/episodic_memory/session_manager/session_manager.py#L145-L149

Maybe the sqlalchemy library isn't properly closing everything

edwinyyyu:
> > > @edwinyyyu Docker isn't supported on arm mac github actions. Specifically they don't support nested virtualization. So running a Linux VM (in order to run the docker container) isn't allowed. See https://github.com/docker/setup-docker-action/blob/bca5e4d3e252bb47ebed85515713050e55f7eb15/README.md?plain=1#L12-L17
> > > I'm wondering if we can explore how to setup neo4j without docker. Maybe having the github action pull it in.
> > > Also if you can take a look at the permission issue with windows
> > 
> > 
> > For Neo4j, it may not be worth setting up testing for all OS. Also, I'm thinking in the future there may be certain integrations that don't natively support all major OS anyway, so I think a solution could be to enable/disable integration tests individually for different OS.
> > Is the Windows permission issue documented anywhere?
> 
> I agree with disabling the inregration tests on macos. Since docker is a convenient way to run these tests.
> 
> The issue with Windows is that the SessionManager isn't closing the sqlite connection before we try to delete the file.
> 
> While in Linux this is hidden due to deletion being unlinks. Allowing files to still stay open after they have been removed from the fs.
> 
> It does look like it should be closed here
> 
> https://github.com/MemMachine/MemMachine/blob/82307ff5d9b812a6c077d62350e2a3a5a353be59/src/memmachine/episodic_memory/session_manager/session_manager.py#L145-L149
> 
> Maybe the sqlalchemy library isn't properly closing everything

@malatewang Do you know what's going on?

o-love:
@edwinyyyu @malatewang 

With the integration tests cleaned out above are the failures running unit tests on windows.
Where it looks like after suposidly closing the sqlite connection it hasn't been properly been closed.
And since this is Windows we can't delete an open file.

cc @jealous if you have any input

plasticity-cloud:
@o-love @edwinyyyu we might look at setting up custom github runners on AWS,
so that you guys could fully leverage it.

We are working with @ChristianKniep on Github ARC integration, via Kubernetes,
but we could spin up EC2 Mac instances on demand, no issues, via AWS Controllers for Kubernetes - https://github.com/aws-controllers-k8s/ec2-controller

PR #300, opened 2025-10-15
Convert profile prompt to dataclass

### LINKED ISSUE #275: [Feat]: The prompt accepted by profile memory should be an interface and not a module

Currently, the prompt used by the Profile Memory component is defined and passed as a module. This makes it rigid to extend, mock, or replace dynamically—especially in cases where multiple prompt sources or runtime prompt composition are needed.

Introduce an interface (e.g., PromptProvider or PromptInterface) that defines the expected contract for the prompt object. Profile Memory would depend on this interface instead of importing a specific module.

_No response_

### Additional context

_No response_

o-love:
### Purpose of the change

Removes the use of modules as the definition of Profile Memory prompt in favor of a data class.

### Description

Removes the dependency on modules inside profile memory. This allows the use of profile memory without needing to define the prompt to be used in a module.
This allows bench-marking and tests to be easily done.

As well as easing the use of the profile memory interface
 
### Fixes/Closes

Fixes #275 

### Type of change

### How Has This Been Tested?

Manually tested + ran all unit tests.

### Checklist

### DIFF
--- a/src/memmachine/profile_memory/profile_memory.py
+++ b/src/memmachine/profile_memory/profile_memory.py
@@ -11,7 +11,6 @@ import datetime
 import json
 import logging
 from itertools import accumulate, groupby, tee
-from types import ModuleType
 from typing import Any
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


svetly-todorov (approved):
Tried to familiarize myself with the code in order to understand the change. I think it makes sense. Before, to test the ProfileMemory class, you had to have a different `test_prompt.py` file containing UPDATE_PROMPT and CONSOLIDATE_PROMPT. Now, you can just have pass strings to an instance of ProfilePrompt

\+ Unit tests pass on my side too

jealous (approved):

PR #318, opened 2025-10-16
[CI] Support Integration tests starting with ProfileMemory

o-love:
### Purpose of the change

Adds support for running integration tests as part of the github action using OpenAi and docker containers. This provides full testing for MemMachine.

This also provides a framwork for adding in end to end tests on the Rest API.

### Description

Multiple changes:
- Run python tests on arm and amd
- Run integration tests using OpenAI and PostgreSQL
- Have Profile Memory Integration test use PostgreSQL instead of in-memory.

### Type of change

### How Has This Been Tested?

Integration test ran locally and passing in CI.

- [X] Integration Test
- [X] Manual verification (list step-by-step instructions)

### Checklist

#99


.github/workflows/pytest-integration.yml
```diff
@@ -0,0 +1,34 @@
+name: Run Python integration tests
+on:
+  push:
+    branches: [ main, integration-test ]
```
o-love: Have branch `integration-test` so that we can see it running as part of the PR. Will remove this when merging.

With PR's it will the be ran through `pull_request_target`

.github/workflows/pytest-integration.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Install the latest version of uv and set the python version
+        uses: astral-sh/setup-uv@v6
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Test ${{ matrix.test-paths }} with python ${{ matrix.python-version }}
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        run: uv run --frozen --all-extras pytest ${{ matrix.test-paths }} --manual-integration
```
edwinyyyu: If you are using testcontainers only, then no need to mark it as manual integration right?

And we should use the same mark for this and the Neo4j integration tests.

.github/workflows/pytest.yml
edwinyyyu, line 1: I believe this will also run the integration tests automatically. They must be explicitly disabled to run unit tests only.

jealous: will it? @o-love

o-love: No it won't. 
I set up the flag to only run those integration test when the flag is set.

https://github.com/MemMachine/MemMachine/blob/3c15a5557b4abc096fb5b7114b3469feeb728246/tests/memmachine/conftest.py#L5-L22

.github/workflows/pytest-integration.yml
o-love: That could be something to discuss. What dependencies are acceptable in the default unit tests and what should be separated into integration tests.

But it could be good to have the docker dependency be seperate from the unit tests

.github/workflows/pytest.yml
edwinyyyu: I assumed the manual-integration flag was for things that require manual setup (cannot automate).

Should I change neo4j tests to use the manual-integration mark?

o-love: Maybe the other way around is better. Remove the `--manual-integration` flag and switch everything to a default false `--integration` flag.
So have a single class of integration tests instead of two. 
And all integration tests have to be tested as part of the CI.

So the Neo4j will be ran in this integration pipeline instead of the base unit test one.

edwinyyyu: I agree with using the name 'integration' but the behavior of the manual-integration flag as in this PR. I think 'manual' should be reserved for tests that are not automated.

o-love: @edwinyyyu Updated to have all integration tests under `--integration` and have them all running on the integration CI only

edwinyyyu (approved):
LGTM

PR #326, opened 2025-10-17
Validate OpenAI-compatible embedder dimensions

### LINKED ISSUE #304: [Bug]: Dimensions Parameter Should Be Configurable for Exclusion in OpenAIEmbedder
### Describe the bug

Currently, the OpenAIEmbedder excludes the dimensions parameter for specific hardcoded model names (text-embedding-ada-002). This logic is not configurable via the config file, making it difficult to support new models or custom exclusion rules without code changes.

### Steps to reproduce

1. Add a new embedding model to the config that does not support the dimensions parameter.
2. Set dimensions in the config file.
3. The embedder will attempt to send dimensions unless the model name is hardcoded in the exclusion list.

### Expected behavior

The exclusion of the dimensions parameter should be configurable via the config file, allowing users to specify which models should omit dimensions without modifying the source code.

### Environment

K8S + vLLM

### Additional context

_No response_

edwinyyyu:
### Purpose of the change

Non-recognized OpenAI-compatible embedding models were assumed to support the dimensions parameter. This is a wrong assumption, so the OpenAIEmbedder has been updated to validate the user-provided dimensions by making a dummy embedding.

### Description

Validate dimensions by embedding a dummy string.

### Fixes/Closes

Fixes #304

### Type of change

Breaks any configurations where a custom dimensions was already specified.

- [ ] Documentation update

### How Has This Been Tested?

Not tested.

- [ ] Unit Test
- [ ] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### DIFF
--- a/src/memmachine/common/embedder/openai_embedder.py
+++ b/src/memmachine/common/embedder/openai_embedder.py
@@ -38,17 +38,9 @@ class OpenAIEmbedder(Embedder):
                   Name of the OpenAI embedding model to use
                   (default: "text-embedding-3-small").
                 - dimensions (int | None, optional):
-                  Dimensionality of the embedding vectors.
-                  Required if model is not recognized
+                  Dimensionality of the embedding vectors,
+                  if different from the model's default
                   (default: None).
-                - use_dimensions_parameter (bool, optional):
-                  Whether to use the dimensions parameter
-                  in the OpenAI embedding API call.
-                  The model must support configurable dimensions
-                  if this is set to True.
-                  This must be set to True for configured dimensions
-                  to take effect.
-                  (default: False).
                 - base_url (str, optional):
                   Base URL of the OpenAI embedding model to use.
                 - metrics_factory (MetricsFactory, optional):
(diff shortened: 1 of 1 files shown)


src/memmachine/common/embedder/openai_embedder.py
```diff
@@ -111,6 +119,11 @@ def __init__(self, config: dict[str, Any]):
                     )
 
         self._dimensions = dimensions
+        use_dimensions_parameter = config.get("use_dimensions_parameter", True)
```
malatewang: This extra parameter makes the configuration more complicated. We can probe the dimension size instead.

malatewang (changes requested):
Make change according to our discussion.

src/memmachine/common/embedder/openai_embedder.py
```diff
@@ -69,46 +69,43 @@ def __init__(self, config: dict[str, Any]):
...
-                    raise ValueError(
-                        "Dimensions for text-embedding-3-large "
-                        "cannot be greater than 3072"
-                    )
-            case "text-embedding-ada-002":
-                if dimensions != 1536:
-                    raise ValueError(
-                        "Dimensions for text-embedding-ada-002 must be exactly 1536"
-                    )
+            if len(response.data[0].embedding) != dimensions:
```
malatewang, line 105: Should be: len(response.data[0].embedding) < dimensions ?

src/memmachine/common/embedder/openai_embedder.py
```diff
@@ -69,46 +69,43 @@ def __init__(self, config: dict[str, Any]):
...
+            dimensions = len(response.data[0].embedding)
+            self._use_dimensions_parameter = False
+        else:
+            if not isinstance(dimensions, int):
+                raise TypeError("Dimensions must be an integer")
+            if dimensions <= 0:
+                raise ValueError("Dimensions must be positive")
+
+            # Validate dimensions by embedding a dummy string.
+            try:
```
malatewang, line 91: Should we just call the API. If exception happens, fails the initialization. Then, check if dimension is provided in the configuration?

malatewang (changes requested):

src/memmachine/common/embedder/openai_embedder.py
edwinyyyu: It's confusing to the user if they provide a dimensions that is never used. Also, if they want to customize dimensions, we must check the user-provided dimensions anyway.

src/memmachine/common/embedder/openai_embedder.py
edwinyyyu: No, we care about an exact comparison because it is possible that the user provides a number of dimensions that is too large, and they don't know about it. We want to invalidate the configuration in that case.

malatewang: > No, we care about an exact comparison because it is possible that the user provides a number of dimensions that is too large, and they don't know about it. We want to invalidate the configuration in that case.

We need to check the user's configuration. I think the logic should be:
if self._user_dimensional_parameter:
      if len(response.data[0].embedding) > dimensions:
           raise exception
else:
      if len(response.data[0]).embedding) != dimensions:
          raise exception

malatewang (approved):

svetly-todorov:
Did some smoke tests. Dimensions larger than the default fail as expected:
```
  File "/app/.venv/lib/python3.12/site-packages/memmachine/common/embedder/openai_embedder.py", line 106, in __init__
    raise ValueError(
ValueError: Invalid dimensions 2000 for model text-embedding-3-small

ERROR:    Application startup failed. Exiting.
```

Dimensions <= default works as expected:
```
svetly@sr2:~/memmachine$ docker container logs ca24935d2555
Syncing schema to {'host': 'postgres', 'port': '5432', 'user': 'memmachine', 'password': '****', 'database': 'memmachine'}
Re-initializing ...
INFO:     Started server process [73]
INFO:     Waiting for application startup.
INFO    HTTP Request: POST https://api.openai.com/v1/embeddings "HTTP/1.1 200 OK"
INFO    StreamableHTTP session manager started
[...]
```

Negative dimensions fail:
```
  File "/app/.venv/lib/python3.12/site-packages/memmachine/common/embedder/openai_embedder.py", line 88, in __init__
    raise ValueError("Dimensions must be positive")
ValueError: Dimensions must be positive

ERROR:    Application startup failed. Exiting.
```

PR #360, opened 2025-10-21
Pass user/group/session ID to `language_model`, to enable per-user metrics tracking

svetly-todorov:
### Purpose of the change

Expose per-user token/openAI latency data in the existing `<host>:<port>/metrics` endpoint.

### Description

All of the instrumentation necessary for this already existed in `openai_language_model.py` and its sibling files. The only missing piece was passing the user/group/session IDs, which exist at the top-level `app.py`, down to the `<counter/gauge>.increment()` calls in the language_model files. To accomplish this, added some default fields in `self._user_metrics_labels` that are manually updated before each call to `generate_response()`.

### Fixes/Closes

N/A

### Type of change

- [X] New feature (non-breaking change which adds functionality)

### How Has This Been Tested?

- [X] Manual verification (list step-by-step instructions)

**Test Results:**

Full set of metrics produced after a few sample POSTs to `http://127.0.0.1:8080/v1/memories`:

https://gist.github.com/svetly-todorov/e598d86d1d46c48adf6ef759e9c42bdb

Screenshot showing the input token counts for two different users:

Screenshot 2025-10-21 at 1 20 51 PM

### Checklist

- [X] I have signed the commit(s) within this pull request
- [X] I have performed a self-review of my own code
- [X] I have commented my code
- [X] I have made corresponding changes to the documentation

### Maintainer Checklist

- [ ] Reviewed the code

### Further comments

Putting this PR up partly to request comments. From recent discussions it seems that the user/group/agent/session ID paradigm may not persist in the core layer of `memmachine`, in which case, this will need to change significantly. Is this worth pursuing?

### DIFF
--- a/src/memmachine/common/language_model/language_model.py
+++ b/src/memmachine/common/language_model/language_model.py
@@ -3,13 +3,27 @@ Abstract base class for a language model.
 """
 
 from abc import ABC, abstractmethod
+from dataclasses import fields
 from typing import Any
+
+from memmachine.common.data_types import SessionData
 
 
 class LanguageModel(ABC):
     """
     Abstract base class for a language model.
     """
+
+    def __init__(self):
+        """Initialize the language model with an empty metrics labels dict."""
+        # Instance variable for metrics labels, managed by base class
+        self._user_metrics_labels: dict[str, str] = {}
+
+    def _set_session_metrics_labels(self):
+        """Append the SessionData field names to the user metrics labels."""
+        self._user_metrics_labels.update(
+            {session_data_field.name: "" for session_data_field in fields(SessionData)}
+        )
 
     @abstractmethod
     async def generate_response(
@@ -19,6 +33,7 @@ class LanguageModel(ABC):
         tools: list[dict[str, Any]] | None = None,
         tool_choice: str | dict[str, str] | None = None,
         max_attempts: int = 1,
+        session_data: SessionData | None = None,
     ) -> tuple[str, Any]:
         """
         Generate a response based on the provided prompts and tools.
--- a/src/memmachine/profile_memory/profile_memory.py
+++ b/src/memmachine/profile_memory/profile_memory.py
@@ -19,6 +19,7 @@ from pydantic import BaseModel
 from memmachine.common.data_types import ExternalServiceAPIError
 from memmachine.common.embedder.embedder import Embedder
 from memmachine.common.language_model.language_model import LanguageModel
+from memmachine.common.utils import isolations_to_session_data
 
 from .prompt_provider import ProfilePrompt
 from .storage.storage_base import ProfileStorageBase
@@ -564,7 +565,12 @@ class ProfileMemory:
         # Use chain-of-thought to get entity profile update commands.
         try:
             response_text, _ = await self._model.generate_response(
-                system_prompt=self._update_prompt, user_prompt=user_prompt
+                system_prompt=self._update_prompt,
+                user_prompt=user_prompt,
+                session_data=isolations_to_session_data(
+                    isolations=isolations,
+                    default_user_id=user_id,
+                ),
             )
         except (ExternalServiceAPIError, ValueError, RuntimeError) as e:
             logger.error("Eror when update profile: %s", str(e))
@@ -694,13 +700,17 @@ class ProfileMemory:
                 user_id, thresh=5, isolations=isolations
             )
             await asyncio.gather(
-                *[self._deduplicate_profile(user_id, section) for section in s]
+                *[
+                    self._deduplicate_profile(user_id, section, isolations)
+                    for section in s
+                ]
             )
 
     async def _deduplicate_profile(
         self,
         user_id: str,
         memories: list[dict[str, Any]],
+        isolations: dict[str, bool | int | float | str] | None = None,
     ):
         """
         sends a list of features to an llm to consolidated
@@ -709,6 +719,10 @@ class ProfileMemory:
             response_text, _ = await self._model.generate_response(
                 system_prompt=self._consolidation_prompt,
                 user_prompt=json.dumps(memories),
+                session_data=isolations_to_session_data(
+                    isolations=isolations,
+                    default_user_id=user_id,
+                ),
             )
         except (ExternalServiceAPIError, ValueError, RuntimeError) as e:
             logger.error("Model Error when deduplicate profile: %s", str(e))
--- a/src/memmachine/common/data_types.py
+++ b/src/memmachine/common/data_types.py
@@ -2,6 +2,9 @@
 Common data types for MemMachine.
 """
 
+from dataclasses import dataclass
+from typing import Iterable
+
 
 class ExternalServiceAPIError(Exception):
     """
@@ -9,3 +12,24 @@ class ExternalServiceAPIError(Exception):
     """
 
     pass
+
+
+@dataclass
+class SessionData:
+    """Class for session data."""
+
+    group_id: str
+    user_id: list[str]
+    session_id: str
+    agent_id: list[str]
+
+    def generate_all_combinations(self) -> Iterable[dict[str, str]]:
+        """Generate all combinations of user_id, agent_id, group_id, and session_id."""
+        for user_id in self.user_id:
+            for agent_id in self.agent_id:
+                yield {
+                    "group_id": self.group_id,
+                    "session_id": self.session_id,
+                    "user_id": user_id,
+                    "agent_id": agent_id,
+                }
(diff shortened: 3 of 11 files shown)


src/memmachine/common/language_model/language_model.py
```diff
@@ -11,6 +11,35 @@ class LanguageModel(ABC):
...
+        """Initialize the language model with an empty metrics labels dict."""
+        # Instance variable for metrics labels, managed by base class
+        self._user_metrics_labels: dict[str, str] = {}
+
+    def set_default_metrics_labels(
+        self,
+        user_id: str = "",
+        agent_id: str = "",
+        group_id: str = "",
+        session_id: str = "",
```
edwinyyyu: @malatewang @jealous 
I believe this should not be in LanguageModel.
Such labels should be provided by layers above.

edwinyyyu (changes requested):

src/memmachine/common/language_model/language_model.py
svetly-todorov: I agree, it's not necessary to hardcode these here. This can be removed, and callers of `set_default_metrics_labels` can just as easily set `_user_metrics_labels` themselves.

src/memmachine/profile_memory/profile_memory.py
```diff
@@ -694,18 +703,31 @@ async def _update_user_profile_think(
...
+                metrics_labels = extract_metrics_labels_from_isolations(
+                    isolations=isolations,
+                    default_user_id=user_id,
+                )
+                # Set the default metrics labels before generating response
+                self._model.set_default_metrics_labels(**metrics_labels)
+
             response_text, _ = await self._model.generate_response(
                 system_prompt=self._consolidation_prompt,
                 user_prompt=json.dumps(memories),
```
o-love: Consider that the model is being used concurrently between all the different memories and REST api requests.

src/memmachine/profile_memory/profile_memory.py
```diff
@@ -694,18 +703,31 @@ async def _update_user_profile_think(
...
+                    self._deduplicate_profile(user_id, section, isolations)
+                    for section in s
+                ]
             )
 
     async def _deduplicate_profile(
         self,
         user_id: str,
         memories: list[dict[str, Any]],
+        isolations: dict[str, bool | int | float | str] | None = None,
```
o-love, line 713: Also of note we are planning in the next weeks to abstract out the user id, group id, etc. from the memory layers. 

It might be a nicer implementation for the future and for the cucurrency issue to have the metric tracker be a wrapper around the LLM that gets initialized with the ids. 
So that the LanguageModel that gets passed is specified with the metric metadata already.

svetly-todorov: > planning in the next weeks to abstract out the user id, group id, etc.

I was concerned about this too. Shu mentioned as much in the platform meeting yesterday.

> It might be a nicer implementation for the future and for the cucurrency issue to have the metric tracker be a wrapper around the LLM that gets initialized with the ids.

Does this mean that every unique set of IDs will have its own LanguageModel object?

src/memmachine/profile_memory/profile_memory.py
svetly-todorov: Right, this is something I had a question about. The current paradigm only works if there is a 1:1 correspondence between `set_default_metrics_labels` and `generate_response` calls, and if `generate_response` is called concurrently after the metrics labels are set. But it seems like that might not necessarily be the case. i.e. we might have `set_default_metrics_labels(USER A)` -> `generate_response(USER B)` -> `set_default_metrics_labels(USER B)` -> `generate_response(USER A)` -- things get interlaced.

svetly-todorov: If the correspondence doesn't exist then I need to edit the params of `generate_response` so that the IDs get passed in with the prompt. I didn't want to touch the function signature so I didn't do it in this first pass.

edwinyyyu:
Potentially we can refactor the metrics injection so that things get the metrics they want directly, but it involves more wiring.

src/memmachine/profile_memory/profile_memory.py
o-love: > Does this mean that every unique set of IDs will have its own LanguageModel object?

Yes, the idea is that each unique set would have its own reference to a Lanugauge model. 
But in our implementation without metrics all the references would be to the same underlying object

So something like the following is what I was thinking.
Where they still use the same underlying client and the metric tracking is wrapped around that.

```
# implements LanguageModel, wrapping another
class MetricsLanguageModel(LanguageModel):
    def __init__(
            self, 
            llm: LanguageModel, 
            metrics: dict[str, Any],
    )
```

src/memmachine/profile_memory/profile_memory.py
o-love: > we might have set_default_metrics_labels(USER A) -> generate_response(USER B) -> set_default_metrics_labels(USER B) -> generate_response(USER A) -- things get interlaced.

Correct. We share the same OpenAI connection between everyone, so when things are running concurrently there isn't any garentee on the order of use I would say.

src/memmachine/profile_memory/profile_memory.py
svetly-todorov: I see...

I'll ponder this

svetly-todorov: If we stick to having Prometheus counters right up next to our openAI invocations, then I agree, this design makes the most sense.

src/memmachine/profile_memory/profile_memory.py
svetly-todorov: OK. Then this current design has to go. Either we do the MetricsLanguageModel as described below, or we pass a dictionary of labels to each `generate_response` invocation and eliminate this race.

svetly-todorov:
> Potentially we can refactor the metrics injection so that things get the metrics they want directly, but it involves more wiring.

What do you have in mind? @edwinyyyu 

At first, I didn't know that there were Prometheus counters already integrated in the `language_model` files. I thought the instrumentation would happen somewhere higher on the stack, closer to `app.py`. My thought was to propagate a struct containing metrics from `generate_response()` to the higher-level calls. I still kind of think this is a good idea, because as it stands, anyone who imports `memmachine` to their Python project has to run a bunch of Prometheus instrumentation under the hood... 

But it turned out that the counters/gauges were there. Therefore the task is to get the high-level session IDs down to the counter increments.

If we decide to keep the Prometheus dependency, then the solution that Oscar described above makes sense. Each LanguageModel can contain a little bit of session info, paired with a pointer to an underlying, object. When a LanguageModel calls `generate_response`, it'll update its counters with labels corresponding to that session info.

src/memmachine/common/language_model/language_model.py
jealous: Could you just pass a `SessionData`?  It's easier to change if we add more fields in this class.

edwinyyyu:
> > Potentially we can refactor the metrics injection so that things get the metrics they want directly, but it involves more wiring.
> 
> What do you have in mind? @edwinyyyu
> 
> At first, I didn't know that there were Prometheus counters already integrated in the `language_model` files. I thought the instrumentation would happen somewhere higher on the stack, closer to `app.py`. My thought was to propagate a struct containing metrics from `generate_response()` to the higher-level calls. I still kind of think this is a good idea, because as it stands, anyone who imports `memmachine` to their Python project has to run a bunch of Prometheus instrumentation under the hood...
> 
> But it turned out that the counters/gauges were there. Therefore the task is to get the high-level session IDs down to the counter increments.
> 
> If we decide to keep the Prometheus dependency, then the solution that Oscar described above makes sense. Each LanguageModel can contain a little bit of session info, paired with a pointer to an underlying, object. When a LanguageModel calls `generate_response`, it'll update its counters with labels corresponding to that session info.

I defined interfaces for counters, gauges, etc. in MetricsFactory. Any implementations of MetricsFactory will provide implementations of those interfaces.

Right now the LanguageModel and some other classes optionally take the whole MetricsFactory as a dependency and create their own metrics objects.

I was suggesting that they take the individual metric objects directly as dependencies, so that the metrics objects are configurable from the outside, maybe with some contextvar trick.

src/memmachine/common/language_model/language_model.py
svetly-todorov: That's a good idea. I can move `SessionData` into `data_types.py` and pass it in.

malatewang: Is this the correct way to collect metrics for individual user? The same model instance can be shared by different users.

svetly-todorov: No, I don't think so anymore either. Based on what Oscar said below:

> > we might have set_default_metrics_labels(USER A) -> generate_response(USER B) -> set_default_metrics_labels(USER B) -> generate_response(USER A) -- things get interlaced.

> Correct. We share the same OpenAI connection between everyone, so when things are running concurrently there isn't any garentee on the order of use I would say.

I think the right way to implement this pattern is to to pass the SessionData into generate_response, and use the info in it to set the counters

svetly-todorov:
Closing in advance of semantic memory refactor

PR #371, opened 2025-10-23
v0.2 prep: Config changes and semantic memory changes

o-love:
Working branch to stage v0.2

### DIFF
--- a/src/memmachine/episode_store/episode_sqlalchemy_store.py
+++ b/src/memmachine/episode_store/episode_sqlalchemy_store.py
@@ -54,7 +54,7 @@ class Episode(BaseEpisodeStore):
     produced_for_id = mapped_column(String, nullable=True)
     episode_type = mapped_column(
         SAEnum(EpisodeType, name="episode_type"),
-        nullable=True,
+        default=EpisodeType.MESSAGE,
     )
 
     json_metadata = mapped_column(


src/memmachine/common/language_model/language_model.py
```diff
@@ -11,6 +11,44 @@ class LanguageModel(ABC):
...
+                (default: None).
+            user_prompt (str | None, optional):
+                The user prompt containing the main input
+                (default: None).
+            max_attempts (int, optional):
+                The maximum number of attempts to make before giving up
+                (default: 1).
+            output_format (Any):
+                The expected output format or schema for parsing the response.
+                Implementation-specific (e.g., Pydantic model, JSON schema)
```
edwinyyyu, line 35: Must it be implementation-specific? If so, is it possible that this goes in config instead of at call site?

src/memmachine/common/language_model/language_model.py
```diff
@@ -11,6 +11,44 @@ class LanguageModel(ABC):
...
     """
 
+    @abstractmethod
+    async def generate_parsed_response(
+        self,
+        output_format: Any,
+        system_prompt: str | None = None,
+        user_prompt: str | None = None,
+        max_attempts: int = 1,
+    ):
```
edwinyyyu: NIT: return type annotation

src/memmachine/semantic_memory/semantic_memory.py
```diff
@@ -0,0 +1,286 @@
...
+
+This module contains the `SemanticMemoryManager` class, which is the central component
+for creating, managing, and searching feature sets based on their
+conversation history. It integrates with language models for intelligent
+information extraction and a vector database for semantic search capabilities.
+"""
+
+import asyncio
+import logging
+from typing import Any, Optional
```
edwinyyyu: Existing code uses `| None`. Should we enforce one style or the other?

src/memmachine/semantic_memory/semantic_model.py
```diff
@@ -0,0 +1,135 @@
+from dataclasses import dataclass
+from datetime import datetime
+from enum import Enum
+from types import ModuleType
+from typing import Any, Optional, Protocol, Tuple, runtime_checkable
```
edwinyyyu: typing.Tuple is deprecated.
Use just tuple instead.
https://docs.python.org/3/library/typing.html#typing.Tuple

src/memmachine/semantic_memory/storage/alembic_pg/versions/3d6aaebdc526_sync_sqlalchemy_schema.py
```diff
@@ -0,0 +1,165 @@
+"""sync_sqlalchemy_schema
+
+Revision ID: 3d6aaebdc526
+Revises: 001
+Create Date: 2025-11-04 20:32:38.622715
+
+"""
+
+from typing import Sequence, Union
```
edwinyyyu: typing.Sequence is deprecated.
Use collections.abc.Sequence instead.
https://docs.python.org/3/library/typing.html#typing.Sequence

Existing code uses `|` instead of Union. Should we enforce one style or the other?

tests/memmachine/semantic_memory/storage/alembic_pg/test_sqlalchemy_vs_alembic.py
```diff
@@ -0,0 +1,185 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Mapping
```
edwinyyyu: typing.Mapping is deprecated.
Use collections.abc.Mapping instead.
https://docs.python.org/3/library/typing.html#typing.Mapping

tests/memmachine/semantic_memory/test_semantic_ingestion.py
```diff
@@ -0,0 +1,343 @@
+"""Tests for the ingestion service using the in-memory semantic storage."""
+
+from typing import Iterable
```
edwinyyyu: typing.Iterable is deprecated.
Use collections.abc.Iterable instead.
https://docs.python.org/3/library/typing.html#typing.Iterable

src/memmachine/semantic_memory/semantic_llm.py
```diff
@@ -0,0 +1,102 @@
...
+def _features_to_llm_format(
+    features: list[SemanticFeature],
+) -> dict[str, dict[str, str]]:
+    structured_features: dict[str, dict[str, str]] = {}
+
+    for feature in features:
+        if feature.tag not in structured_features:
+            structured_features[feature.tag] = {}
+
+        structured_features[feature.tag][feature.feature] = feature.value
```
edwinyyyu:
```suggestion
        structured_features.setdefault(feature.tag, {})[feature.feature] = feature.value
```

src/memmachine/semantic_memory/semantic_llm.py
```diff
@@ -0,0 +1,102 @@
...
+    features: list[SemanticFeature],
+    message_content: str,
+    model: InstanceOf[LanguageModel],
+    update_prompt: str,
+) -> list[SemanticCommand]:
+    user_prompt = (
+        "The old feature set is provided below:\n"
+        "<OLD_PROFILE>\n"
+        "{feature_set}\n"
+        "</OLD_PROFILE>\n"
```
edwinyyyu, line 56: I see remnants of the profile concept. Should these be changed or is profile easier to understand for the model?

src/memmachine/semantic_memory/semantic_llm.py
```diff
@@ -0,0 +1,102 @@
...
+        output_format=_SemanticFeatureUpdateRes,
+    )
+
+    validated_output = TypeAdapter(_SemanticFeatureUpdateRes).validate_python(
+        parsed_output
+    )
+    return validated_output.commands
+
+
+class LLMReducedFeature(BaseModel):
```
edwinyyyu, line 76: This data structure looks like it is more general and can be shared.
Why not just `ReducedFeature` or make this `Feature` and have what is currently `SemanticFeature` contain an instance of this?

src/memmachine/server/app_semantic.py
edwinyyyu, line 1: What's the point of this file?

src/memmachine/semantic_memory/semantic_model.py
```diff
@@ -0,0 +1,135 @@
...
+    DELETE = "delete"
+
+
+class SemanticCommand(BaseModel):
+    """Normalized instruction emitted by the LLM to mutate semantic features."""
+
+    command: SemanticCommandType
+    feature: str
+    tag: str
+    value: str
```
edwinyyyu, line 31: Any time these three appear in the same BaseModel, probably they can be replaced by an instance of a BaseModel that includes all three.

src/memmachine/semantic_memory/semantic_model.py
```diff
@@ -0,0 +1,135 @@
...
+
+@dataclass
+class SemanticPrompt:
+    """Pair of prompt templates driving update and consolidation LLM calls."""
+
+    update_prompt: str
+    consolidation_prompt: str
+
+    @staticmethod
+    def load_from_module(prompt_module: ModuleType):
```
edwinyyyu: Why not plain text?
Do we not allow customization of what prompts are available without going into the package code?

src/memmachine/semantic_memory/semantic_model.py
```diff
@@ -0,0 +1,135 @@
...
+
+        citations: Optional[list[HistoryMessage]] = None
+        id: Optional[int] = None
+        other: Optional[dict[str, Any]] = None
+
+    set_id: Optional[str] = None
+    type: str
+    tag: str
+    feature: str
+    value: str
```
edwinyyyu, line 74: Having a "feature" field in a SomeNameThatEndsInFeature may be confusing in some situations.

src/memmachine/semantic_memory/semantic_model.py
```diff
@@ -0,0 +1,135 @@
...
+    metadata: Metadata = Metadata()
+
+    @staticmethod
+    def group_features(
+        features: list["SemanticFeature"],
+    ) -> dict[Tuple[str, str, str], list["SemanticFeature"]]:
```
edwinyyyu: I don't see this used anywhere.
Can you show an example of when this would be useful?

src/memmachine/semantic_memory/semantic_model.py
```diff
@@ -0,0 +1,135 @@
...
+
+            if key not in grouped_features:
+                grouped_features[key] = []
+
+            grouped_features[key].append(f)
+
+        return grouped_features
+
+
+class SemanticType(BaseModel):
```
edwinyyyu: "Type" is overloaded and it was harder to tell what it meant when I first saw this name in the other files. Maybe "Category" is better.

src/memmachine/semantic_memory/semantic_session_resource.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def role_profile_id(self) -> str | None:
+        raise NotImplementedError
+
+
+@runtime_checkable
+class SessionIdIsolationTypeChecker(Protocol):
+    """Protocol for determining the isolation type associated with a set_id."""
+
+    def set_id_isolation_type(self, _id: str) -> IsolationType:
```
edwinyyyu: Is the parameter name _id incomplete? Why does it start with an underscore?

src/memmachine/semantic_memory/semantic_session_resource.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def session_id(self) -> str | None:
+        raise NotImplementedError
+
+    def role_profile_id(self) -> str | None:
+        raise NotImplementedError
+
+
+@runtime_checkable
+class SessionIdIsolationTypeChecker(Protocol):
```
edwinyyyu, line 35: Why do we need this Protocol? Do we expect more implementations in the future? Lazy evaluation?

src/memmachine/semantic_memory/semantic_ingestion.py
```diff
@@ -0,0 +1,278 @@
...
+                        set_ids=[set_id],
+                        type_names=[type_name],
+                        feature_names=[command.feature],
+                        tags=[command.tag],
+                    )
+
+                case _:
+                    logger.error("Command with unknown action: %s", command.command)
+
+    async def _consolidate_set_memories_if_applicable(
```
edwinyyyu, line 196: Why "if_applicable"? I don't see obvious applicability logic here.

src/memmachine/semantic_memory/semantic_memory.py
```diff
@@ -0,0 +1,286 @@
...
+
+        self._is_shutting_down = True
+        await self._ingestion_task
+
+    async def search(
+        self,
+        set_ids: list[str],
+        query: str,
+        *,
+        min_distance: float = 0.7,
```
edwinyyyu, line 91: min_distance depends on model. It may be confusing for users if they get 0 results because they used a different model.

src/memmachine/semantic_memory/semantic_session_resource.py
o-love: Sort of. In general all my code tries to depend on interfaces rather than on concrete classes

The reason I use a protocol rafter than an abstract class of because I'm having the class define its own direct dependencies. Rather than defining an entire interface. 

I also want all the data isolation code to be decently modular and at various levels.

src/memmachine/semantic_memory/semantic_memory.py
o-love: We can decide on a consistent style and modify the linter rules to enforce and auto apply them

src/memmachine/semantic_memory/semantic_model.py
o-love: Yeah, Category sounds better. I'll play around with what could be a better name.

src/memmachine/semantic_memory/semantic_model.py
o-love: I found these functions useful when testing or actually using the memory.

A lot of times I wanted to do an operation on each group of tags, set_id...

It's sort of a utility function for anyone who wants to group a list of features

src/memmachine/semantic_memory/semantic_model.py
o-love: mmm. Maybe rename to `name` or `feature_name`

src/memmachine/semantic_memory/storage/storage_base.py
```diff
@@ -0,0 +1,209 @@
...
+        end_time: Optional[AwareDatetime] = None,
+    ):
+        raise NotImplementedError
+
+    @abstractmethod
+    async def get_history_messages(
+        self,
+        *,
+        set_ids: Optional[list[str]] = None,
+        k: Optional[int] = None,
```
edwinyyyu: k is not meaningful.
```suggestion
        limit: Optional[int] = None,
```

src/memmachine/semantic_memory/util/lru_cache.py
```diff
@@ -55,6 +55,13 @@ def _add_to_front(self, node: Node) -> None:
             self.head.next.prev = node
         self.head.next = node
 
+    def clean(self) -> None:
```
edwinyyyu:
```suggestion
    def clear(self) -> None:
```

src/memmachine/semantic_memory/semantic_model.py
o-love: I'd prefer plain text only. 

I have this to legacy support the way we used to define prompts.
ProfileMemory only supported defining prompts as modules, without the option of passing strings.

And our current config accepts a module path to be loaded.

Note that this factory just makes the loading of those old prompts easier. But I'd prefer if people used SemanticPrompt directly going forward.

If you look at the modifications I made to the prompts as part of this PR I define SemanticPrompt/SemanticType for them. To hopefully have the future config use them.

I also think that once we allow dynamic prompt addition/modification people will use the module definitions for prompts less.

src/memmachine/semantic_memory/semantic_model.py
o-love: Nice. Something additional to add to the linter

src/memmachine/server/prompt/profile_prompt.py
edwinyyyu, line 1: Consider allowing prompt writers to provide example inputs and outputs in a structured way so that there is no risk of typos or a lack of understanding of the schema causing the model to output undesired results.

src/memmachine/common/language_model/language_model.py
o-love: I think swithcing to Pydantic only is reasonable


The goal is to limit the output structure is done by the llm call.
One call might want

```
class User(BaseModel):
  name: str
  age: int
```

And just after make a call wanting the format to be

```
class Pet(BaseModel)
   name: str
   owner: str
   age: int
```


And openai and ollama really do a good job of forcing the llm to produce that output. (I've never had them fail yet) vs our old way of specifying the structure ourselves in the prompt which did ocassionally fail. Especially on smaller and older models

src/memmachine/semantic_memory/semantic_ingestion.py
o-love: It's being filtered by

```
                tag_threshold=self._consolidation_threshold,
```


Which indicates the minimum number of features with a tag that need to be present before returning any features of that tag.

src/memmachine/common/language_model/openai_chat_completions_language_model.py
```diff
@@ -119,6 +119,52 @@ def __init__(self, config: dict[str, Any]):
...
+
+        start_time = time.monotonic()
+
+        try:
+            response = await self._client.with_options(
+                max_retries=max_attempts
+            ).chat.completions.parse(
+                model=self._model,  # type: ignore[arg-type]
+                messages=input_prompts,  # type: ignore[arg-type]
+                response_format=output_format,
```
malatewang, line 144: The structured output is model and model version dependent. How can the user know which API to call?

o-love: How each model provider does it is provider dependent. But as far as I could find there isn't any change in behavior inside a specific model provider.

In the case of OpenAI compatible, all the work with `chat/completions` is being done by the local openai client and not the end server.

src/memmachine/semantic_memory/semantic_llm.py
o-love: Not sure, right now the only real prompt is still the profile one. When I start adding new prompts and modifying existing ones I'll spend some time testing and looking into this.

src/memmachine/semantic_memory/semantic_memory.py
o-love: Not sure how else to do it. 
Maybe best to just not have a default and depend on a default top_k

o-love:
I think the comments left can be left for future PRs if they cause conflict.
Mostly NITs that I'm not fully sure how I prefer to address

src/memmachine/semantic_memory/semantic_memory.py
edwinyyyu: I use an integer limit on number of results. I think it's also a bit safer that way.

src/memmachine/server/prompt/profile_prompt.py
edwinyyyu: Could be future feature.

src/memmachine/common/embedder/amazon_bedrock_embedder.py
```diff
@@ -135,14 +138,15 @@ async def _embed(
...
+        embeddings = []
         sleep_seconds = 1
         for attempt in range(1, max_attempts + 1):
             logger.debug(
                 "[call uuid: %s] "
                 "Attempting to create embeddings using %s Amazon Bedrock model: "
                 "on attempt %d with max attempts %d",
                 embed_call_uuid,
-                self._model_id,
+                self.model_id,
```
edwinyyyu, line 131: I think this breaks the code.
```suggestion
                self._model_id,
```

src/memmachine/common/embedder/amazon_bedrock_embedder.py
```diff
@@ -173,7 +177,7 @@ async def _embed(
                     type(e).__name__,
                 )
                 await asyncio.sleep(
-                    min(sleep_seconds, self._max_retry_interval_seconds)
+                    min(sleep_seconds, self._conf.max_retry_interval_seconds)
```
edwinyyyu: _conf is not defined anywhere.
```suggestion
                    min(sleep_seconds, self._max_retry_interval_seconds)
```

src/memmachine/common/embedder/amazon_bedrock_embedder.py
```diff
@@ -135,14 +138,15 @@ async def _embed(
 
         start_time = time.monotonic()
 
+        embeddings = []
```
edwinyyyu, line 123: I don't think this is ever desirable. It should either return an embedding vector of the correct size or fail.

src/memmachine/common/ralational_table_storage/relational_data_store.py
edwinyyyu, line 1: Unused file?

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -409,7 +409,7 @@ async def delete_nodes(
     ):
         async with self._semaphore:
             await self._driver.execute_query(
-                """
+                """delete
```
edwinyyyu: This is a database query. This change will break it.
```suggestion
                """
```

src/memmachine/episodic_memory/data_types.py
```diff
@@ -15,6 +15,14 @@ class ContentType(Enum):
     # Other content types like 'vector', 'image' could be added here.
 
 
+class EpisodeType(Enum):
+    """Enumeration for the type of an Episode."""
+
+    MESSAGE = "message"
+    THOUGHT = "thought"
+    ACTION = "action"
```
edwinyyyu: Do not include until we can process them.
Including a comment like for ContentType is good enough.
```suggestion
```

src/memmachine/episodic_memory/data_types.py
```diff
@@ -92,7 +100,11 @@ class Episode:
 
     uuid: UUID
     """A unique identifier (UUID) for the episode."""
-    episode_type: str
+    sequence_num: int
+    """Sequence number of the Episode"""
```
edwinyyyu: How is this used?

src/memmachine/episodic_memory/data_types.py
```diff
@@ -103,12 +115,10 @@ class Episode:
...
     timestamp: datetime
     """The date and time when the episode occurred."""
-    group_id: str
-    """Identifier for the group (e.g., a specific chat room or DM)."""
-    session_id: str
-    """Identifier for the session to which this episode belongs."""
     producer_id: str
     """The identifier of the user or agent that created this episode."""
+    producer_role: str
+    """The role of the producer (e.g., 'HR', 'agent', 'engineer')."""
```
edwinyyyu: Can a producer have multiple roles?
How is this supposed to work?

src/memmachine/common/configuration/database_conf.py
edwinyyyu, line 1: Implementations of different interfaces should not be configured under the same place in the config. What is the benefit of the refactored configuration if this is not the case?

VectorGraphStores should be configured under a key like vector_graph_store.
SemanticStorages should be configured under a key like semantic_storage.

This makes it confusing which storage providers are usable for which memory systems.
This also requires a lot of custom vendor/provider-specific logic.
What happens when it's possible to use Neo4j/Postgres to support multiple types of memories using different interfaces?

src/memmachine/common/configuration/embedder_conf.py
```diff
@@ -0,0 +1,129 @@
+from typing import Self
+
+from pydantic import BaseModel, Field, SecretStr
+
+from memmachine.common.configuration.metrics_conf import WithMetricsFactoryId
+from memmachine.common.data_types import SimilarityMetric
+
+
+class AmazonBedrockEmbedderConfig(BaseModel):
```
edwinyyyu, line 12: Confusing mix of Conf and Config

src/memmachine/common/resource_manager/resource_manager.py
```diff
@@ -0,0 +1,58 @@
...
+            self._conf.model
+        )
+        self._reranker_manager: RerankerManager = RerankerManager(self._conf.reranker)
+        self._metric_factory: dict[str, MetricsFactory] = {"prometheus": PrometheusMetricsFactory}
+
+    def build(self):
+        self._storage_manager.build_all(validate=True)
+        self._embedder_manager.build_all()
+        self._model_manager.build_all()
+        self._reranker_manager.build_all(self._embedder_manager.embedders)
```
edwinyyyu: Fixed-order intialization is not extensible and requires checking this logic/implementing custom logic for every single new resourcce implementation of an interface that depends on another resource.

edwinyyyu:
Obsoleted by #542

PR #402, opened 2025-10-30
Handle the incorrect json response from model.

Kevin-K-W:
### Purpose of the change

When I tried to use AWS Nova Pro as the backend LLM of MemMachine, sometime I found that the model will response **markdown** formatted json string, such as:
````
```json
{ "key": "value" } # the real json
```
````
This will cause the json.loads() function failed.

### Description

This PR add one more CMD into the prompt to force the LLM to output the correct string, but also handle the response json string in case the LLM missing it.

### Fixes/Closes

Fixes #(issue number)

### Type of change

### How Has This Been Tested?

- [X] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### Further comments

The added CMD will increase the input token, I'm not quite sure if it's needed.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -82,7 +82,7 @@ services:
       MCP_BASE_URL: ${MCP_BASE_URL:-http://memmachine:8080}
       GATEWAY_URL: ${GATEWAY_URL:-http://localhost:8080}
       FAST_MCP_LOG_LEVEL: ${FAST_MCP_LOG_LEVEL:-INFO}
-      OPENAI_API_KEY: ${OPENAI_API_KEY}
+      OPENAI_API_KEY: ${OPENAI_API_KEY:-}
       LOG_LEVEL: ${LOG_LEVEL:-INFO}
     volumes:
       - ./configuration.yml:/app/configuration.yml:ro,Z

### CODE REVIEW

File: src/memmachine/profile_memory/profile_memory.py [hunk 1]
```diff
@@ -554,6 +554,31 @@ async def process_messages(messages):
...
+        if response_text.strip().startswith("```json"):
+            # Remove ```json at the beginning
+            response_text = response_text.strip()[7:]
+        elif response_text.strip().startswith("```"):
+            # Remove ``` at the beginning
+            response_text = response_text.strip()[3:]
+
+        if response_text.strip().endswith("```"):
+            # Remove ``` at the end
+            response_text = response_text.strip()[:-3]
```
edwinyyyu, line 577: response_text.strip() is a bit repetitive. I think it may be better to just do it once. Also consider these instead:
https://docs.python.org/3/library/stdtypes.html#str.removeprefix
https://docs.python.org/3/library/stdtypes.html#str.removesuffix

Also, you may want to look at Mem0's impl: https://github.com/mem0ai/mem0/blob/main/mem0/memory/utils.py#L72

Kevin-K-W: Agree.

I find that Gauri's PR: https://github.com/MemMachine/MemMachine/pull/441 seems already covered my changes.

I'll close this PR.

PR #477, opened 2025-11-06
Add an installation script for windows and macos

mwqgithub:
### Purpose of the change

Add an installation script for windows and macos

### Description

This script automates the setup of memmachine within a Python virtual environment and installs Neo4j as its graph database backend.

- On Windows, it downloads and installs OpenJDK along with the Neo4j Community Edition, then configures Neo4j as a Windows service.
- On macOS, it installs Neo4j via Homebrew.

After completing the installation, the script prompts the user to provide an OpenAI API key and automatically generates both a configuration file and a startup script.

### Fixes/Closes

Fixes #(issue number)

### Type of change

### How Has This Been Tested?

- [ ] Unit Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### DIFF
--- a/.github/workflows/docs-checks.yml
+++ b/.github/workflows/docs-checks.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false  # Continue running all matrix jobs even if one fails
       matrix:
-        node-version: [18, 20, 22]  # Test against multiple Node.js versions
+        node-version: [20, 22]  # Test against multiple Node.js versions
 
     steps:
       # Step 1: Checkout the repository code
--- a/docs/examples/mcp.mdx
+++ b/docs/examples/mcp.mdx
@@ -194,9 +194,11 @@ To configure the Claude Desktop App to connect to your MCP server, follow these 
 {
   "mcpServers": {
     "memmachine": {
-      "command": "MEMORY_CONFIG=config.yml MM_USER_ID="user" uv run memmachine-mcp-stdio",
+      "command": "uv run memmachine-mcp-stdio",
       "env": {
-        "PYTHONUNBUFFERED": "1"
+        "PYTHONUNBUFFERED": "1",
+        "MEMORY_CONFIG": "config.yml",
+        "MM_USER_ID": "user"
       }
     }
   }
@@ -227,6 +229,7 @@ MEMORY_CONFIG=config.yml MM_USER_ID="user" memmachine-mcp-stdio
       "args": [
         "mcp-remote",
         "http://localhost:8080/mcp",
+        "--allow-http",
         "--header",
         "user-id: user"
       ]
--- a/docs/open_source/memory_types.mdx
+++ b/docs/open_source/memory_types.mdx
@@ -6,9 +6,10 @@ icon: "memory"
 ## Memory Types
 
 MemMachine works as an intelligent memory layer on top of a user's Agentic LLM experience.  It gathers input via Python APIs in a Memory Instance, then determines if that input is profile, short-term and/or long-term:  
- - **Profile memory** consists of information specific to the user and their experience.  
- - **Short-term memory** consists of episodic memory and summaries of the episodic memory.  A raw episode consists of 1 message. 
- - **Long-term memory** consists of a batch of episodes that is generated from the short-term memories.  The Episode batch is then ranked using a Reranker and deduplicated for the most frequent hits.  
+ - **Semantic memory** consists of information specific to the user and their experience.  
+ - **Episodic memory** consists of both short-term and long-term memory function:
+     - **Short-term memory** consists of episodic memory and summaries of the episodic memory.  A raw episode consists of 1 message. 
+     - **Long-term memory** consists of a batch of episodes that is generated from the short-term memories.  The Episode batch is then ranked using a Reranker and deduplicated for the most frequent hits.  
 
 ## How does MemMachine sort into it's Memory Types?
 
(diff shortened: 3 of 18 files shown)


installation/install_memmachine.py
```diff
@@ -0,0 +1,363 @@
...
+        with open(os.path.join(self.install_dir, "configuration.yml"), "w", encoding="utf-8") as config_file:
+            config_content = CONFIGURATION_TEMPLATE.format(OPENAI_KEY=api_key)
+            config_file.write(config_content)
+
+    def remove_directory(self, dir_path: str):
+        print(f"Removing directory {dir_path}, Proceed? (y/n): ", end="")
+        choice = input().lower()
+        if choice != "y":
+            raise Exception("Operation cancelled by user.")
+        shutil.rmtree(dir_path)
```
jealous, line 187: I would prefer to use python library call instead of shutil.  Can you check if we have a library call to remove the directory.  The library function is supposed to be OS independent.

installation/install_memmachine.py
```diff
@@ -0,0 +1,363 @@
...
+    type: "identity"
+  bm_ranker_id:
+    type: "bm25"
+
+prompt:
+  profile: profile_prompt
+"""
+
+BASH_STARTUP_TEMPLATE = """\
+# Database configuration
```
jealous, line 95: First of all, can we extract those long string to a file in the source folder?  During runtime, we can read them.
Secondly, can we merge bash template and powershell template?  For example, replace certain string like `export` to `$env:`.  The purpose is to reduce the chance of making mistake.  Say if we change any of the field in the future, we only need to change once.

installation/install_memmachine.py
```diff
@@ -0,0 +1,363 @@
...
+NEO4J_ZIP_NAME = "neo4j-community-2025.09.0-windows.zip"
+MEMMACHINE_ZIP_NAME = "MemMachine.zip"
+WINDOWS_RUN_SCRIPT_NAME = "run_memmachine.ps1"
+MACOS_RUN_SCRIPT_NAME = "run_memmachine.sh"
+
+JDK_DIR_NAME = "jdk-21.0.9"
+NEO4J_DIR_NAME = "neo4j-community-2025.09.0"
+
+CONFIGURATION_TEMPLATE = """\
+logging:
```
jealous, line 24: Same here, can we move this to a file?  Because if it's a separate yml file in the source code, we get all the IDE support for that kind of file.  It would be less possible for people to make mistakes when they update those files later.

installation/install_memmachine.py
```diff
@@ -0,0 +1,363 @@
...
+        print(
+            f"Please enter installation directory (default: {self.default_install_dir}): ", end="")
+        user_input = input().strip()
+        if user_input:
+            self.install_dir = user_input
+
+    def install(self):
+        self.ask_install_dir()
+        # make sure the directory does not already exist
+        if os.path.exists(self.install_dir):
```
jealous, line 196: This part could be a separate function to make this `install` function cleaner.

installation/install_memmachine.py
```diff
@@ -0,0 +1,363 @@
...
+        self.environment.install_neo4j()
+
+    def uninstall(self):
+        print("Uninstalling...")
+        self.remove_directory(self.install_dir)
+
+
+class WindowsEnvironment:
+    def download_file(self, url: str, dest: str):
+        curl_path = shutil.which('curl.exe')
```
jealous: Can we use requests to get it instead of curl?  I would like to have as less third party binary dependencies as possible.

installation/install_memmachine.py
```diff
@@ -0,0 +1,363 @@
...
+        if not curl_path:
+            urllib.request.urlretrieve(url, dest)
+            return
+        subprocess.run(["curl.exe", "-L", "-o", dest, url], check=True)
+
+    def extract_zip(self, zip_path: str, extract_to: str):
+        tar_path = shutil.which('tar.exe')
```
jealous: Same here, is it possible to use the `tarfile` module?  The purpose is to reduce system dependency as much as possible.

installation/install_memmachine.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def uninstall(self):
+        print("Uninstalling...")
+        self.environment.uninstall_neo4j_service(self.install_dir)
+        self.remove_directory(self.install_dir)
+
+
+def main():
+    try:
+        print("Installing...")
```
jealous, line 369: Please replace all `print` function with logger calls, use a logger format with timestamp.  The benefits are:
* support log level and timestamps.
* we can switch to debug for more development time info.
* more flexible.

installation/install_memmachine.py
jealous: Forget it.

jealous:
Hi @mwqgithub , I got some comments from Steve.  Please let me know if you can corporate them into this PR or you want to do it in another PR.
* We need to support Linux in this script as well because we want a universal installation script for all the platforms.
* We need some way to set the address/port for the host.  I think default to 0.0.0.0:8080 is good.  We can have an extra question for the host IP/port.

mwqgithub:
> Hi @mwqgithub , I got some comments from Steve. Please let me know if you can corporate them into this PR or you want to do it in another PR.
> 
> * We need to support Linux in this script as well because we want a universal installation script for all the platforms.
> * We need some way to set the address/port for the host.  I think default to 0.0.0.0:8080 is good.  We can have an extra question for the host IP/port.

Hi @jealous , I think I can add them into this PR.

PR #473, opened 2025-11-05
will be split: refactor/rewrite: DeclarativeMemory and VectorGraphStore

edwinyyyu:
### Purpose of the change

1. Get rid of interfaces DerivativeDeriver, DerivativeMutator, RelatedEpisodePostulator in order to trade configurability for performance and simplicity.
2. Implement batch ingestion.
3. Improved indexing
  a. Automatic index creation is only done during ingestion.
  b. There is now a threshold before indexes are created for performance.
  c. There is added synchronization for index creation for performance.
  d. Create indexes in background for improved response time.
4. Improved vector search
  a. fudge factor added for ANN search when the search is filtered
  b. fall back to exact similarity search if there are not enough results
  c. use exact similarity search if indexes are not ready

### Description

Please include a summary of the change and the issue it addresses. Please also include relevant motivation and context. List any dependencies that are required for this change.

### Fixes/Closes

Fixes #(issue number)

### Type of change

[Please delete options that are not relevant.]

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (does not change functionality, e.g., code style improvements, linting)
- [ ] Documentation update
- [ ] Project Maintenance (updates to build scripts, CI, etc., that do not affect the main project)
- [ ] Security (improves security without changing functionality)

### How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

[Please delete options that are not relevant.]

- [ ] Unit Test
- [ ] Integration Test
- [ ] End-to-end Test
- [ ] Test Script (please provide)
- [ ] Manual verification (list step-by-step instructions)

**Test Results:** [Attach logs, screenshots, or relevant output]

### Checklist

[Please delete options that are not relevant.]

- [ ] I have signed the commit(s) within this pull request
- [ ] My code follows the style guidelines of this project (See STYLE_GUIDE.md)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added unit tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

### Maintainer Checklist

[...]

PR #485, opened 2025-11-06
Refactor vector graph store

edwinyyyu:
### Purpose of the change

Performance improvements and enabling future changes.

### Description
This description roughly follows the order of files changed in the GitHub viewer.

Move SimilarityMetric out of memmachine.common.embedder.data_types into memmachine.common.data_types because it is shared between Embedder and VectorGraphStore.

Rename Property --> PropertyValue because it's clearer that Property does not represent a key: value pair.

OrderedPropertyValue now enforces that the values passed to search_directional_nodes are ordered.

Introduce EntityType enum with NODE and EDGE values as a vector graph store data type.
It is not used in the required API, but it is useful for implementations which may want to conditionally do something different for nodes and edges.

Separate out embedding properties from other properties in the Node and Edge types.
This is to support non-Neo4j vector stores in the future, as well as to not conflate other array or vector values with embeddings.

Embedding properties are tied to a SimilarityMetric. This is so that index creation can be done during ingestion. Previously, automatic index creation was considered during search--this design was not chosen because it is fairly likely that a huge amount of data is ingested before performing the first search, and during search the initial index creation would result in high response times.

Added helper functions to mangle and demangle property and embedding names. This is to avoid naming conflicts. This also means that the Node uuid is special: It now represents solely the Node uuid and does not necessarily represent any identifier for the payload, although it is not wrong to share uuid between payload and node, as is done by DeclarativeMemory in this change.

VectorGraphStore API has been updated to use collections to group Nodes and relations to group Edges. This allows us to properly isolate data and create indexes at the right scope. This also allows future implementation of batched operations which improve performance.

VectorGraphStore API now includes get_nodes to allow getting nodes by UUID.

Add the following parameters to Neo4jVectorGraphStore:
- filtered_similarity_search_fudge_factor: When using ANN search, Neo4j does not support pre-filtering so our solution is to increase the search limit, then post-filter.
- exact_similarity_search_fallback_threshold: If the number of ANN search results falls below this threshold, use exact similarity search. There are a couple of cases to consider:
  1. There are few total nodes: Exact similarity search is fast enough and more accurate.
  2. There are many nodes, but the filters are strict: Exact similarity search is okay if there are indexes created, because the exact search allows pre-filtering.
- range_index_hierarchies: indexes can be created for certain hierarchies of filter properties if they are known beforehand
- range_index_creation_threshold: index creation can be expensive so automatic index creation is only done when the number of nodes in a "collection" (having a Neo4j label) or the number of edges having a "relation" (Neo4j relationship type) reaches this quantity
- vector_index_creation_threshold: same reasoning as range_index_creation_threshold--it can be tuned separately

Add a bunch of automatic index creation logic. This involves keeping a rough count of nodes in each collection and caching index state. Vector search will default to exact similarity search when a vector index is being created but isn't ready yet.

Update search_directional_nodes to support searching directionally by multiple properties. The "direction" supports lexicographic ordering by those properties. This allows adding support for searching by sequence numbers in addition to timestamp. It can be useful for when there are multiple nodes with the same timestamp.

Updated PreviousRelatedEpisodePostulator to use new API.
Updated DeclarativeMemory to use new API.

Update tests. Add new tests.

Update VectorGraphStoreBuilder and configuration. Breaks previous Neo4j configuration in YAML.

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

Unit tests pass. LoCoMo ingestion works. LoCoMo search works. Indexes are created when the thresholds are set to 0. Indexes are not created for a small amount of ingestion when the thresholds are 10000.

Threshold for fallback to exact similarity search not tested.
Fudge factor not tested.

- [x] Unit Test
- [x] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### DIFF
--- a/src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
+++ b/src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
@@ -193,18 +193,8 @@ class Neo4jVectorGraphStore(VectorGraphStore):
                 collection
             )
 
-        if (
-            self._collection_node_counts[collection]
-            >= self._range_index_creation_threshold
-        ):
-            asyncio.create_task(
-                self._create_initial_indexes_if_not_exist(
-                    EntityType.NODE,
-                    collection,
-                )
-            )
-
         sanitized_collection = Neo4jVectorGraphStore._sanitize_name(collection)
+        sanitized_embedding_names = set()
 
         query_nodes = []
         for node in nodes:
@@ -228,30 +218,12 @@ class Neo4jVectorGraphStore(VectorGraphStore):
                     )
                 )
 
+                sanitized_embedding_names.add(sanitized_embedding_name)
+
                 query_node_properties[sanitized_embedding_name] = embedding
                 query_node_properties[sanitized_similarity_metric_name] = (
                     similarity_metric.value
                 )
-
-                if (
-                    self._collection_node_counts[collection]
-                    >= self._vector_index_creation_threshold
-                    and Neo4jVectorGraphStore._index_name(
-                        EntityType.NODE,
-                        sanitized_collection,
-                        sanitized_embedding_name,
-                    )
-                    not in self._index_state_cache
-                ):
-                    asyncio.create_task(
-                        self._create_vector_index_if_not_exists(
-                            entity_type=EntityType.NODE,
-                            sanitized_collection_or_relation=sanitized_collection,
-                            sanitized_embedding_name=sanitized_embedding_name,
-                            dimensions=len(embedding),
-                            similarity_metric=similarity_metric,
-                        )
-                    )
 
             query_node: dict[str, PropertyValue | dict[str, PropertyValue]] = {
                 "uuid": str(node.uuid),
@@ -269,6 +241,40 @@ class Neo4jVectorGraphStore(VectorGraphStore):
 
         self._collection_node_counts[collection] += len(query_nodes)
 
+        if (
+            self._collection_node_counts[collection]
+            >= self._range_index_creation_threshold
+        ):
+            asyncio.create_task(
+                self._create_initial_indexes_if_not_exist(
+                    EntityType.NODE,
+                    sanitized_collection,
+                )
+            )
+
+        if (
+            self._collection_node_counts[collection]
+            >= self._vector_index_creation_threshold
+        ):
+            for sanitized_embedding_name in sanitized_embedding_names:
+                if (
+                    Neo4jVectorGraphStore._index_name(
+                        EntityType.NODE,
+                        sanitized_collection,
+                        sanitized_embedding_name,
+                    )
+                    not in self._index_state_cache
+                ):
+                    asyncio.create_task(
+                        self._create_vector_index_if_not_exists(
+                            entity_type=EntityType.NODE,
+                            sanitized_collection_or_relation=sanitized_collection,
+                            sanitized_embedding_name=sanitized_embedding_name,
+                            dimensions=len(embedding),
+                            similarity_metric=similarity_metric,
+                        )
+                    )
+
     async def add_edges(
         self,
         relation: str,
@@ -280,15 +286,8 @@ class Neo4jVectorGraphStore(VectorGraphStore):
             # Not async-safe but it's not crucial if the count is off.
             self._relation_edge_counts[relation] = await self._count_edges(relation)
 
-        if self._relation_edge_counts[relation] >= self._range_index_creation_threshold:
-            asyncio.create_task(
-                self._create_initial_indexes_if_not_exist(
-                    EntityType.EDGE,
-                    relation,
-                )
-            )
-
         sanitized_relation = Neo4jVectorGraphStore._sanitize_name(relation)
+        sanitized_embedding_names = set()
 
         query_edges = []
         for edge in edges:
@@ -312,30 +311,12 @@ class Neo4jVectorGraphStore(VectorGraphStore):
                     )
                 )
 
+                sanitized_embedding_names.add(sanitized_embedding_name)
+
                 query_edge_properties[sanitized_embedding_name] = embedding
                 query_edge_properties[sanitized_similarity_metric_name] = (
                     similarity_metric.value
                 )
-
-                if (
-                    self._relation_edge_counts[relation]
-                    >= self._vector_index_creation_threshold
-                    and Neo4jVectorGraphStore._index_name(
-                        EntityType.EDGE,
-                        sanitized_relation,
-                        sanitized_embedding_name,
-                    )
-                    not in self._index_state_cache
-                ):
-                    asyncio.create_task(
-                        self._create_vector_index_if_not_exists(
-                            entity_type=EntityType.EDGE,
-                            sanitized_collection_or_relation=sanitized_relation,
-                            sanitized_embedding_name=sanitized_embedding_name,
-                            dimensions=len(embedding),
-                            similarity_metric=similarity_metric,
-                        )
-                    )
 
             query_edge = {
                 "uuid": str(edge.uuid),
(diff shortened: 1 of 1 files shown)


edwinyyyu:
Split 1 of #473

src/memmachine/common/vector_graph_store/data_types.py
```diff
@@ -52,3 +66,95 @@ def __eq__(self, other):
...
+    to avoid conflicts with other names.
+
+    Args:
+        property_name (str):
+            The original property name.
+
+    Returns:
+        str: The mangled property name.
+    """
+    return f"property_{property_name}"
```
malatewang: It is better to make the "property_" to be a constant value.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -75,120 +165,276 @@ def __init__(self, params: Neo4jVectorGraphStoreParams):
...
+                query_node_properties[sanitized_embedding_name] = embedding
+                query_node_properties[sanitized_similarity_metric_name] = (
+                    similarity_metric.value
+                )
+
+                if (
+                    self._collection_node_counts[collection]
+                    >= self._vector_index_creation_threshold
+                ):
+                    asyncio.create_task(
```
malatewang, line 268: It is cheaper to check if the index already exists before calling the create_task

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -75,120 +165,276 @@ def __init__(self, params: Neo4jVectorGraphStoreParams):
...
+        relation: str,
+        source_collection: str,
+        target_collection: str,
+        edges: Iterable[Edge],
+    ):
+        if relation not in self._relation_edge_counts.keys():
+            # Not async-safe but it's not crucial if the count is off.
+            self._relation_edge_counts[relation] = await self._count_edges(relation)
+
+        if self._relation_edge_counts[relation] >= self._range_index_creation_threshold:
```
malatewang: I did not see where is the count increased

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -75,120 +165,276 @@ def __init__(self, params: Neo4jVectorGraphStoreParams):
...
+                        }
+                    ),
+                    vector_index_name=vector_index_name,
                 )
-                exact_similarity_search = True
 
-        # ANN search requires a finite limit.
-        if limit is None and not exact_similarity_search:
-            limit = 100_000
+            records = records[:limit]
```
malatewang: Does neo4j guarantee the returned records ordered by similarity score?

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -75,120 +165,276 @@ def __init__(self, params: Neo4jVectorGraphStoreParams):
...
+
+            if (
+                required_properties
+                and len(records)
```
malatewang, line 460: Since the number of nodes for each collection is populated, the search result also can be checked against the collection size to prevent unnecessary fallback.

src/memmachine/common/vector_graph_store/vector_graph_store.py
```diff
@@ -88,46 +106,60 @@ async def search_similar_nodes(
     @abstractmethod
     async def search_related_nodes(
         self,
-        node_uuid: UUID,
-        allowed_relations: Collection[str] | None = None,
+        relation: str,
+        other_collection: str,
```
malatewang, line 110: Should we allow "other_connection" to be empty?

src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
```diff
@@ -142,6 +142,14 @@ def __init__(self, params: DeclarativeMemoryParams):
             )
         )
 
+        self._episode_collection = "Episode"
```
malatewang, line 145: Should make the collection name to include the session.

src/memmachine/common/vector_graph_store/data_types.py
edwinyyyu: Do you mean to make the prefix a named constant?

src/memmachine/common/vector_graph_store/vector_graph_store.py
edwinyyyu: I think it is safer to not allow this for performance reasons. I don't see a use case now where it is necessary to allow arbitrary other collection.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
edwinyyyu: Added.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
edwinyyyu: It is guaranteed according to documentation but they do not document whether it is in descending order. I do not remember the order. I will move the sort and limit to the query for clarity.

src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
edwinyyyu: That's addressed in #492. Should it be implemented here as well?

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
edwinyyyu: Done for vector index. I don't think the performance difference is that big.

src/memmachine/common/vector_graph_store/data_types.py
malatewang: Yes. It is clearer to use a constant as the prefix. And it can help to prevent bug.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
edwinyyyu: If the collection is small enough for this to matter, then typically (unless the limit is set very small) there is no vector index.

The count is only maintained right now for node and edge addition. I think the logic will get more complex if it is also done during search.

src/memmachine/common/vector_graph_store/data_types.py
edwinyyyu: Done. Also done for filterable property names, embedding names, and sanitized Neo4j names.

src/memmachine/common/vector_graph_store/vector_graph_store.py
edwinyyyu: Actually performance should not be impacted much.

I think it's better to keep it for now anyway because the logic is simpler this way and it makes the API more like the other search methods (with the desired collection near the top of the arguments). It also means that we know exactly where the results are coming from.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -75,410 +165,934 @@ def __init__(self, params: Neo4jVectorGraphStoreParams):
...
+                    )
+                )
+
+                query_edge_properties[sanitized_embedding_name] = embedding
+                query_edge_properties[sanitized_similarity_metric_name] = (
+                    similarity_metric.value
+                )
+
+                if (
+                    self._relation_edge_counts[relation]
```
malatewang: Because the edge count is only increased at the end of the function, the optimal way is to only do the check and call the function once.

edwinyyyu: Made a change to address this.

PR #522, opened 2025-11-13
Shared raw history store

o-love:
### Purpose of the change

Centralizes the raw history store to avoid duplication of data between memory storage types.

The central raw data store supports and is tested with SQLite and PostgreSQL.


src/memmachine/history_store/history_model.py
```diff
@@ -0,0 +1,30 @@
+from datetime import datetime
+from typing import Any
+
+from memmachine.episodic_memory.data_types import EpisodeType
+from pydantic import BaseModel
+
+HistoryIdT = str
+
+
+class HistoryMessage(BaseModel):
```
edwinyyyu, line 10: Is there an id for the HistoryMessage?

o-love: Yeah, Each HistoryMessage gets a unique id once it has been written to the DB.

In such a case the optional id in the metadata gets populated

malatewang: I think this should be Episode UUID

src/memmachine/history_store/history_storage.py
```diff
@@ -0,0 +1,67 @@
+from abc import ABC, abstractmethod
+
+from pydantic import AwareDatetime
+
+from memmachine.episodic_memory.data_types import EpisodeType
+from memmachine.history_store.history_model import HistoryIdT, HistoryMessage
+
+
+class HistoryStorage(ABC):
```
malatewang, line 9: Probably we should rename this to be EpisodeStorage?

src/memmachine/history_store/history_storage.py
```diff
@@ -0,0 +1,67 @@
...
+    @abstractmethod
+    async def add_history(
```
malatewang, line 11: Probably just use Episode as the parameter

src/memmachine/history_store/history_model.py
malatewang: Can we just unify this with episode to one data structure?

jealous (approved):

PR #523, opened 2025-11-13
Neo4j semantic store

o-love:
### Purpose of the change

Adds Neo4j backend to SemanticMemory

### DIFF
--- a/src/memmachine/semantic_memory/storage/neo4j_semantic_storage.py
+++ b/src/memmachine/semantic_memory/storage/neo4j_semantic_storage.py
@@ -1,0 +1,925 @@
+"""Neo4j-backed implementation of :class:`SemanticStorageBase`."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import datetime, timezone
+import re
+from typing import Any
+
+import numpy as np
+from neo4j import AsyncDriver
+from pydantic import InstanceOf, validate_call
+
+from memmachine.episode_store.episode_model import EpisodeIdT
+from memmachine.semantic_memory.semantic_model import SemanticFeature
+from memmachine.semantic_memory.storage.storage_base import (
+    FeatureIdT,
+    SemanticStorageBase,
+)
+
+
+def _utc_timestamp() -> float:
+    return datetime.now(timezone.utc).timestamp()
+
+
+@dataclass
+class _FeatureEntry:
+    feature_id: FeatureIdT
+    set_id: str
+    category_name: str
+    tag: str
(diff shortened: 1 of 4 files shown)


src/memmachine/semantic_memory/storage/neo4j_semantic_storage.py
```diff
@@ -0,0 +1,685 @@
...
+        *,
+        set_id: str,
+        category_name: str,
+        feature: str,
+        value: str,
+        tag: str,
+        embedding: InstanceOf[np.ndarray],
+        metadata: dict[str, Any] | None = None,
+    ) -> FeatureIdT:
+        feature_id = await self._next_feature_id()
```
malatewang: Will this be too expensive to call every time? Can we just use UUID here?

src/memmachine/semantic_memory/storage/neo4j_semantic_storage.py
```diff
@@ -0,0 +1,685 @@
...
+            target_dimensions = int(len(np.array(embedding, dtype=float)))
+        await self._ensure_set_embedding_dimensions(target_set_id, target_dimensions)
+
+        assignments: list[str] = ["f.updated_at_ts = $updated_at_ts"]
+        params: dict[str, Any] = {
+            "feature_id": int(feature_id),
+            "updated_at_ts": _utc_timestamp(),
+            "embedding_dimensions": target_dimensions,
+        }
+        if set_id is not None:
```
malatewang: When updating feature, I think the set_id is mandatory.

src/memmachine/semantic_memory/storage/storage_base.py
```diff
@@ -40,7 +45,7 @@ async def delete_all(self):
     @abstractmethod
     async def get_feature(
         self,
-        feature_id: int,
```
malatewang: The scope of get_feature should be constrained to one specific set_id?

src/memmachine/semantic_memory/storage/neo4j_semantic_storage.py
o-love: Will do

src/memmachine/semantic_memory/storage/neo4j_semantic_storage.py
o-love: The idea is the set_id is only needed if the feature is being moved from one set to another. 

If not we don't really need the set_id, just the feature_id

jealous (approved):

PR #542, opened 2025-11-19
Merge v0.2 dev branch

o-love:


src/memmachine/server/api_v2/filter_parser.py
```diff
@@ -0,0 +1,30 @@
...
+    if not filter_str:
+        return filter_dict
+
+    try:
+        conditions = filter_str.split("and")
+        for condition in conditions:
+            if "=" in condition:
+                key, value = condition.split("=", 1)
+                # Store as string for now - could be enhanced to parse types
+                filter_dict[key.strip()] = value.strip()
```
edwinyyyu: Parsing should use a proper tokenizer/lexer/parser. This is unsafe.

src/memmachine/server/api_v2/router.py
```diff
@@ -0,0 +1,453 @@
...
+            "org_id": org_id,
+            "project_id": project_id,
+        }
+        for org_id, project_id in (
+            session.split("/", 1) for session in sessions if "/" in session
+        )
+    ]
+
+
+@router.post("/projects/delete")
```
jgong: maybe we should use proper HTTP methods (DELETE) for delete restapi?

PR #565, opened 2025-11-20
Implement filters for LongTermMemory

edwinyyyu:
### Purpose of the change

Implement new filtering behavior for LongTermMemory.

Semantics:
None/NULL means missing/not provided/invalid rather than a valid value.

During ingestion, None/NULL is not allowed as a value, so the item can be missing the key in the database if schemaless.

For search, the retrieved item will not have any keys with None value.FilterablePropertyValue represents a proper value, so it cannot be None. But the property filters themselves can have None to indicate checking for missing value, using type hint FilterablePropertyValue | None.

### DIFF
--- a/tests/memmachine/episodic_memory/long_term_memory/test_long_term_memory.py
+++ b/tests/memmachine/episodic_memory/long_term_memory/test_long_term_memory.py
@@ -1,0 +1,391 @@
+from datetime import UTC, datetime, timedelta
+
+import pytest
+import pytest_asyncio
+from neo4j import AsyncGraphDatabase
+from sentence_transformers import CrossEncoder, SentenceTransformer
+from testcontainers.neo4j import Neo4jContainer
+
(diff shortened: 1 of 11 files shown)


tests/memmachine/episodic_memory/long_term_memory/test_long_term_memory.py
```diff
@@ -0,0 +1,188 @@
...
+    ]
+
+    await long_term_memory.add_episodes(episodes)
+
+    all_episodes = await long_term_memory.get_matching_episodes()
+    assert len(all_episodes) == len(episodes)
+    assert set(all_episodes) == set(episodes)
+
+
+@pytest.mark.asyncio
```
jealous, line 168: What are those empty tests for?

edwinyyyu: WIP. Will implement.

PR #608, opened 2025-11-25
Implement tree filter for long-term memory

edwinyyyu:


jealous:
Could you update the document to add some examples of the filter?  For example, what does `m.xxx` mean?  What does `?` mean in the end of a key?

tests/memmachine/episodic_memory/declarative_memory/test_declarative_memory.py
```diff
@@ -297,7 +303,11 @@ async def test_search(declarative_memory):
     results = await declarative_memory.search(
         query="Who wrote the test?",
         max_num_episodes=4,
-        property_filter={"project": "memmachine"},
+        property_filter=FilterComparison(
```
o-love, line 306: NIT: I think its nice for the tests creating a string and sending it through parse_filter
Both to showcase and test the filter string.

o-love: so `parse_filter("project = memmachine")`

o-love (approved):

tests/memmachine/episodic_memory/declarative_memory/test_declarative_memory.py
edwinyyyu: It's possible for a consumer to construct a FilterExpr manually, in which case there can be special characters. The current test makes sure that everything is mangled and sanitized properly.

edwinyyyu:
> Could you update the document to add some examples of the filter? For example, what does `m.xxx` mean? What does `?` mean in the end of a key?

`?` is part of the string. It doesn't mean anything.

PR #674, opened 2025-12-03
Update slow integration test for MemMachine

o-love:
Session creation is now done in the pytest fixture.

Update the slow tests to reflect changes.


edwinyyyu:
> Add a little to the description. This fixes the failing tests in main.

Looks like there are still failing tests in main...

PR #697, opened 2025-12-05
CI Fix: Disable bedrock llm test in PR and bedrock llm fixes

o-love:
Temporarily disable Bedrock LLM PR checks for now until results are more stable.

### DIFF
--- a/src/memmachine/common/language_model/amazon_bedrock_language_model.py
+++ b/src/memmachine/common/language_model/amazon_bedrock_language_model.py
@@ -283,6 +283,8 @@ class AmazonBedrockLanguageModel(LanguageModel):
                 output_format=output_format,
             )
             if parsed is not None:
+                self._collect_metrics({}, start_time, time.monotonic())
+
                 return parsed
             raise
 
--- a/tests/memmachine/conftest.py
+++ b/tests/memmachine/conftest.py
@@ -175,6 +175,7 @@ def bedrock_integration_language_model_config(bedrock_integration_config):
 @pytest.fixture(scope="session")
 def boto3_bedrock_runtime_client(bedrock_integration_config):
     import boto3
+
     config = bedrock_integration_config
 
     return boto3.client(
@@ -189,6 +190,7 @@ def boto3_bedrock_runtime_client(bedrock_integration_config):
 @pytest.fixture(scope="session")
 def boto3_bedrock_agent_runtime_client(bedrock_integration_config):
     import boto3
+
     config = bedrock_integration_config
 
     return boto3.client(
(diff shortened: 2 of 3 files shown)


o-love:
slow tests for bedrock pass locally but I think it might be best to have them run only on main branch until we see that they are more stable

tests/memmachine/conftest.py
```diff
@@ -164,6 +164,7 @@ def bedrock_integration_config():
         "aws_secret_access_key": aws_secret_access_key,
         "aws_session_token": aws_session_token,
         "aws_region": aws_region,
+        "model": "qwen.qwen3-32b-v1:0",
```
edwinyyyu: This config is used by the reranker as well. Can model id be added elsewhere?

tests/memmachine/main/conftest.py
```diff
@@ -110,7 +109,6 @@ def bedrock_language_model_config(
     bedrock_integration_config,
```
edwinyyyu: Specify the LLM model_id in this config instead of bedrock_integration_config.

src/memmachine/common/language_model/amazon_bedrock_language_model.py
```diff
@@ -273,13 +275,26 @@ async def generate_parsed_response(
...
+        try:
+            response = await client.chat.completions.create(**converse_kwargs)
+        except instructor.core.exceptions.InstructorRetryException as exc:  # type: ignore[attr-defined]
+            parsed = self._try_parse_bedrock_completion(
+                completion=getattr(exc, "last_completion", None),
+                output_format=output_format,
+            )
+            if parsed is not None:
+                return parsed
+            raise
```
edwinyyyu, line 289: Fails to collect metrics in some paths.

edwinyyyu (changes requested):

src/memmachine/common/language_model/amazon_bedrock_language_model.py
o-love: I'll see what can be done. This is catching an exception that wasn't handled before. So it maintains the same behavior of not collecting metrics on failures.

src/memmachine/common/language_model/amazon_bedrock_language_model.py
```diff
@@ -406,6 +421,78 @@ async def generate_response(  # noqa: C901
...
+        if not text_blocks:
+            return None
+
+        raw_text = "\n".join(text_blocks)
+        cleaned_text = re.sub(
+            pattern=r"<think>.*?</think>\s*",
+            repl="",
+            string=raw_text,
+            flags=re.DOTALL,
+        ).strip()
```
edwinyyyu, line 486: Is this generic?

o-love: Maybe or maybe not. Right now bedrock is the only one that needs this. It can be refactored to a util function if it needs to be reused.

src/memmachine/common/language_model/amazon_bedrock_language_model.py
edwinyyyu: Can we get latency at least, in the case `return parsed`?

o-love: Addressed

edwinyyyu (approved):

PR #771, opened 2025-12-11
Fix Neo4jVectorGraphStore timestamp comparisons

edwinyyyu:
### Purpose of the change

Neo4j does not treat timestamps as equal if they have different timezones even if they represent the same point in time. This makes searching by timestamp result in unexpected behavior, and caused duplicate episodes to appear in contextualization.

### Description

Use seconds since epoch and nanoseconds to compare timestamps in Neo4j.
Neo4j DateTime can cause SEGFAULTS when using unexpected tzinfo types, so we must sanitize the Python timestamp value.

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

- [ ] Unit Test
- [x] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### DIFF
--- a/src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
+++ b/src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
@@ -6,6 +6,7 @@ of a vector graph store using Neo4j as the backend database.
 """
 
 import asyncio
+import datetime
 import logging
 import re
 import time
@@ -859,7 +860,12 @@ class Neo4jVectorGraphStore(VectorGraphStore):
                 (
                     " OR ("
                     + " AND ".join(
-                        f"(n.{sanitized_by_property} = $starting_at[{index}])"
+                        Neo4jVectorGraphStore._render_comparison(
+                            f"n.{sanitized_by_property}",
+                            "=",
+                            f"$starting_at[{index}]",
+                            starting_at[index],
+                        )
                         for index, sanitized_by_property in enumerate(
                             sanitized_by_properties,
                         )
@@ -897,7 +903,10 @@ class Neo4jVectorGraphStore(VectorGraphStore):
             "RETURN n\n"
             f"{query_order_by}"
             f"{'LIMIT $limit' if limit is not None else ''}",
-            starting_at=starting_at,
+            starting_at=[
+                Neo4jVectorGraphStore._sanitize_python_value(starting_at_value)
+                for starting_at_value in starting_at
+            ],
             limit=limit,
             query_filter_params=query_filter_params,
         )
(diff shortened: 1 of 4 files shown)


Tianyang-Zhang (commented):
Can we simply convert the datetime object to naive or same timezone before store? User/caller can also convert it back if they need. If we want it to be more user friendly, we can store the timezone info and convert back before returning the episode. Would that make the entire logic much simpler?

edwinyyyu:
> Can we simply convert the datetime object to naive or same timezone before store? User/caller can also convert it back if they need. If we want it to be more user friendly, we can store the timezone info and convert back before returning the episode. Would that make the entire logic much simpler?

It can also make sense to convert everything to UTC before store. It would make the database logic simpler but our application logic more complex. Since the search is done on the server side, we would either have to store the timezone information or ask the user during search.

Tianyang-Zhang:
> > Can we simply convert the datetime object to naive or same timezone before store? User/caller can also convert it back if they need. If we want it to be more user friendly, we can store the timezone info and convert back before returning the episode. Would that make the entire logic much simpler?
> 
> It would make the database logic simpler but our application logic more complex.

I think we can still make the application logic simple by doing:
1. when ingest, convert any filterable datetime object to UTC, no need to store the original timezone.
2. when search, temporary cache the given datetime object's timezone, then convert to UTC and search. Before returning results, convert all result episode's datetime objects to the cached timezone. (when search, user's given datetime must have timezone info, otherwise we should return error. Because the server and client may not on the same host or timezone)

So that user doesn't need to provice any specific timezone info, and we don't need to store timezone info. The only downside is if user login to backend database manually, they will see all datetime objects in UTC instead of original timezone.

What do you think?

edwinyyyu:
> > > Can we simply convert the datetime object to naive or same timezone before store? User/caller can also convert it back if they need. If we want it to be more user friendly, we can store the timezone info and convert back before returning the episode. Would that make the entire logic much simpler?
> > 
> > 
> > It would make the database logic simpler but our application logic more complex.
> 
> I think we can still make the application logic simple by doing:
> 
>     1. when ingest, convert any filterable datetime object to UTC, no need to store the original timezone.
> 
>     2. when search, temporary cache the given datetime object's timezone, then convert to UTC and search. Before returning results, convert all result episode's datetime objects to the cached timezone. (when search, user's given datetime must have timezone info, otherwise we should return error. Because the server and client may not on the same host or timezone)
> 
> 
> So that user doesn't need to provice any specific timezone info, and we don't need to store timezone info. The only downside is if user login to backend database manually, they will see all datetime objects in UTC instead of original timezone.
> 
> What do you think?

This could work and but I think it may require an API change.
One consideration is whether we want to support keeping local timestamps for someone who travels. It's an edge case.

Tianyang-Zhang:
> This could work and but I think it may require an API change.

Isn't current memory layer API only taking datetime object?

> One consideration is whether we want to support keeping local timestamps for someone who travels

Not sure if I understand correctly. Do you mean that user may want to retrieve the timestamp in the timezone same as the ingestion timezone? I feel the timezone of retrieved timestamps should always follow the given one(s) when searching. So that everything is aligned.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -1656,3 +1715,57 @@ def _render_filter_expr(
...
         raise TypeError(f"Unsupported filter expression type: {type(expr)!r}")
+
+    @staticmethod
+    def _render_comparison(
+        left: str,
+        op: str,
+        right: str,
+        value: FilterablePropertyValue | list[FilterablePropertyValue],
+    ) -> str:
+        if isinstance(value, list):
```
malatewang, line 1728: If can not accept list value, why allow list in the parameters?

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -1656,3 +1715,57 @@ def _render_filter_expr(
...
+
+            if op in ("!=", "<>"):
+                return (
+                    "("
+                    f"{left} <> {right}"
+                    " AND "
+                    "("
+                    f"{left}.epochSeconds <> {right}.epochSeconds"
+                    " OR "
+                    f"{left}.nanosecond <> {right}.nanosecond"
```
malatewang, line 1752: Is it necessary to check nanosecond?

src/memmachine/episodic_memory/declarative_memory/declarative_memory.py
```diff
@@ -98,6 +98,8 @@ def __init__(self, params: DeclarativeMemoryParams) -> None:
 
         self._derived_from_relation = f"DERIVED_FROM_{session_id}"
 
+        self._score_single_episodes_threshold = 10
```
malatewang, line 101: What is the meaning of this number? Please add a comment so other people can understand the logic.

edwinyyyu: Sorry, may need to rebase this. Without this, a long term memory test will fail because fixing the timestamp comparisons results in some different episodes being returned.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
edwinyyyu: We need to compare nanoseconds to ensure that two timestamps are different points in time if we keep time zones.

edwinyyyu:
> > This could work and but I think it may require an API change.
> 
> Isn't current memory layer API only taking datetime object?
> 
> > One consideration is whether we want to support keeping local timestamps for someone who travels
> 
> Not sure if I understand correctly. Do you mean that user may want to retrieve the timestamp in the time zone same as the ingestion  ? I feel the time zone of retrieved timestamps should always follow the given one(s) when searching. So that everything is aligned.

1. I don't think we know the time zone during search with what we have right now.
2. Yes, we may have edge cases where a user talks about a time of day like morning or night.

edwinyyyu:
Just thought of something: Before merging, it may be good to check the performance of this and whether indexing works.

Or we just go UTC.

Tianyang-Zhang:
> 1. I don't think we know the time zone during search with what we have right now.

I thought the user/caller provided datetime object must include the timezone info? If not, we can't really compare/filtering the timestamp user gives.

edwinyyyu:
Outdated.

PR #777, opened 2025-12-11
Potential fix for code scanning alert no. 37: Workflow does not contain permissions

sscargal:
Potential fix for [https://github.com/MemMachine/MemMachine/security/code-scanning/37](https://github.com/MemMachine/MemMachine/security/code-scanning/37)

To fix the problem, add an explicit `permissions` block to the workflow to restrict the `GITHUB_TOKEN` permissions to the minimum required. In this case, since the workflow only reads files from the repository and does not need to write or manage anything, setting `contents: read` is sufficient. The best place to add the block is at the workflow root (before the `jobs:` key in `.github/workflows/check-uvlockfile.yml`) so it applies to all jobs, or optionally inside the `check-lock` job. Our goal is not to change any functionality, but just restrict permissions.

### DIFF
--- a/.github/ISSUE_TEMPLATE/question.yml
+++ b/.github/ISSUE_TEMPLATE/question.yml
@@ -20,7 +20,7 @@ body:
     attributes:
       label: Additional details
       description: Environment, steps, code snippets, logs, or screenshots (optional).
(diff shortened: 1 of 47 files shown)


jgong:
Hi @sscargal , thanks for taking care of these, may i ask where can i check the code scanning alerts? it seems i don't have permission for security advisories.

Issue #706, opened 2025-12-05
[Bug]: cannot find where debug logs are written


tomw-mv:
### Describe the bug

in amazon_bedrock_reranker.py, AmazonBedrockReranker.score(), it write some logs:
e.g.
                logger.debug(
                    "[call uuid: %s] "
                    "Scoring %d candidates for query using %s Amazon Bedrock model",
                    score_call_uuid,
                    len(candidates),
                    self._model_id,
                )

I set debug logs in configuration.yml

logging:
  level: debug
  path: mem-machine.log

Restart memmachine
e.g.
2025-12-05 23:21:57,591 [10.2.100.33] [DEBUG pexpect_util.py::pexpect_interact::329::MainThread] pinter: send cmd: uv run bash memmachine-compose.sh stop
2025-12-05 23:22:31,606 [10.2.100.33] [DEBUG pexpect_util.py::pexpect_interact::329::MainThread] pinter: send cmd: uv run bash memmachine-compose.sh
2025-12-05 23:23:12,385 [10.2.100.33] [DEBUG pexpect_util.py::pexpect_read::262::MainThread] pread: got a line(0):qa_idx=None: b'\x1b[0;32m[SUCCESS]\x1b[0m \xf0\x9f\x8e\x89 MemMachine is now running!'

Run test, search memory is happening
e.g.
2025-12-05 23:23:24,775 [10.2.100.33] [DEBUG pexpect_util.py::pexpect_interact::329::MainThread] pinter: send cmd: uv run python restapiv2_locomo_search.py --data-path ../locomo10.json --target-path /home/memmachine/MemMachine/evaluation/locomo/episodic_memory/mmai_gpto
2025-12-05 23:23:25,602 [10.2.100.33] [DEBUG pexpect_util.py::pexpect_read::264::MainThread] pread: got a line(0): b'mmai health:'
2025-12-05 23:23:25,602 [10.2.100.33] [DEBUG pexpect_util.py::pexpect_read::264::MainThread] pread: got a line(0): b'k=status v=healthy'
2025-12-05 23:23:25,603 [10.2.100.33] [DEBUG pexpect_util.py::pexpect_read::264::MainThread] pread: got a line(0): b'k=service v=memmachine'
2025-12-05 23:23:25,607 [10.2.100.33] [DEBUG pexpect_util.py::pexpect_read::264::MainThread] pread: got a line(0): b'top_k=50'
[...]

PR #779, opened 2025-12-12
[GH-706] Validate log level is applied

o-love:
Add tests to validate correct behavior of apply log level config.


sscargal (approved):
LGTM

PR #811, opened 2025-12-16
fix: update MemMachine API URL to /docs endpoint

dxu104:
### Purpose of the change

Fix incorrect MemMachine API URL in startup script output.

### Description

The script was displaying `http://localhost:8080` as the MemMachine API URL, but the root path `/` returns 404 as no route is defined. Updated the URL to point to `/docs` which serves the Swagger UI documentation page.

**Changes:**
- Updated MemMachine API URL from root path to `/docs` (Swagger UI documentation)

### Type of change

### How Has This Been Tested?

Tested the endpoint after starting services:
- ✅ `http://localhost:8080/docs` - Returns Swagger UI documentation
- ✅ Verified root path `http://localhost:8080` correctly returns 404

### DIFF
--- a/memmachine-compose.sh
+++ b/memmachine-compose.sh
@@ -751,7 +751,7 @@ show_service_info() {
     print_success "🎉 MemMachine is now running!"
     echo ""
     echo "Service URLs:"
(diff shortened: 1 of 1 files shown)


sscargal (approved):
LGTM, thanks.

PR #827, opened 2025-12-18
Add openai-compatible model provider in memmachine-compose.sh

mwqgithub:
### Purpose of the change

Allow user to set an openai compatible model provider like qwen and deepseek.

### Description

Add a openai-compatible model provider in memmachine-compose.sh. This allows user to set an openai compatible model provider like qwen and deepseek.

### Type of change

### How Has This Been Tested?

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### DIFF
--- a/memmachine-compose.sh
+++ b/memmachine-compose.sh
@@ -148,6 +148,12 @@ select_llm_model() {
             llm_model=$(echo "${llm_model:-llama3}" | tr -d '\n\r')
             print_success "Selected Ollama LLM model: $llm_model" >&2
             ;;
+        "OPENAI_COMPATIBLE")
+            print_prompt
+            read -p "Which OpenAI-compatible LLM model would you like to use? [qwen-flash]: " llm_model
+            llm_model=$(echo "${llm_model:-qwen-flash}" | tr -d '\n\r')
+            print_success "Selected OpenAI-compatible LLM model: $llm_model" >&2
+            ;;
         *)
             print_warning "Unknown provider: $provider. Using default LLM model." >&2
             llm_model="gpt-4o-mini"
@@ -180,6 +186,12 @@ select_embedding_model() {
             read -p "Which Ollama embedding model would you like to use? [nomic-embed-text]: " embedding_model
             embedding_model=$(echo "${embedding_model:-nomic-embed-text}" | tr -d '\n\r')
             print_success "Selected Ollama embedding model: $embedding_model" >&2
+            ;;
+        "OPENAI_COMPATIBLE")
+            print_prompt
+            read -p "Which OpenAI-compatible embedding model would you like to use? [text-embedding-v4]: " embedding_model
+            embedding_model=$(echo "${embedding_model:-text-embedding-v4}" | tr -d '\n\r')
+            print_success "Selected OpenAI-compatible embedding model: $embedding_model" >&2
             ;;
         *)
             print_warning "Unknown provider: $provider. Using default embedding model." >&2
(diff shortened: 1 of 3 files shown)


memmachine-compose.sh
```diff
@@ -136,6 +136,12 @@ select_llm_model() {
             llm_model=$(echo "${llm_model:-gpt-4o-mini}" | tr -d '\n\r')
             print_success "Selected OpenAI LLM model: $llm_model" >&2
             ;;
+        "OTHERS")
+            print_prompt
+            read -p "Which OpenAI-compatible LLM model would you like to use? [gpt-4o-mini]: " llm_model
```
Kevin-K-W: I'd like to change the default value to "qwen-flush"

memmachine-compose.sh
```diff
@@ -169,6 +175,12 @@ select_embedding_model() {
             embedding_model=$(echo "${embedding_model:-text-embedding-3-small}" | tr -d '\n\r')
             print_success "Selected OpenAI embedding model: $embedding_model" >&2
             ;;
+        "OTHERS")
+            print_prompt
+            read -p "Which OpenAI-compatible embedding model would you like to use? [text-embedding-3-small]: " embedding_model
```
Kevin-K-W: change to text-embedding-v4

memmachine-compose.sh
```diff
@@ -476,15 +496,15 @@ check_config_file() {
             MEMMACHINE_IMAGE="memmachine/memmachine:latest-cpu"
         fi
 
-        # Ask user for provider path (OpenAI, Bedrock, or Ollama)
+        # Ask user for provider path (OpenAI, Others, Bedrock, or Ollama)
         print_prompt
-        read -p "Which provider would you like to use? (OpenAI/Bedrock/Ollama) [OpenAI]: " provider_input
+        read -p "Which provider would you like to use? (OpenAI/Others/Bedrock/Ollama) [OpenAI]: " provider_input
```
Kevin-K-W: OpenAI/Bedrock/Ollama/Others

memmachine-compose.sh
```diff
@@ -905,6 +967,10 @@ case "${1:-}" in
         echo "             Default LLM: gpt-4o-mini"
         echo "             Default embedding: text-embedding-3-small"
         echo "             Requires: OpenAI API key"
+        echo "  Others   - Uses an OpenAI-compatible endpoint (custom base URL)"
```
Kevin-K-W: Put this to the end of list

sscargal (commented):
Thank you for opening this PR and for working to improve model provider options! I wanted to share some thoughts and get your input on the naming and clarity:

- The addition of an "Others" provider might be confusing or redundant, as the existing "openai" and "ollama" options already support any OpenAI-compatible endpoint, not just the official OpenAI API.
- Users may not always realize that "openai" and "ollama" cover both the OpenAI service and any compatible endpoints from other providers. The current naming could imply it's limited to OpenAI only.
- To help with clarity and reduce potential confusion, it could be helpful to update the documentation or add tooltips specifying that the "openai" option supports all OpenAI-compatible APIs—including third-party endpoints.
- Alternatively, renaming the proposed "Others" option to something like "OpenAI/Compatible" or "Custom (OpenAI-compatible)" might make the scope more obvious to users.

I am open to suggestions and feedback. I'm okay with the last bullet of renaming this proposal from "Others" to "Custom (OpenAI-compatible)". However, we will need to change the input method from asking the user to type the option to selecting the option from a numbered list. This approach would improve the user experience.

Furthermore, we would need to add this new option to the `memmachine-configure` command so that they remain in sync.

jealous (approved):

jealous:
Just notice Steve's comment.  I think it makes sense.  You may need to update the change.

mwqgithub:
Hi @sscargal , thanks for your comment.
The reason for adding the "others" option is the "openai" option uses the response API, which most openai-compatible providers don't support. If we can make the "openai" options use the chat completion API too, I think it's simpler and less confusing to let users use "openai" for both openai and third-party services.
I'm not sure about the benefits of the responses API over the chat completions API. If there's no significant difference between them, I think we can just change the API type for the "openai" option.
@jealous @Kevin-K-W

sscargal:
@mwqgithub Thanks for the update. I agree. Perhaps we should consider renaming the proposed "Others" option to something like "OpenAI/Compatible" or "Custom (OpenAI-compatible)", which might make the scope more obvious to users.  "Others" is too general IMHO and could mean anything.

mwqgithub:
Hi, @Kevin-K-W @sscargal . I've updated the code. Could you take a look again?

PR #832, opened 2025-12-18
Fix Type issues in main branch

o-love:
Regression was introduced into main due to a dependency missing, causing ty to not be able to validate types. 

Add a lint check to validate `uv.lock` and fix main issues.


.github/workflows/lint.yml
```diff
@@ -42,5 +42,8 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Ensure uv.lock is correct
```
sscargal, line 45: This is a good idea. 

FYI, we have `.github/workflows/check-uvlockfile.yml` that runs `uv lock --check`. It is/was designed to highlight out-of-sync `uv.lock` files at PR time and have the developer update their file. 

This change would ensure it's always synchronized, making the `check-uvfile` redundant, which is a good thing. However, it **must be the first workflow to run** to avoid issues with other workflows that require the uvlock to be in sync. Alternatively, we create a dedicated uv workflow to sync the file and ensure it's the first to run.

I'm open to suggestions.

sscargal (commented):
The remaining changes look good to me.

.github/workflows/lint.yml
```diff
@@ -42,5 +42,8 @@ jobs:
...
+        run: uv lock --refresh && git diff --exit-code
```
edwinyyyu, line 46: How is this supposed to work?
Why is it the --refresh option?

o-love: `--refresh` refreshes the cache before running `uv lock`. 
With the git diff returning an error code if there are any changes.

.github/workflows/lint.yml
o-love: We can remove check-uvlockfile or maybe have it run always. I'll check if it would have caught this.

With the way are workflows are currently set up (all independent) there isn't a easy way to create that type of dag flow.

I think in general it is ok to have a single lint check validate the uv, assuming people only merge when all tests are successful.

PR #882, opened 2025-12-31
Semantic Memory SetID refactor: customization

o-love:
Adds support for dynamic set id configuration based on metadata and customization of set ids.
Categories are now also configurable.

Copilot main points:
- Replaces the `IsolationType` enum (USER/ROLE/SESSION) with a new org/project-based `DefaultType` system (OrgSet/ProjectSet/OtherSet)
- Introduces a `SemanticConfigStorage` layer backed by SQLAlchemy for persisting set-level configuration overrides
- Updates `SemanticSessionManager` to derive set IDs from `org_id`, `project_id`, and metadata fields rather than fixed user/session/role properties

### DIFF
--- a/src/memmachine/semantic_memory/semantic_ingestion.py
+++ b/src/memmachine/semantic_memory/semantic_ingestion.py
@@ -56,10 +56,17 @@ class IngestionService:
         self._debug_fail_loudly = params.debug_fail_loudly
 
     async def process_set_ids(self, set_ids: list[SetIdT]) -> None:
+        async def _run(set_id: SetIdT) -> None:
+            try:
+                await self._process_single_set(set_id)
+            except Exception:
+                logger.exception("Failed to process set_id %s", set_id)
+                raise
+
         logger.info("Starting ingestion processing for set ids: %s", set_ids)
 
         results = await asyncio.gather(
-            *[self._process_single_set(set_id) for set_id in set_ids],
+            *[_run(set_id) for set_id in set_ids],
             return_exceptions=True,
         )
 
@@ -99,14 +106,11 @@ class IngestionService:
         raw_messages = [tasks[h_id].result() for h_id in history_ids]
         none_h_ids = [h_id for h_id, task in tasks.items() if task.result() is None]
 
-        if none_h_ids:
-            logger.warning(
-                "Failed to retrieve messages. Invalid episode_ids: %s", none_h_ids
+        if len(none_h_ids) != 0:
+            raise ValueError(
+                "Failed to retrieve messages. Invalid episode_ids exist for set_id "
+                f"{set_id}: {none_h_ids}"
             )
-            if self._debug_fail_loudly:
-                raise ValueError(
-                    f"Failed to retrieve messages. Invalid episode_ids: {none_h_ids}"
-                )
 
         raw_messages = [m for m in raw_messages if m is not None]
         messages = TypeAdapter(list[Episode]).validate_python(raw_messages)
@@ -123,8 +127,7 @@ class IngestionService:
                     )
 
                     raise ValueError(
-                        "Message ID is None for message %s",
-                        message.model_dump(),
+                        f"Message ID is None for message {message.model_dump()}"
                     )
 
                 filter_expr = And(
--- a/tools/chatgpt2memmachine/migration.py
+++ b/tools/chatgpt2memmachine/migration.py
@@ -52,6 +52,7 @@ class MigrationHack:
         self.dry_run = dry_run
         self.max_workers = max_workers
         self.conversations = {}
+        self.total_messages = 0
 
     def load_and_extract(self):
         if self.input_file is None:
@@ -60,7 +61,25 @@ class MigrationHack:
         if not self.dry_run:
             print(f"Found {self.num_conversations} conversation(s)")
 
-        for conv_id in range(1, self.num_conversations + 1):
+        # Check if user specified a specific conversation index
+        user_index = self.filters.get("index", 0) if self.filters else 0
+        
+        # Determine which conversations to process
+        if user_index and user_index > 0:
+            # User specified a specific conversation, only process that one
+            if self.verbose:
+                print(f"DEBUG: User specified conversation index: {user_index}")
+            if user_index > self.num_conversations:
+                raise Exception(
+                    f"ERROR: Conversation index {user_index} is out of range. "
+                    f"File contains {self.num_conversations} conversation(s)."
+                )
+            conv_ids_to_process = [user_index]
+        else:
+            # Process all conversations
+            conv_ids_to_process = list(range(1, self.num_conversations + 1))
+
+        for conv_id in conv_ids_to_process:
             extracted_file_name = (
                 f"{self.input_file_base_name}_{conv_id}_extracted.json"
             )
@@ -69,20 +88,37 @@ class MigrationHack:
                 # load from file directly
                 with open(extracted_file, "r") as f:
                     self.conversations[conv_id] = json.load(f)
-                continue
-            if not self.dry_run:
-                print(f"Loading conversation {conv_id}...")
-            # Build filters for this conversation (merge global filters with conv index)
-            filters = dict(self.filters or {})
-            filters["index"] = conv_id
-            messages = self.parser.load(self.input_file, filters=filters)
-            if not self.dry_run:
-                print(f"Loaded {len(messages)} message(s) from conversation {conv_id}")
-            self.conversations[conv_id] = messages
-            # Dump extracted messages for this conversation
-            self.parser.dump_data(
-                messages, output_format="json", outfile=extracted_file
-            )
+            else:
+                if not self.dry_run:
+                    print(f"Loading conversation {conv_id}...")
+                # Build filters for this conversation (merge global filters with conv index)
+                filters = dict(self.filters or {})
+                filters["index"] = conv_id
+                messages = self.parser.load(self.input_file, filters=filters)
+                if not self.dry_run:
+                    print(f"Loaded {len(messages)} message(s) from conversation {conv_id}")
+                self.conversations[conv_id] = messages
+                # Dump extracted messages for this conversation
+                self.parser.dump_data(
+                    messages, output_format="json", outfile=extracted_file
+                )
+            
+            # Count messages for this conversation (applying user_only filter if enabled)
+            messages = self.conversations[conv_id]
+            if isinstance(messages, list):
+                if self.user_only:
+                    # Apply same filtering as _process_conversation
+                    for msg in messages:
+                        if isinstance(msg, str):
+                            self.total_messages += 1
+                        elif isinstance(msg, dict):
+                            role = msg.get("role", "")
+                            if isinstance(role, str) and role.lower() != "assistant":
+                                self.total_messages += 1
+                else:
+                    self.total_messages += len(messages)
+            elif isinstance(messages, str):
+                self.total_messages += 1
 
     def _format_message(self, message):
         """Format message for MemMachine API"""
@@ -263,6 +299,19 @@ class MigrationHack:
         # Process conversations using ThreadPoolExecutor
         # Default max_workers=1 for sequential processing
         workers = min(self.max_workers, len(contents))
+        
+        # Warn user if using sequential processing with many messages
+        if workers <= 1 and self.total_messages > 1000:
+            print(
+                f"WARNING: Processing {self.total_messages} messages with {workers} worker(s) (sequential). "
+                f"Consider setting --workers to larger value to speed up migration."
+            )
+            response = input("Do you want to continue? (yes/no): ").strip().lower()
+            if response not in ["yes", "y"]:
+                print("Exiting. Please adjust --workers and try again.")
+                return
+            
+        
         with ThreadPoolExecutor(max_workers=workers) as executor:
             futures = {
                 executor.submit(self._process_conversation, conv_id, messages): conv_id
--- a/tools/chatgpt2memmachine/parsers/openai_parser.py
+++ b/tools/chatgpt2memmachine/parsers/openai_parser.py
@@ -306,6 +306,7 @@ class OpenAIParser(BaseParser):
         data = self.load_json(infile)
         messages = []
         chat_count = 0
+        processed_chat_count = 0
         msg_count = 0
 
         for chat in data:
@@ -317,6 +318,7 @@ class OpenAIParser(BaseParser):
             ):
                 continue
 
+            processed_chat_count += 1
             chat_title_actual = chat.get("title", "")
             self.logger.info(f"Processing chat {chat_count}: {chat_title_actual}")
 
@@ -337,7 +339,7 @@ class OpenAIParser(BaseParser):
             )
 
         self.logger.info(
-            f"Total messages loaded: {len(messages)} from {chat_count} chats"
+            f"Total messages loaded: {len(messages)} from {processed_chat_count} chat(s)"
         )
         return messages
 


src/memmachine/semantic_memory/config_store/config_store.py
```diff
@@ -0,0 +1,109 @@
...
+@runtime_checkable
+class SemanticConfigStorage(Protocol):
+    """Contract for persisting and retrieving semantic memory configuration."""
+
+    async def set_setid_config(
+        self,
+        *,
+        set_id: SetIdT,
+        embedder_name: str | None = None,
+        llm_name: str | None = None,
```
malatewang, line 28: Each semantic memory also can set specific prompt

sample_configs/episodic_memory_config.cpu.sample
```diff
@@ -18,12 +18,13 @@ semantic_memory:
...
   embedding_model: openai_embedder
   database: profile_storage
+  config_database: profile_storage
 
 session_manager:
   database: profile_storage
 
 prompt:
-  session:
+  project_level:
```
malatewang: Could we have a better name?

src/memmachine/main/memmachine.py
```diff
@@ -481,3 +481,173 @@ async def delete_features(
...
+        )
+
+    async def create_semantic_set_type(
+        self,
+        *,
+        session_data: SessionData,
+        is_org_level: bool,
+        metadata_tags: list[str],
+    ) -> str:
+        semantic_session = await self._resources.get_semantic_session_manager()
```
malatewang, line 866: When creating a new set, embedder, reranker and prompt can be specified.

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,483 @@
...
+            raise NotImplementedError
+
+        stmt = ins.values(
+            set_id=set_id,
+            embedder_name=embedder_name,
+            language_model_name=llm_name,
+        ).on_conflict_do_update(
+            index_elements=["set_id"],
+            set_={
+                "embedder_name": embedder_name,
```
malatewang, line 280: Do we want to allow embedder update? What will happen if the value is None?

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,483 @@
...
+        )
+
+    async def create_category(
+        self,
+        *,
+        set_id: SetIdT,
+        category_name: str,
+        description: str,
+    ) -> CategoryIdT:
+        stmt = (
```
malatewang, line 501: Check if set_id exists first?

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,483 @@
...
+    async def add_tag(
+        self,
+        *,
+        category_id: CategoryIdT,
+        tag_name: str,
+        description: str,
+    ) -> TagIdT:
+        category_id_int = int(category_id)
+
+        tag_stmt = (
```
malatewang, line 659: It is better to check duplicate here.

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,483 @@
...
+        async with self._create_session() as session:
+            await session.execute(delete(Tag).where(Tag.id == tag_id_int))
+            await session.commit()
+
+    async def add_org_set_id(
+        self,
+        *,
+        org_id: str,
+        org_level_set: bool = False,
+        metadata_tags: list[str],
```
malatewang, line 713: Why need this?

src/memmachine/main/memmachine.py
```diff
@@ -481,3 +481,199 @@ async def delete_features(
     ) -> None:
         semantic_session = await self._resources.get_semantic_session_manager()
         await semantic_session.delete_features(feature_ids)
+
+    async def add_feature(
+        self,
```
malatewang, line 744: Please add docstring for the functions

src/memmachine/semantic_memory/semantic_session_manager.py
```diff
@@ -275,48 +290,304 @@ async def delete_feature_set(
...
+        is_org_level: bool
+        tags: Mapping[str, str]
+
+    async def _get_set_id_entries(
+        self,
+        *,
+        session_data: SessionData,
+        metadata: Mapping[str, JsonValue] | None = None,
+    ) -> list[SetIdEntry]:
+        org_set_ids = await self._semantic_config.list_org_set_ids(
```
malatewang: It is too expensive to query database for every episode. Should be cached.

src/memmachine/semantic_memory/semantic_session_manager.py
```diff
@@ -275,48 +290,304 @@ async def delete_feature_set(
...
+                return def_type
+
+        raise RuntimeError(f"Invalid set_id: {set_id}")
+
+    async def create_org_set_type(
+        self,
+        *,
+        session_data: SessionData,
+        is_org_level: bool = False,
+        metadata_tags: list[str],
```
malatewang: I think we make this too complicated. It may be better to just allow creating semantic memory for one metadata key. Otherwise, it will be very difficult to handle the value combinations.

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,483 @@
...
+            categories=categories,
+            disabled_categories=disabled_categories,
+        )
+
+    async def create_category(
+        self,
+        *,
+        set_id: SetIdT,
+        category_name: str,
+        description: str,
```
malatewang: The description should is used as prompt? Rename it to prompt

src/memmachine/main/memmachine.py
```diff
@@ -481,3 +481,199 @@ async def delete_features(
...
+            description=description,
+        )
+
+    async def semantic_disable_default_categories(
+        self,
+        *,
+        session_data: SessionData,
+        set_metadata_keys: list[str],
+        is_org_level: bool = False,
+        category_name: str,
```
malatewang, line 1089: If disable default category, why require category name?

src/memmachine/main/memmachine.py
```diff
@@ -481,3 +481,199 @@ async def delete_features(
...
+            is_org_level=is_org_level,
+            category_name=category_name,
+        )
+
+    async def semantic_delete_category(
+        self,
+        *,
+        category_id: CategoryIdT,
+    ) -> None:
+        semantic_session = await self._resources.get_semantic_session_manager()
```
malatewang, line 1145: When deleting a category, probably we should also delete it data.

src/memmachine/main/memmachine.py
```diff
@@ -481,3 +481,199 @@ async def delete_features(
...
+        return await semantic_session.update_feature(
+            feature_id=feature_id,
+            category_name=category_name,
+            feature=feature,
+            value=value,
+            tag=tag,
+            metadata=metadata,
+        )
+
+    async def create_semantic_set_type(
```
malatewang, line 842: why call it "set_type" ? I think this function create a semantic set?

src/memmachine/main/memmachine.py
```diff
@@ -481,3 +481,199 @@ async def delete_features(
...
+
+        return await semantic_session.list_org_set_types(
+            session_data=session_data,
+        )
+
+    async def configure_semantic_set_type(
+        self,
+        *,
+        session_data: SessionData,
+        is_org_level: bool = False,
```
malatewang: Why not just use the set ID?

src/memmachine/main/memmachine.py
```diff
@@ -481,3 +481,199 @@ async def delete_features(
...
+            tag=tag,
+            metadata=metadata,
+        )
+
+    async def create_semantic_set_type(
+        self,
+        *,
+        session_data: SessionData,
+        is_org_level: bool,
+        metadata_tags: list[str],
```
malatewang, line 847: I think it is better to allow use to set a unique name and description

src/memmachine/main/memmachine.py
```diff
@@ -481,3 +481,199 @@ async def delete_features(
...
+        )
+
+    async def semantic_add_category(
+        self,
+        *,
+        session_data: SessionData,
+        set_metadata_keys: list[str],
+        is_org_level: bool = False,
+        category_name: str,
+        description: str,
```
malatewang: For one category, we probably allow both description and prompt

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,483 @@
...
+    language_model_name = mapped_column(String, nullable=True)
+
+    disabled_categories: Mapped[list["DisabledDefaultCategories"]] = relationship(
+        "DisabledDefaultCategories",
+        cascade="all, delete-orphan",
+        single_parent=True,
+    )
+
+
+class DisabledDefaultCategories(BaseSemanticConfigStore):
```
malatewang, line 85: What is the default meaning here?

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,483 @@
...
+            embedder_name=embedder_name,
+            categories=categories,
+            disabled_categories=disabled_categories,
+        )
+
+    async def create_category(
+        self,
+        *,
+        set_id: SetIdT,
+        category_name: str,
```
malatewang, line 497: I think the code requires the category name to be unique in one set.

src/memmachine/main/memmachine.py
```diff
@@ -481,3 +481,199 @@ async def delete_features(
...
+            is_org_level=is_org_level,
+            set_metadata_keys=set_metadata_keys,
+            embedder_name=embedder_name,
+            llm_name=llm_name,
+        )
+
+    async def semantic_add_category(
+        self,
+        *,
+        session_data: SessionData,
```
malatewang: It is clearer to use set_id as the key

malatewang (changes requested):

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
o-love: good catch.

I think for now I want to disable embedder updates.
I was thinking of maybe clearing out the set id along with any vector index on an embedder change. 
The other obvious solution would be re-embedding

But any way of allowing embedder change would have serious side effects. Which should probably be made explicit.

Maybe in the future a clone/migrate memory endpoint that allows for embedder rebuilding and other high cost changes to the memory structure.

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
o-love: Prompts that are specified in the MemMachine config instead of in the database

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
o-love: Yes. The table has a unique constraint to enforce it.

```
UniqueConstraint("set_id", "name", name="_set_id_name_uc"),
```

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
o-love: We had a discussion in person after this message. 
But to record it here:

The idea is that an admin can dynamically create new 'isolations' based on metadata tags.
So if the admin wishes to, for example, create user profiles based on `user_id` at the org level.
They would call `add_org_set_id(<org_id>, False, ["user_id"])`.

Then when `add_episode` gets called with a `user_id` metadata, it will be ingested into the set_id corresponding to that user_id.

src/memmachine/semantic_memory/config_store/config_store.py
o-love: The prompts of the set_id is what I call `category` where each set_id can have multiple. That gets set in the following functions regarding categories and tags.

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
malatewang: But that is just configuration, which should be transparent to the code logic.

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,479 @@
...
+
+class Category(BaseSemanticConfigStore):
+    """Semantic category definition with its prompt description."""
+
+    __tablename__ = "semantic.config.category"
+
+    id = mapped_column(Integer, primary_key=True, nullable=False)
+    set_id = mapped_column(String, nullable=False, index=True)
+    name = mapped_column(String, nullable=False)
+    prompt_description = mapped_column(String, nullable=False)
```
malatewang: I would suggest two columns: prompt and description. prompt is used to process data and description is for human

src/memmachine/semantic_memory/semantic_session_manager.py
```diff
@@ -196,26 +198,35 @@ async def add_feature(
...
-            raise ValueError("Invalid set_ids", set_ids)
-        set_id = set_ids[0]
+        self._assert_session_data_implements_protocol(session_data=session_data)
+
+        metadata = (
+            {key: session_data.metadata.get(key) for key in set_metadata_keys}
+            if session_data.metadata is not None
+            else {}
+        )
+        set_id = self._generate_set_id(
```
malatewang: Would it is easier to requires set_id as the parameter?

src/memmachine/main/memmachine.py
o-love: This function is that admin level creation of a set of metadata tags to seperate and ingest messages by.

src/memmachine/main/memmachine.py
o-love: I guess my main thought was to not have the user be aware of the set id. 

I can create a function to get the set id that matches a set of metadata tags.
As well as one that returns a list of set_ids belonging to the user.

src/memmachine/main/memmachine.py
o-love: I can do it. Add in a comment clarifying that this also deletes ingested data. Not just the prompt.

src/memmachine/common/configuration/__init__.py
```diff
@@ -153,18 +153,18 @@ def episode_summary_user_prompt(self) -> str:
...
         self,
-    ) -> dict[IsolationType, list[SemanticCategory]]:
+    ) -> dict[SemanticSessionManager.DefaultType, list[SemanticCategory]]:
         """Build the default semantic categories for each isolation type."""
         semantic_categories = PREDEFINED_SEMANTIC_CATEGORIES
 
         return {
-            IsolationType.SESSION: [
-                semantic_categories[s_name] for s_name in self.session
+            SemanticSessionManager.DefaultType.OrgSet: [
```
malatewang: I think we should not define default type

src/memmachine/main/memmachine.py
```diff
@@ -483,5 +659,371 @@ async def delete_features(
...
+            None.
+
+        """
         semantic_session = await self._resources.get_semantic_session_manager()
         await semantic_session.delete_features(feature_ids)
+
+    async def add_feature(
+        self,
+        session_data: SessionData,
+        *,
```
malatewang, line 745: Would it be better to use set_id + category_name when adding a new feature?

src/memmachine/main/memmachine.py
```diff
@@ -483,5 +659,371 @@ async def delete_features(
...
+        return await semantic_session.get_feature(
+            feature_id=feature_id,
+            load_citations=load_citations,
+        )
+
+    async def update_feature(
+        self,
+        *,
+        feature_id: FeatureIdT,
+        category_name: str | None = None,
```
malatewang, line 808: Why need category_name because feature_id is supposed to uniquely identify the feature?

src/memmachine/main/memmachine.py
malatewang: I think set_id is friendly to program. And we may also allow the user to provide a unique set name, which can be used too and it is more friendly to human user.

src/memmachine/main/memmachine.py
```diff
@@ -483,5 +659,371 @@ async def delete_features(
...
+
+        return await semantic_session.add_new_category(
+            session_data=session_data,
+            set_metadata_keys=set_metadata_keys,
+            is_org_level=is_org_level,
+            category_name=category_name,
+            description=description,
+        )
+
+    async def semantic_disable_default_categories(
```
malatewang: And again, I think the default category concept should be only in configuration

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,479 @@
...
+
+class SetIdResources(BaseSemanticConfigStore):
+    """Resource-level configuration associated with a set identifier."""
+
+    __tablename__ = "semantic.config.setidresources"
+
+    set_id = mapped_column(String, primary_key=True, nullable=False)
+    embedder_name = mapped_column(String, nullable=True)
+    language_model_name = mapped_column(String, nullable=True)
+
```
malatewang, line 59: I would suggest two more columns: name and description

src/memmachine/semantic_memory/semantic_memory.py
```diff
@@ -245,25 +373,282 @@ async def update_feature(
...
+                results[set_id] = tasks[embedder_str].result()[0]
+
+        return results
+
+    async def _set_ids_embedders(
+        self,
+        *,
+        set_ids: list[SetIdT],
+    ) -> tuple[dict[str, str | None], dict[str, Embedder], bool]:
+        async def _get_embedder_name(s_id: SetIdT) -> str | None:
```
malatewang, line 466: It is too expensive to query database for each episode. The configuration should be cached somewhere.

src/memmachine/semantic_memory/semantic_memory.py
```diff
@@ -245,25 +373,282 @@ async def update_feature(
...
+            has_default_embedder,
+        ) = await self._set_ids_embedders(set_ids=set_ids)
+
+        tasks: dict[str, asyncio.Task[list[list[float]]]] = {}
+        default_task: asyncio.Task[list[list[float]]] | None = None
+        async with asyncio.TaskGroup() as tg:
+            for e_str, e_cls in embedders.items():
+                tasks[e_str] = tg.create_task(e_cls.search_embed(queries=[query]))
+
+            if has_default_embedder:
```
malatewang, line 443: I think the default embedder should be set when the semantic set is created. If the user does not specify the embedder when creating the set, use the default embedder at that time. This can prevent problems if the default embedder was changed later.

src/memmachine/semantic_memory/semantic_memory.py
```diff
@@ -245,25 +373,282 @@ async def update_feature(
...
+            embedder=embedder,
+            language_model=llm_model,
+            semantic_categories=categories,
+        )
+
+    async def delete_set_id(self, *, set_ids: list[SetIdT]) -> None:
+        logger.info("Deleting set ids %s", set_ids)
+
+        async with asyncio.TaskGroup() as tg:
+            tg.create_task(self._semantic_storage.delete_history_set(set_ids=set_ids))
```
malatewang, line 542: TODO: we may need transaction or garbage collection for this function.

src/memmachine/semantic_memory/semantic_session_manager.py
```diff
@@ -275,48 +290,304 @@ async def delete_feature_set(
...
+        await self._semantic_service.set_set_id_config(
+            set_id=set_id,
+            embedder_name=embedder_name,
+            llm_name=llm_name,
+        )
+
+    async def add_new_category(
+        self,
+        *,
+        session_data: SessionData,
```
malatewang, line 624: Again, we may use set_id/set_name as the key

src/memmachine/semantic_memory/semantic_memory.py
o-love: Yeah. 
Just to note. There isn't a user create set call necesarilly. Instead sets are 'created' automatically when a message is ingested.
I'll add in a check to validate a set_id config exists when storing information to it. And if not add in the curent default embedder as the set_ids embedder.

src/memmachine/semantic_memory/semantic_session_manager.py
o-love: To clarify

Currently by accepting the `set_metadata_keys` it allows the user to call us once with the following.

```
    category_id = await memmachine.semantic_add_category(
        session_data=session_data,
        set_metadata_keys=["user_id"],
        category_name="profile",
        prompt="profile category",
        description="Category for user profile",
    )

```

Instead of breaking into first a `get_set_id` and then calling us again with the operation they want to do.

```
    set_id = await memmachine.semantic_get_set_id(
        session_data=session_data,
        is_org_level=False,
        metadata_tags=["user_id"],
    )
    category_id = await memmachine.semantic_add_category(
        set_id=<set_id>,
        category_name="profile",
        prompt="profile category",
        description="Category for user profile",
    )
```

Seperating out the set_id makes the individual functions simpler but requires the user to make multiple calls to MemMachine.

o-love: A question I have is how platform plans to enforce access control when the org/project info isn't part of the request.
If the user is operatng with Episode Id, Set Id... Will platform need to call us to try to determine if the user is authorized?

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
o-love: Prohibitting changes is already done at the `semantic_session_manager` level.

src/memmachine/main/memmachine.py
o-love: It''s used not to identify but to change. If you want to change the cattegory a feature belongs to.

src/memmachine/main/memmachine.py
o-love: Should the user be able to modify/configure the defualt prompts then?

My original implementation was to load the prompt files into the database and have the category be marked as shared in the database, making it read only. Then the user would need to clone it to be  able to modify the default prompt.

My final thought though was to keep the default prompts(categories) in the configuration and only make disabling of those default prompts dynamic. Which is what this function does.

When you get all your categories the defualt ones (those configured in the server config fiile) don't have an id.
If you want to modiify it you use this function to disble that default config and then you create a new category (now in the sql server) which you can modify.

src/memmachine/main/memmachine.py
o-love: Mentioned above. https://github.com/MemMachine/MemMachine/pull/882/files#r2686693160

The short versoin is that default categories are those that are defned in the config file, and not in the config database.

I debated loading the into the database but the final implementation doesn't.
This means that the database doesn't give these categories ids. Making them identifyable only by name.
As part of the config file they need to be unique by name so this stores into the database the name of prompts from the config file to not be used for this set_id,

src/memmachine/common/configuration/__init__.py
o-love: How would we want the defualt prompts to be used configured?
I try to keep backward compatability by allowing the config file to configure prompts for some basic isolations which is want this does. 

With the database then allowing dynamic configuration.

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
o-love: Related: 
https://github.com/MemMachine/MemMachine/pull/882#discussion_r2686722304
https://github.com/MemMachine/MemMachine/pull/882#discussion_r2686693160

The objective with disabling is to produce a list of Category names that are configured in the config file but shouldn't be used for that set id.
This enables the functionalty of allowing users to modify the default prompts by disbling it and "moving" it to the sql database.

sample_configs/episodic_memory_config.cpu.sample
o-love: I want to communicate that the configuration is applying a set of prompts to an isolatoin level.

Maybe rename `prompt` to `default_prompt` since these are now just the base prompt. Maybe `base_prompts`

Then for the level maybe rename `prompt` to  `isolation_default_prompt`. And the lower levels can be 
just their name 

Maybe:
```
isolation_base_prompts:
    project:
        - cat_a
        - cat_b
    organization:
        - cat_a
        - cat_c
```

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
o-love: A bit tricky. I don't want the user to need to call any `create_set` in order for messages to be ingested.

Instead I built it around an organization admin defining isolation which automatically ingest messages.

Because of that there isn't really a source of truth of what set_ids exist.
It's related to another comment you made as well.

I think the approach I'm goiing to take is the first message to be ingested into a set_id initializes the database with a default config.

src/memmachine/common/configuration/__init__.py
```diff
@@ -97,17 +101,13 @@ def _read_txt(filename: str) -> str:
...
-        default=[],
-        description="The default prompts to use for semantic role memory",
-    )
-    session: list[str] = Field(
+    org_level: list[str] = Field(
         default=["profile_prompt"],
-        description="The default prompts to use for semantic session memory",
+        description="The default prompts to use for semantic organization memory",
+    )
+    project_level: list[str] = Field(
```
jealous: Maybe just `project` is enough?

src/memmachine/common/configuration/__init__.py
```diff
@@ -97,17 +101,13 @@ def _read_txt(filename: str) -> str:
...
-    profile: list[str] = Field(
-        default=["profile_prompt", "writing_assistant_prompt", "coding_prompt"],
-        description="The default prompts to use for semantic user memory",
-    )
-    role: list[str] = Field(
-        default=[],
-        description="The default prompts to use for semantic role memory",
-    )
-    session: list[str] = Field(
+    org_level: list[str] = Field(
```
jealous: Maybe just `org`?  I am not sure.

src/memmachine/common/errors.py
```diff
@@ -95,3 +107,42 @@ class InvalidEmbedderError(MemMachineError):
...
+        metadata: Mapping[str, JsonValue],
+    ) -> None:
+        """Initialize with the org/project ids and metadata."""
+        self.org_id = org_id
+        self.project_id = project_id
+        self.metadata = metadata
+        self.is_org_level = is_org_level
+
+        super().__init__(
+            f"Passed metadata combination not enabled for org '{self.org_id}', project '{self.project_id}', is org level group '{self.is_org_level}', and metadata '{self.metadata}'"
```
jealous: `is org level group` in the message is somehow confusing.  May be just add `(org level)` after project id when it's true?

src/memmachine/common/errors.py
```diff
@@ -95,3 +107,42 @@ class InvalidEmbedderError(MemMachineError):
...
+        )
+
+    def __repr__(self) -> str:
+        """Return a helpful debug representation."""
+        return f"SetIdNotEnabledError('{self.org_id}', '{self.project_id}', {self.is_org_level}, {self.metadata})"
```
jealous: A boolean value in the repr would be confusing to me.

src/memmachine/semantic_memory/semantic_memory.py
```diff
@@ -245,25 +373,293 @@ async def update_feature(
...
+        new_category_name: str,
+    ) -> CategoryIdT:
+        logger.info("Cloning category %s to %s", category_id, new_category_name)
+
+        return await self._semantic_config_storage.clone_category(
+            category_id=category_id,
+            new_name=new_category_name,
+        )
+
+    async def disable_default_category(
```
malatewang: Should be just disable_category?

src/memmachine/semantic_memory/semantic_memory.py
```diff
@@ -245,25 +373,293 @@ async def update_feature(
...
+            set_id=set_id,
+            category_name=category_name,
+        )
+
+    async def delete_category_and_its_tags(
+        self,
+        *,
+        category_id: CategoryIdT,
+    ) -> None:
+        logger.info("Deleting category %s", category_id)
```
malatewang, line 685: Need to delete the data belonging to the category

src/memmachine/semantic_memory/semantic_memory.py
```diff
@@ -217,6 +343,8 @@ async def update_feature(
         tag: str | None = None,
         metadata: dict[str, str] | None = None,
     ) -> None:
+        logger.info("Updating feature %s", feature_id)
+
```
malatewang, line 359: Since the feature's category is updated, need to verify the category is valid here.

src/memmachine/semantic_memory/semantic_session_manager.py
```diff
@@ -44,116 +68,91 @@ class SessionData(Protocol):
...
+        def metadata(self) -> Mapping[str, JsonValue] | None: ...
 
-    @staticmethod
-    def is_session_id(_id: str) -> bool:
-        return _id.startswith(_SESSION_ID_PREFIX)
-
-    @staticmethod
-    def is_producer_id(_id: str) -> bool:
-        return _id.startswith(_USER_ID_PREFIX)
+    class DefaultType(Enum):
```
malatewang: Just name it ot: SetTypes?

src/memmachine/common/configuration/__init__.py
malatewang: In the configuration, it should be fine to call it default categories.

src/memmachine/semantic_memory/semantic_memory.py
o-love: In a way I want to keep the semantic memory decoupled from the category definitions. 
In which the features have a `category` that isn't tied down to a config definition.

The main use case for this is to allow users to directly create and manage semantic features without needing to register/define prompts for our LLM processing.
I think it could be useful to have MCP or user co manage features. 


It also takes a some steps to derive the set_id(s) corresponding to a category_id to then do the cleanup.
Compared to the user who likely already knows the set_ids they would want to cleanup. Since they would have needed those set_ids to discover the category_id.
The user can in one call using `FilterExpr` specify the set_ids and categories they want deleted in one go.

src/memmachine/semantic_memory/semantic_memory.py
o-love: Related: https://github.com/MemMachine/MemMachine/pull/882#discussion_r2690586659

src/memmachine/main/memmachine.py
```diff
@@ -483,5 +662,456 @@ async def delete_features(
...
+            description=description,
+        )
+
+    async def semantic_add_set_type_category(
+        self,
+        *,
+        set_type_id: str,
+        category_name: str,
+        prompt: str,
+        description: str | None = None,
```
malatewang: Besides the prompt, do we want to allow different embedder and models for different categories? By default, category inherits the embedder and model from set. But we may allow use to set a different one.

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,695 @@
...
+
+class BaseSemanticConfigStore(DeclarativeBase):
+    """Declarative base class for Semantic Config Store."""
+
+
+class SetIdResources(BaseSemanticConfigStore):
+    """Resource-level configuration associated with a set identifier."""
+
+    __tablename__ = "semantic.config.setidresources"
```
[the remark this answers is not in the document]
jealous: Which database does not support table name with dots?

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
```diff
@@ -0,0 +1,733 @@
...
+        tags = self.metadata_tags_sig.split(_TAG_SEP)
+
+        return OrgSetIdEntry(
+            id=str(self.id),
+            tags=tags,
+            is_org_level=self.org_level_set,
+        )
+
+
+_TAG_SEP = "\x1f"
```
jealous, line 217: Interesting, what does this character mean?  Why do you choose this character?

src/memmachine/semantic_memory/semantic_memory.py
```diff
@@ -217,21 +353,35 @@ async def update_feature(
...
+                embedding = np.array(e)
 
-            embedding = (await resources.embedder.ingest_embed([value]))[0]
-        else:
-            embedding = None
+            if category_name is not None:
+                category_names = {cat.name for cat in resources.semantic_categories}
+
+                if category_name not in category_names:
+                    from memmachine.common.errors import CategoryNotFoundError
```
jealous: Any special reason to import it locally?

src/memmachine/semantic_memory/semantic_session_manager.py
```diff
@@ -275,48 +292,373 @@ async def delete_feature_set(
...
+        return org_project
+
+    @staticmethod
+    def _generate_set_id(
+        *,
+        org_id: str,
+        project_id: str | None = None,
+        metadata: Mapping[str, JsonValue],
+    ) -> SetIdT:
+        org_project = SemanticSessionManager._org_set_id(
```
jealous: Replace `@staticmethod` with `@classmethod` and then you can call `cls._org_set_id`.

src/memmachine/common/configuration/__init__.py
o-love: Are you ok wiith the new names `default_org_categories` and `default_project_categories`?

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
o-love: It is the ASCII standard separator character.  It is non printable.

I prohibit tags from having it and use it as the csv seperator to store tags in a single string.

src/memmachine/main/memmachine.py
```diff
@@ -483,5 +662,477 @@ async def delete_features(
...
+
+        """
         semantic_session = await self._resources.get_semantic_session_manager()
         await semantic_session.delete_features(feature_ids)
+
+    async def add_feature(
+        self,
+        *,
+        set_id: SetIdT,
+        feature_metadata: dict[str, JsonValue] | None = None,
```
malatewang: It should be better to move the feature_metadata after value or tag parameters because set_id and category_name are used together as the location of the new feature.

src/memmachine/main/memmachine.py
```diff
@@ -483,5 +662,477 @@ async def delete_features(
...
+            tag=tag,
+            metadata=metadata,
+        )
+
+    async def create_semantic_set_type(
+        self,
+        *,
+        session_data: SessionData,
+        is_org_level: bool,
+        metadata_tags: list[str],
```
malatewang, line 847: We should allow a name and description here. Otherwise, it will be difficult for human to understand.

src/memmachine/main/memmachine.py
```diff
@@ -483,5 +662,477 @@ async def delete_features(
...
+        return await semantic_session.create_org_set_type(
+            session_data=session_data,
+            is_org_level=is_org_level,
+            metadata_tags=metadata_tags,
+        )
+
+    async def semantic_list_set_ids(
+        self,
+        *,
+        session_data: SessionData,
```
malatewang, line 879: I think we should provide API to list Set instead of just the IDs. ID itself is useless.

src/memmachine/semantic_memory/semantic_session_manager.py
```diff
@@ -275,48 +292,372 @@ async def delete_feature_set(
...
+
+        return set_ids
+
+    @dataclass(frozen=True, slots=True)
+    class SetIdEntry:
+        """Resolved set_id qualifiers for a session message."""
+
+        org_set_id: str
+        is_org_level: bool
+        tags: Mapping[str, str]
```
malatewang, line 343: It is better to have a name here. It will be more friendly to human

src/memmachine/semantic_memory/semantic_session_manager.py
```diff
@@ -275,48 +292,372 @@ async def delete_feature_set(
...
+        description: str | None,
+    ) -> CategoryIdT:
+        return await self._semantic_service.add_new_category_to_set_id(
+            set_id=set_id,
+            category_name=category_name,
+            prompt=prompt,
+            description=description,
+        )
+
+    async def add_new_org_set_category(
```
malatewang: Should we unify them to the same function?

src/memmachine/common/configuration/__init__.py
jealous: Fine with me.

src/memmachine/semantic_memory/config_store/config_store_sqlalchemy.py
jealous: Good to know.

src/memmachine/semantic_memory/semantic_session_manager.py
o-love: Not sure, they are slightly different.
In that one adds categories to set_ids and the other adds categories for isolations.

I can have it be done by an enum.

malatewang (approved):

PR #946, opened 2026-01-14
Upgrade ty to 0.0.11

o-love:
Upgrade ty to 0.11.
Fixes new default linters added to ty.


jealous (approved):

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -174,7 +174,7 @@ def __init__(self, params: Neo4jVectorGraphStoreParams) -> None:
         """Initialize the graph store with the provided parameters."""
         super().__init__()
 
-        self._driver = params.driver
+        self._driver: Any = params.driver
```
edwinyyyu: Does AsyncDriver not work here?

src/memmachine/common/reranker/identity_reranker.py
```diff
@@ -6,7 +6,7 @@
 class IdentityReranker(Reranker):
     """Reranker that returns candidates in their original order."""
 
-    async def score(self, _query: str, candidates: list[str]) -> list[float]:
+    async def score(self, query: str, candidates: list[str]) -> list[float]:  # noqa: ARG002
```
edwinyyyu: Prefer @typing.override over suppressing.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
o-love: It was giving errors with not matching the function overload for `execute_query`.
Looking into it deeper it was an issue with LiteralString. Where it accepts the types `Query` or `LiteralString` but we were passing dynamically created `str` type. I've wrapped our queries in neo4j `Query` so that they can accept strings.

https://typing.python.org/en/latest/spec/literal.html#literalstring

o-love: I've done a casting to `LiteralString` for the string formats that you use. Maybe long term we can refactor it to not build the queries dynamiclly but actually use constant strings and use anything dynamic as an argument/parameter.

```
def _neo4j_query(text: str) -> Query:
    return Query(cast(LiteralString, text))
```

Issue #940, opened 2026-01-12
[Bug]: Search memory got 500 error during load testing: neo4j.exceptions.ConnectionAcquisitionTimeoutError: failed to obtain a connection from the pool within 60.0s (timeout)


szou-mv:
### Describe the bug

During load testing, when there are 100 concurrent episodic memory search. About 30 requests failed with 500 error.

Request example:
```
POST /api/v2/memories/search
payload: {"org_id": "o_260112_1", "project_id": "2", "query": "has the user mentioned any specific database optimization tasks?", "top_k": 9, "types": ["episodic"]}
return code: 500
body: Internal Server Error
```

### Steps to reproduce

```
Traceback (most recent call last):
  File "/app/.venv/lib/python3.12/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/fastapi/applications.py", line 1139, in __call__
    await super().__call__(scope, receive, send)
  File "/app/.venv/lib/python3.12/site-packages/starlette/applications.py", line 107, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/app/.venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/app/.venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/app/.venv/lib/python3.12/site-packages/starlette/middleware/base.py", line 191, in __call__
    with recv_stream, send_stream, collapse_excgroups():
                                   ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/app/.venv/lib/python3.12/site-packages/starlette/_utils.py", line 85, in collapse_excgroups
    raise exc
  File "/app/.venv/lib/python3.12/site-packages/starlette/middleware/base.py", line 193, in __call__
[...]

Issue #837, opened 2025-12-19
[Bug]: Episodic memory search's respond time increases drastically when number of concurrency increases


szou-mv:
### Describe the bug

Following is the respond time chart of load test of POST /memories/search with types = ['episodic']. 
When number of user is less than 10, the response time doesn't significantly raise. 
However, when number of user reached about 12, the response time raised drastically. 
When the number of user reached 30, it took about half a minute to return.
And then as a result, we can only handle 2~3 search requests per second.
(Pure semantic memory search doesn't have this issue)

<img width="1513" height="966" alt="Image" src="https://github.com/user-attachments/assets/14e4aa75-49b2-49a7-a069-a8908119b159" />

### Steps to reproduce

1. Load testing tool: locust.
2. API:  POST /memories/search    payload: types = ['episodic']. 
3. There are 407 episodic memories in a single project.
4. Every user will send requests every 1~3 seconds.
5. Example payload:
`{"org_id": "org-bedbug", "project_id": "default", "query": "how often should I provide examples?", "top_k": 14, "types": ["episodic"]}`
6. Full API response time report.
[loadtest_20251219_120117_test_info.log](https://github.com/user-attachments/files/24254728/loadtest_20251219_120117_test_info.log)

### Expected behavior

If we can't handle big concurrency, maybe we should add throttling mechanisms.

### Environment

Memmachine Version: [65578fc](https://github.com/szou-mv/MemMachine/commit/65578fc91fa1b9077b1b3d1c25e1ef01b4fe1535)

Env: 
MacOS(M2chip 16GB ram) 
Embedding model and memmachine are deployed on the same host.
embedding model: embeddinggemma:latest (ollama)
No significant memory/GPU pressure found during load testing.

<img width="449" height="107" alt="Image" src="https://github.com/user-attachments/assets/fdf817ad-01ae-43cc-9019-819c51e84eab" />

<img width="826" height="244" alt="Image" src="https://github.com/user-attachments/assets/7644ddb9-b312-47ac-9056-690d3c5fdef8" />

sscargal:
Thanks @szou-mv. I did similar Locust testing on MemMachine v0.1.x some time ago and found similar scalability problems. Here is what I saw:

<img width="1024" height="759" alt="Image" src="https://github.com/user-attachments/assets/681f09e8-b42a-415f-9cfd-d03787f72e82" />

[...]

PR #949, opened 2026-01-14
[GH-909] Improve async summary performance.

### LINKED ISSUE #909: [Bug]: Improve Asynchronous Summary Processing in `ShortTermMemory.add_episodes`
### Describe the bug

During load testing in #837, I identified a potential bottleneck that limits Locust load generation.

Image

Image
Image
Image
Image

### Steps to reproduce

Use the locust and profile scripts from #837 

### Expected behavior

- Reduce or remove serialization locks in highly use code paths. 
- Design locks to hold as short a time as possible and only in critical sections only.

### Environment

- OS: Ubuntu 25.04
- MemMachine v0.2.2

### Additional context

The `ShortTermMemory.add_episodes` method blocks the caller if an eviction is triggered and a previous summary is still running.

jealous:
### Purpose of the change

fixes performance degradation in short-term memory caused by blocking summary generation during episode addition. 

### Description

Previously, triggering a new summary while a previous one was still running would block the add operation and slow down ingestion.

The fix extracts summary logic into a dedicated consolidator class that runs consolidation asynchronously and avoids blocking writes.

Key changes:

- Move episode summarization logic into a new consolidator class
- Run consolidation in a background routine instead of blocking add operations
- Cache newly added episodes when consolidation is already running and include them in the next consolidation pass
- Ensure queries wait for consolidation to finish so results are always stable and consistent

### Fixes/Closes

Fixes #909 

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code


src/memmachine/common/episode_store/episode_model.py
```diff
@@ -60,3 +63,43 @@ class Episode(BaseModel):
     def __hash__(self) -> int:
         """Hash an episode by its UID."""
         return hash(self.uid)
+
+
+def episodes_to_string(
```
edwinyyyu, line 68: This method was public. Is it safe to move?

tests/memmachine/common/episode_store/test_episode_model.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def test_episodes_to_string_with_episode_response(base_episode_data):
+    """Verify it works with EpisodeResponse instances (score included)."""
+    # Since EpisodeResponse inherits from EpisodeEntry/Episode, we mock it similarly
+    er = EpisodeResponse(**base_episode_data, score=0.95)
+    result = episodes_to_string([er])
+
+    assert "user_1" in result
+    assert '"Hello world"' in result
```
edwinyyyu: Does it make sense to test these behaviors? None of these should be necessary for a correct implementation of the function.

edwinyyyu:
Did you manually verify the performance improvement?

tests/memmachine/common/episode_store/test_episode_model.py
jealous: OK.  I am keeping the test for `Episode` and `EpisodeResponse` only.  Just to make sure it works as expected.

src/memmachine/common/episode_store/episode_model.py
jealous: Yes.  The reason to move this is that it's duplicated in two places.  I think I have updated all the references.

jealous:
Here is my test code, insert 100 entires each has 1000 random characters to episodic memory only:
```python
def run():
    client = MemMachineClient(base_url="http://localhost:8080")
    project = client.get_project(
        org_id="performance_test",
        project_id="multi_add",
    )
    memory = project.memory(
        metadata={
            "group_id": "test_group",
            "agent_id": "agent1",
            "user_id": "user1",
        }
    )
    start = time.time()
    characters = string.ascii_letters + string.digits + string.punctuation
    for i in range(100):
        result = memory.add(
            content=''.join(random.choices(characters, k=1000)),
            memory_types=[MemoryType.Episodic],
        )
        assert len(result) == 1
        if i % 10 == 0:
            print(f"Added entry i={i}")
    duration = time.time() - start
    print(f"Added 1000 entries in {duration:.2f} seconds ({1000/duration:.2f} ops/sec)")
```

I run the test on my macbook.

Before the change:
```
Added 1000 entries in 153.24 seconds (6.53 ops/sec)
```

After the update:
```
Added 1000 entries in 31.84 seconds (31.40 ops/sec)
```

About 5 times faster.

sscargal:
@jealous @edwinyyyu This looks extremely promising!!

I used the Locust load script from #837 and ran the Episodic Write-only test to produce the following results:

image

Until this PR, I could only achieve up to ~20 requests per second (RPS) with 50 users - see the results in #837. With this fix, I'm getting up to 150 RPS at 50 users (7.5x increase). The RPS looks stable, which is a good sign. 

The Locust workload starts to fail when we hit 100 users with "neo4j - Failed to read four byte Bolt handshake response from server ResolvedIPv4Address(('172.18.0.3', 7687)) (deadline Deadline(timeout=60.0))". This is not the fault of MemMachine or this PR; it's actually a good sign, as it shows the bottleneck is now in Neo4j. See #940 for similar issues. We have tunables available to help, so I'll continue that effort to see if I can successfully complete the load test with 500 users.

In the future, we may need to consider a back-off algorithm when we detect database/storage latency issues or timeouts.

src/memmachine/episodic_memory/short_term_memory/short_term_memory.py
```diff
@@ -522,3 +456,154 @@ def _compute_episode_length(self, episode: Episode) -> int:
...
+    async def _run_summary_loop(self) -> None:
+        """Background worker that drains the pending episodes."""
+        while True:
+            # 1. Grab the current batch and clear the buffer
+            async with self._lock:
+                if not self._pending_episodes:
+                    # No more episodes to process, exit the worker
+                    break
+
+                batch_to_process = self._pending_episodes[:]
```
malatewang, line 529: The problem here is the episodes can accumulate. Eventually it can exceed the model context window size.

jealous: I updated the logic so it will recursively split the episodes into 2 parts until they fit into the context window.  If single episode + summary has already exceeds the context window, I have no other choice but to drop that episode.

jealous:
@malatewang , there is another push for the fix in the meeting.  I think maybe we can merge this first because it's helpful for batch inserting.   And we can fix other issued caused by the change later.

PR #969, opened 2026-01-22
Define vector store interface

edwinyyyu:
### Purpose of the change

Whereas:
- Graph features are not heavily used now.
- Current memory components primarily depend on vector database features.
- It is more difficult to work with a completely schemaless data store.
- There may be asynchronous setup logic for data storage.
- Querying logic for different search methods can be better unified.
- The existing `VectorGraphStore` interface is harder to implement with a pure vector database.
Thus:
- Add a simpler interface to implement for different database providers.

### Description

Add a new ABC named `VectorStore` to wrap vector databases.

Will be part of larger refactor.
- If keeping support for VectorGraphStore, we can do any of the following:
  - VectorGraphStore-based VectorStore (VectorGraphStore will be redefined but externally it will be mostly the same)
  - new memory implementation
    - based on what exists in the config, route to either new or old implementation

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

No tests for interface.

- [ ] Unit Test
- [ ] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### DIFF
--- a/src/memmachine/common/vector_store/vector_store.py
+++ b/src/memmachine/common/vector_store/vector_store.py
@@ -168,7 +168,7 @@ class VectorStore(ABC):
     @abstractmethod
     async def get_collection(self, collection_name: str) -> Collection:
         """
-        Get a collection from the vector store.
+        Get a collection from the vector store. Handle persists across deletion and creation.
 
         Args:
             collection_name (str):


edwinyyyu:
@o-love Any need for pure filtered search without similarity score? I think just SQL would be better for that. I've removed the filter-only search.

src/memmachine/common/vector_store/vector_store.py
```diff
@@ -0,0 +1,166 @@
...
+
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    async def query(
+        self,
+        *,
+        collection: str,
+        query_vector: list[float],
```
o-love: I was wondering if its possible to accept and array as well as a list with the `Sequence[float]` type.

I feel like we are generally overusing/enforcing python's lists instead of using arrays (either numpy or stdlib arrays).

src/memmachine/common/vector_store/vector_store.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    async def get(
+        self,
+        *,
+        collection: str,
+        record_uids: Iterable[str],
+    ) -> list[Record]:
```
o-love: Since steve has mentioned in the future supporting streaming results through http instead of fully loading before returning. 

Would it make sense to start converting the return types of our functions into Iterables?

o-love: For now the implementations can still return a list.

Sql backends do support streaming by returning an iterable. Some vector stores might too.

o-love: If not returning Sequence would also be nice

edwinyyyu: I think Iterable would be good.

src/memmachine/common/vector_store/vector_store.py
edwinyyyu: Numpy arrays don't work with Sequence[float].

I was thinking to use list[float] because it's easy to guarantee behaviors work.

If we allow NumPy/PyTorch, it's more work to check the type at runtime, since the type hint would say nothing about the dimensions of the NDArray/Tensor.

edwinyyyu: I've changed it to Sequence to support tuples and array.array.

edwinyyyu:
Before merge, I want to review the index creation parameters.

jealous:
@edwinyyyu do you want to merge it?

edwinyyyu:
I reviewed several vector databases and it appears that there's a mix of ones that can take a collection name directly for performing data operations and ones that return a handle to a collection.

I will change this to use handles.

edwinyyyu:
Making it very barebones because there's a lot of fragmentation between different vector database providers.

edwinyyyu:
Notes:
- Collection doesn't necessarily map to native collections for each vector database provider. See recommended multitenancy approach for each.
- The Collection object (a handle) should be able to recreate itself if a provider-native collection of the same name gets deleted and recreated. This allows consistent behavior across providers (Chroma Python Collection object seems to persist, could be a bug; Qdrant only uses collection names).
- Some provider(s) don't support add operation, only upsert/put (Amazon S3 Vector, Qdrant).
- To get/delete by filter, an external database (likely SQL) will be maintained.
- Consumer is responsible for ensuring that a collection is only created once.

PR #1028, opened 2026-02-02
docs: add performance tip for large memory stores

dxu104:
Add a note in MCP documentation advising users to use faster LLM models (e.g., Claude Haiku instead of Sonnet) when experiencing slow response times with large memory stores.

### DIFF
--- a/docs/examples/mcp.mdx
+++ b/docs/examples/mcp.mdx
@@ -336,7 +336,7 @@ you MUST actively use MemMachine MCP on EVERY Question
 You can also check the logs by clicking on `View Logs` button in the `Developer` tab.
 
 <Note>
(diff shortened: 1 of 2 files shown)


sscargal (approved):
@dxu104 Is there a downside to using Haiku vs Sonet with respect to the quality of results? Rather, what are the Pros and Cons? Performance is definitely a reason to switch. If there are no Cons, we should recommend that users use the lower-parameter, lower-cost, faster-response models.

sscargal:
The `ruff` failures have nothing to do with this PR.

dxu104:
> The `ruff` failures have nothing to do with this PR.

Thanks for the reminder. Do you know why the CI failed?

sscargal:
`ruff` was automatically upgraded to 0.15.0 via GitHub Workflows. This new version introduced new rules, which are the ones that are failing. I filed #1037 to resolve the problem.

dxu104:
I will discard this PR because, based on today's test, both Haiku and Sonnet models are very fast with exactly same settings as yesterday.  I can get responses within 5 seconds on multiple different queries I sent via Claude Desktop stdio MCP.  @jealous @sscargal

PR #1051, opened 2026-02-04
Switch semantic to stream results from database

o-love:
Switches semantic memory to stream results from the database and operate on the stream.
Reducing memory overhead from rebuilding lists multiple times.

Will also enable any future work on making the REST api support streaming results.


o-love:
Superseded by #1239

PR #1047, opened 2026-02-04
Add semantic memory management and episodic memory configuration API

jealous:
### Purpose of the change

Add REST API endpoints and SDK client methods for managing semantic memory configuration (set types, sets, categories, tags) and episodic memory configuration (enable/disable episodic, LTM, STM per session).

### Description

This PR exposes the MemMachine semantic memory configuration methods and adds new episodic memory configuration capabilities through the REST API and SDK client.

**Set Type & Set Management:**
- `create_semantic_set_type`: Create new set types with metadata tags
- `list_semantic_set_types`: List all set types
- `delete_semantic_set_type`: Delete a set type
- `get_semantic_set_id`: Get or create a set ID
- `list_semantic_set_ids`: List all sets with optional metadata filter
- `configure_semantic_set`: Configure embedder/LLM for a set

**Category Management:**
- `get_semantic_category`: Get category by ID
- `add_semantic_category`: Add category to a set
- `add_semantic_category_template`: Add category template to set type
- `list_semantic_category_templates`: List templates for a set type
- `disable_semantic_category`: Disable category for a specific set
- `get_semantic_category_set_ids`: Get set IDs using a category
- `delete_semantic_category`: Delete a category

**Tag Management:**
- `add_semantic_tag`: Add tag to a category
- `delete_semantic_tag`: Delete a tag

**Episodic Memory Configuration:**
- `POST /api/v2/config/memory/episodic/get` - Get current episodic memory settings
- `POST /api/v2/config/memory/episodic` - Configure episodic memory (enable/disable episodic, LTM, STM per session)

### Fixes/Closes

N/A

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (does not change functionality, e.g., code style improvements, linting)
- [ ] Documentation update
- [ ] Project Maintenance (updates to build scripts, CI, etc., that do not affect the main project)
- [ ] Security (improves security without changing functionality)

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test
- [ ] End-to-end Test
- [ ] Test Script (please provide)
- [ ] Manual verification (list step-by-step instructions)

**Test Results:**

```
$ uv run pytest tests/memmachine/server/api_v2/test_router.py tests/memmachine/rest_client/test_memory.py -v -m "not integration"
[...]

PR #1061, opened 2026-02-06
Optimize ChatGPT migration: disable STM during import, fix `user_only` filtering, and make Docker builds work on macOS.

dxu104:
## Purpose of the change

Optimize ChatGPT migration: disable STM during import, fix `user_only` filtering, make Docker builds work on macOS, and add safe embedder defaults.

* **migration.py**

  * Use the new **per-session SDK API** to disable short-term memory (STM) summarization during import and re-enable it after completion.
  * This avoids unnecessary LLM summarization calls during episodic-memory-only imports, reducing cost from **~$100 → ~$0.10**.
  * Default `org_id` / `project_id` to `"memory_migration"` for clearer session isolation.
  * Fix `user_only` filter to exclude **all non-user messages** (assistant/system/tool/etc.), not just assistant messages.

* **parsers/openai_parser.py**

  * Fix `user_only` filter logic to use `role != "user"` for consistency.

* **sample_configs/episodic_memory_config.cpu.sample**, **sample_configs/episodic_memory_config.gpu.sample**

  * Add `max_input_length: 2048` to all embedder configs to prevent OpenAI embedding API 400 errors when messages exceed token limits.
  * Ensures new users get safe defaults out of the box via `./memmachine-compose.sh`.

* **build-docker.sh**

  * Use `sed -E` (extended regex) instead of GNU `\+` syntax for **macOS BSD sed compatibility**, fixing local Docker builds on macOS.

---

## Description

This PR improves the ChatGPT memory migration tool to minimize LLM cost and improve first-time user experience.

### Key Improvements

1. **Disable STM summarization during import**

   * The migration process only imports **episodic memory** using embeddings and does not require LLM summarization.
   * Previously, STM summarization was triggered on every imported message, causing large unnecessary LLM spend.
   * This PR disables STM at the session level before import and re-enables it afterward, using the new per-session SDK API:

     * `sdk_client.get_or_create_project().memory().configure_episodic_memory(short_term_memory_enabled=False)`

2. **Fix incorrect `user_only` filter**

   * The prior filter only excluded `"assistant"` messages.
   * This PR correctly excludes **all non-user roles**, including `assistant`, `system`, and tool messages.

3. **Add safe defaults for embedding input size**

   * During testing, some messages exceeded the embedding model token limit (**8,192 tokens**), causing `400 Bad Request` errors from the OpenAI API.

   * Our code measures input length in **Unicode characters**, not tokens. In the worst case (for example, Chinese text), **1 character can cost up to ~4 tokens**.

   * This PR adds:

     ```yaml
     max_input_length: 2048
     ```

     to all embedder configs in:

     * `sample_configs/episodic_memory_config.cpu.sample`
     * `sample_configs/episodic_memory_config.gpu.sample`

   * This ensures input stays safely within the 8,192 token limit across all four recommended embedding models.

   * When a message exceeds this limit, it is automatically **chunked** into smaller pieces:

     * each chunk is embedded separately
     * embeddings are averaged into a single vector per message

   * This is added as a default so new users get safe behavior out of the box via `./memmachine-compose.sh`.

4. **Fix Docker build script on macOS**

   * `build-docker.sh` used GNU `sed` syntax (`\+`) which fails on macOS.
   * This PR switches to `sed -E`, which works on both macOS (BSD sed) and Linux (GNU sed).

### Dependencies

* Depends on the new per-session episodic memory configuration API:

  * `POST /api/v2/memory/episodic/config`
  * Introduced in **PR #1047**

---

## Fixes/Closes

* Fixes #<issue-number-if-applicable>

---

## Type of change

* [x] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [x] Project Maintenance (updates to build scripts, CI, etc., that do not affect the main project)

---

## How Has This Been Tested?

* [x] Manual verification (list step-by-step instructions)

### Steps

1. Export ChatGPT conversations to `conversations.json`

   * Dataset: **2,304 conversations**

2. Run migration:

   ```bash
   uv run python migration.py -i extracted/conversations.json -v --workers 32
   ```

3. Verify STM is disabled before import and re-enabled after completion.

4. Verify migration stats show **0 failures**.

5. Verify embedder config defaults include `max_input_length: 2048` and that long messages no longer trigger embedding API errors.

### Test Results

```json
{
  "total_conversations": 2304,
  "total_messages": 20884,
  "successful_messages": 20884,
  "failed_messages": 0,
  "duration_seconds": 137.36
}
```

* **0 failures** across **20,884 messages** in **2,304 conversations**
* Total import time: **~2.3 minutes** (32 workers)
* Cost: **~$0.10**

  * Only OpenAI embeddings (`text-embedding-3-small`) were used
  * No LLM costs since STM summarization is disabled at session level
  * No semantic memory import performed

---

## Testing the imported memory

To test the imported memory with **Claude Desktop**, follow the MCP integration guide:

* `docs/install_guide/integrate/claude_code_mcp.mdx`

Make sure the `MM_ORG_ID` and `MM_PROJ_ID` in your `claude_desktop_config.json` match the values used during import.

The migration tool defaults to `memory_migration` for both:

```json
"MM_PROJ_ID=memory_migration",
"MM_ORG_ID=memory_migration",
```

If you used custom `--org-id` or `--project-id` values during migration, update these env vars accordingly so Claude Desktop reads from the correct memory project.

---

## Checklist

* [x] I have signed the commit(s) within this pull request
* [x] My code follows the style guidelines of this project (See STYLE_GUIDE.md)
* [x] I have performed a self-review of my own code
* [x] My changes generate no new warnings

---

## Maintainer Checklist

* [ ] Confirmed all checks passed
* [ ] Contributor has signed the commit(s)
* [ ] Reviewed the code
* [ ] Run, Tested, and Verified the change(s) work as expected

---

## Screenshots/Gifs

N/A

---

## Further comments

* This PR relies on the new per-session episodic memory configuration API added in **PR #1047**.
* The STM disable/enable behavior is intentionally scoped to the migration session to avoid impacting normal runtime behavior.
* The `max_input_length: 2048` default prevents embedding API errors caused by token-length mismatches and improves first-run reliability for new users.
* `build-docker.sh` change is cross-platform: `sed -E` works on both macOS and Linux.

---

### DIFF
--- a/tools/chatgpt2memmachine/requirements.txt
+++ b/tools/chatgpt2memmachine/requirements.txt
@@ -1,3 +1,2 @@
-memmachine
 requests>=2.31.0
 tqdm>=4.65.0


tools/chatgpt2memmachine/migration.py
```diff
@@ -1,29 +1,30 @@
...
 from concurrent.futures import ThreadPoolExecutor, as_completed
-from datetime import timezone
 
 from parsers import get_parser
 from restcli import MemMachineRestClient
 from tqdm import tqdm
-from utils import parse_time, format_timestamp_iso8601, load_run_id_file
+from utils import format_timestamp_iso8601, load_run_id_file, parse_time
+
+from memmachine import MemMachineClient
```
jgong: It looks good to me, nice work! Regarding the MemMachineClient import, since MemMachineRestClient already implements REST logic. Maybe we can consolidate config logic by adding a function to restcli.py. That way, we only have to manage one client for requests.

tools/chatgpt2memmachine/requirements.txt
```diff
@@ -1,2 +1,3 @@
+memmachine
```
jgong: remove requirement if memmachine package is not required now.

tools/chatgpt2memmachine/migration.py
```diff
@@ -178,43 +227,67 @@ def load_and_extract(self):
...
-            extracted_file_name = (
-                f"{self.input_file_base_name}_{conv_id}_extracted.json"
+        # Load full dataset once (only if we need to extract new conversations)
+        needs_extraction = any(
+            not os.path.exists(
+                os.path.join(
+                    self.extract_dir,
+                    f"{self.input_file_base_name}_{conv_id}_extracted.json",
+                )
             )
```
jgong, line 245: Hi @dxu104 , thanks for this change to reduce time of extraction. 
However when i check this code logic again, i find that `needs_extraction` will be `False` if all conversation files have been extracted, and then `all_data` would be `None` in this case.
What if the files has been extracted, and user wants to import again?

tools/chatgpt2memmachine/migration.py
```diff
@@ -178,43 +227,67 @@ def load_and_extract(self):
...
-                messages = self.parser.load(self.input_file, filters=filters)
-                self.conversations[conv_id] = messages
-                # Dump extracted messages for this conversation
-                self.parser.dump_data(
-                    messages, output_format="json", outfile=extracted_file
+            for conv_id in conv_ids_to_process
+        )
+
+        all_data = None
+        if needs_extraction:
```
jgong, line 250: Follow up with above comment, if all conversations have been extracted into file, all_data will not be loaded. Then the parallel extraction may encounter errors.

tools/chatgpt2memmachine/migration.py
```diff
@@ -178,43 +227,67 @@ def load_and_extract(self):
...
+        # Parallel extraction using ThreadPoolExecutor
+        workers = min(self.max_workers, len(conv_ids_to_process))
+
+        if self.verbose:
+            print(f"Extracting conversations with {workers} worker(s)...")
+
+        with ThreadPoolExecutor(max_workers=workers) as executor:
+            futures = {
+                executor.submit(
+                    self._extract_single_conversation, conv_id, all_data
```
jgong, line 266: Check this part with all conversation files have been extracted.

tools/chatgpt2memmachine/migration.py
dxu104: Thanks @jgong  for the review! Let me explain the current behavior:
                                                                                                                                             
  When the user imports again and all extracted files already exist:
                                                                                                                                             
  1. needs_extraction = False → all_data = None (skip loading the large original JSON)                                                       
  2. Each worker enters _extract_single_conversation, checks os.path.exists(extracted_file) → True
  3. It loads messages from the cached extracted file directly, never touches all_data

  Both code paths are safe:

  - All cached files exist → needs_extraction = False → all_data = None → every worker takes Path A (load from cache, never touches all_data)
   ✅
  - Some cached files missing → needs_extraction = True → all_data = loaded → workers with missing cache take Path B (extract from all_data)
  ✅

  I tested this case (all files already extracted, user imports again). As you can see in the screenshot below, "Loading full dataset for
  extraction..." does not appear in the log, which confirms all_data was never loaded or accessed. All 2304 conversations loaded from cached
  files without error.
  image 
  Could you share a specific scenario where this would break? I'd like to understand your concern better so I can learn from it. Thanks!

jgong: Thanks @dxu104 for the case and results, it looks good to me now.

PR #1063, opened 2026-02-06
Add semantic similarity to top level interface

o-love:
Connects the similarity threshold of `memmachine` to the similarity argument for semantic memory.
This allows users to filter by similarity when using semantic memory.

### DIFF
--- a/src/memmachine/main/memmachine.py
+++ b/src/memmachine/main/memmachine.py
@@ -603,7 +603,7 @@ class MemMachine:
                     session_data=session_data,
                     set_metadata=set_metadata,
                     limit=limit,
-                    min_distance=score_threshold,
+                    distance_threshold=score_threshold,
                     search_filter=property_filter,
                 )
             )
--- a/src/memmachine/main/memmachine.py
+++ b/src/memmachine/main/memmachine.py
@@ -603,6 +603,7 @@ class MemMachine:
                     session_data=session_data,
                     set_metadata=set_metadata,
                     limit=limit,
+                    min_distance=score_threshold,
                     search_filter=property_filter,
                 )
             )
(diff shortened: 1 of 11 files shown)


src/memmachine/main/memmachine.py
```diff
@@ -603,6 +603,7 @@ async def query_search(
                     session_data=session_data,
                     set_metadata=set_metadata,
                     limit=limit,
+                    min_distance=score_threshold,
```
malatewang: Can you explain what is the score or distance?

o-love: Both refer to the cosine similarity metric that a result needs in order to be included.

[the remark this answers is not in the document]
edwinyyyu: Why is this resolved?

edwinyyyu (changes requested):
Currently score is usable for reranker score in episodic memory.

min_distance (interpreted as minimum distance) means that we are including all of the worst results and excluding all of the best results. Lower distance is better. Higher similarity is better. Please rename.

To share score thresholds between different memory types using different databases, we should convert all distances/similarities to the same type of score i.e. use the same embedder (and for embedders, normalize the results from database queries to use the same formula for computation) or use the same reranker (easier).

See #1066 for a reference of formulas used for score computation.

jealous (approved):

src/memmachine/main/memmachine.py
```diff
@@ -603,6 +603,7 @@ async def query_search(
                     session_data=session_data,
                     set_metadata=set_metadata,
                     limit=limit,
+                    distance_threshold=score_threshold,
```
edwinyyyu, line 606: ~~We interpret score as higher is better right now.~~

~~Please either rename or redefine score or distance, and ensure that it works with rerankers which return 0-1.~~

o-love:
Suggestions were opened regarding unifying similarity system.
Awaiting unified vector store interface to study

src/memmachine/main/memmachine.py
edwinyyyu: higher_is_better bool is now on SimilarityMetric

PR #1082, opened 2026-02-09
Improve Profile Memory prompt

o-love:
Improvement to the semantic profile memory prompt to increase feature scope and improve feature extraction


edwinyyyu:
Do you have before-and-after examples of how this improves things?

PR #1022, opened 2026-02-02
Future filtering behavior

edwinyyyu:
### Purpose of the change

Nobody can understand the filtering system because it is different across the different memory subcomponents and there is no central design documentation for this. The design needs to be defined and unified.

This design is based on surveying database providers and finding common functionality and limitations.

Valid stored attribute values and values used for filtering have different semantics and support different types.

Existing points in code take shortcuts to list[int] and list[str] by doing list[FilterablePropertyValue] which has the wrong semantics and overpromises compatibility.

None/NULL should not be used to represent a type. Instead it should represent nullability.

Implement NOT, NOT IN, and not equals (!= or <>) filter expressions.
Fix numeric comparisons in SQLAlchemy--everything was converted to string.

Having filterable and non-filterable metadata is hard to reason about.

### Description

Written using Claude Code.

Old FilterablePropertyValue:
- bool | int | float | str | datetime | None

Supported AttributeValue:
- bool | int | float | str | datetime

Supported FilterValue:
- bool | int | float | str | datetime | list[int] | list[str]

Supported operations:
- IS NULL, IS NOT NULL
- <, >, <>/!=, <=, >=, =
- IN, NOT IN

Explicitly unsupported:
- ARRAY CONTAINS (we don't store lists/arrays)
- STRING CONTAINS (not commonly supported by databases)

Remove non-filterable metadata. New filterable metadata is just called 'metadata'.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (does not change functionality, e.g., code style improvements, linting)
- [ ] Documentation update
- [ ] Project Maintenance (updates to build scripts, CI, etc., that do not affect the main project)
- [ ] Security (improves security without changing functionality)

### How Has This Been Tested?

- [x] Unit Test
- [ ] Integration Test
- [ ] End-to-end Test
- [ ] Test Script (please provide)
- [ ] Manual verification (list step-by-step instructions)

### Checklist

- [x] I have signed the commit(s) within this pull request
- [x] My code follows the style guidelines of this project (See STYLE_GUIDE.md)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added unit tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
[...]

PR #1090, opened 2026-02-10
Unify filtering behavior

edwinyyyu:
### Purpose of the change

Reduce scope of #1022.

Nobody can understand the filtering system because it is different across the different memory subcomponents and there is no central design documentation for this. The design needs to be defined and unified.

This design is based on surveying database providers and finding common functionality and limitations.

Valid stored attribute values and values used for filtering have different semantics and support different types.

Existing points in code take shortcuts to list[int] and list[str] by doing list[FilterablePropertyValue] which has the wrong semantics and overpromises compatibility (technically implementable as list[FilterablePropertyValue] but it is not simple/efficient).

None/NULL should not be used to represent a type. Instead it should represent nullability.

Implement NOT, NOT IN, and not equals (!= or <>) filter expressions.

Fix numeric comparisons in SQLAlchemy (used for semantic memory)--everything was converted to string.
Fix metadata field used for ShortTermMemory--should be filterable_metadata.
Fix LongTermMemory to also take `metadata.` prefix in addition to `m.` for user-defined metadata filtering.

### Description

Claude coded.

Replace FilterablePropertyValue by FilterValue, PropertyValue.

Fix bugs.

Implement new filter operations.
FilterExpr will always use != instead of <>, but the query language allows either.
Reject <, >, <=, >= for bools.

Out of scope:
- rejecting/interpreting <, >, <=, >= for bool fields and values -- undefined behavior
- supporting `NOT bool_thing`, `bool_thing1 AND bool_thing2`, `bool_thing1 OR bool_thing2` for bool fields and values
- timezone normalization

### Intentional behavior changes

- **ShortTermMemory filtering now reads from `filterable_metadata` instead of `metadata`**: `episodic_memory.py` auto-copies eligible values from `metadata` to `filterable_metadata` on add, so this only affects episodes already in the STM buffer before this change that only have `metadata` populated (not `filterable_metadata`). In practice STM is ephemeral (in-memory buffer), so this is not a concern for persisted data.
- **In-memory semantic test storage now compares metadata values with native types instead of string-coerced**: The previous `_normalize_metadata_value` converted everything to strings before comparison, which was the bug being fixed. Metadata values are now compared using their actual types (int, float, bool, str).

### Type of change

Should be non-breaking for existing data.

- [ ] Documentation update

### How Has This Been Tested?

- [ ] Unit Test
- [ ] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### DIFF
--- a/src/memmachine/common/filter/filter_parser.py
+++ b/src/memmachine/common/filter/filter_parser.py
@@ -251,7 +251,7 @@ class _Parser:
     def _parse_in_value(self) -> int | str:
         """Parse a single value inside an IN list (only int and str allowed)."""
         value = self._parse_value()
-        if not isinstance(value, (int, str)):
+        if isinstance(value, bool) or not isinstance(value, (int, str)):
             raise FilterParseError(
                 f"IN lists only support int and str values, got {type(value).__name__}"
             )
--- a/tests/memmachine/common/filter/test_filter_parser.py
+++ b/tests/memmachine/common/filter/test_filter_parser.py
@@ -524,6 +524,21 @@ def test_map_filter_fields_nested() -> None:
     assert result.expr.right == Comparison(field="prefix_b", op="=", value=2)
 
 
+def test_parse_filter_in_rejects_bool() -> None:
+    with pytest.raises(FilterParseError, match="IN lists only support int and str"):
+        parse_filter("flag IN (true, false)")
+
+
+def test_parse_filter_in_rejects_float() -> None:
+    with pytest.raises(FilterParseError, match="IN lists only support int and str"):
+        parse_filter("x IN (1.5, 2.5)")
+
+
+def test_parse_filter_in_rejects_mixed_types() -> None:
+    with pytest.raises(FilterParseError, match="Mixed types in IN list"):
+        parse_filter("x IN (1, 'two')")
+
+
 def test_map_filter_fields_with_normalize() -> None:
     """Test map_filter_fields combined with normalize_filter_field."""
     expr = And(
--- a/tests/memmachine/common/filter/test_sql_filter_util.py
+++ b/tests/memmachine/common/filter/test_sql_filter_util.py
@@ -252,6 +252,11 @@ def test_not_and_compound(session):
     assert result == {"alpha", "beta", "delta"}
 
 
+def test_or_simple(session):
+    result = _query_names(session, "m.tag = 'a' OR m.tag = 'c'")
+    assert result == {"alpha", "gamma"}
+
+
 def test_not_or_compound(session):
     # NOT (tag = 'a' OR tag = 'b') => NOT(alpha, beta) => gamma, delta
     result = _query_names(session, "NOT (m.tag = 'a' OR m.tag = 'b')")
--- a/tests/memmachine/common/vector_graph_store/test_neo4j_vector_graph_store.py
+++ b/tests/memmachine/common/vector_graph_store/test_neo4j_vector_graph_store.py
@@ -16,7 +16,13 @@ from memmachine.common.filter.filter_parser import (
     Comparison as FilterComparison,
 )
 from memmachine.common.filter.filter_parser import (
+    In as FilterIn,
+)
+from memmachine.common.filter.filter_parser import (
     IsNull as FilterIsNull,
+)
+from memmachine.common.filter.filter_parser import (
+    Not as FilterNot,
 )
 from memmachine.common.filter.filter_parser import (
     Or as FilterOr,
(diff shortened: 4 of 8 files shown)


src/memmachine/common/filter/filter_parser.py
```diff
@@ -16,13 +17,31 @@ class FilterExpr(Protocol):
...
 
+ComparisonOp = Literal["=", "!=", ">", "<", ">=", "<="]
+
+
 @dataclass(frozen=True)
 class Comparison(FilterExpr):
-    """Filter comparison of a field against a value or list of values."""
+    """Scalar comparison of a field against a value."""
+
+    field: str
```
malatewang, line 27: Should be field be PropertyValue?

src/memmachine/common/filter/filter_parser.py
```diff
@@ -16,13 +17,31 @@ class FilterExpr(Protocol):
...
+    op: ComparisonOp
+    value: PropertyValue
+
+
+@dataclass(frozen=True)
+class In(FilterExpr):
+    """Membership test of a field against a list of values."""
+
+    field: str
```
malatewang, line 36: field should be int | str ?

src/memmachine/common/filter/filter_parser.py
edwinyyyu: Field is the name of the field.

In a query: "our_field_name > [PropertyValue value here]"

src/memmachine/common/filter/filter_parser.py
edwinyyyu: field is the name of the field

edwinyyyu:
@sscargal  Set as high because I would like this to be merged before finalizing the database abstractions, or else this PR would become larger and more complex.

src/memmachine/common/filter/filter_parser.py
```diff
@@ -154,43 +191,54 @@ def _parse_primary(self) -> FilterExpr:
...
-        op_tok = self._accept("EQ", "GE", "LE", "GT", "LT")
+        op_tok = self._accept("EQ", "NE", "GE", "LE", "GT", "LT")
         if op_tok:
-            # field =/>=/>/</<= value
             value = self._parse_value()
-            op = {"EQ": "=", "GE": ">=", "LE": "<=", "GT": ">", "LT": "<"}[op_tok.type]
-            return Comparison(field=field, op=op, value=value)
+            return Comparison(field=field, op=_SCALAR_OPS[op_tok.type], value=value)
 
+        # IN / NOT IN
```
jealous, line 211: The `# type: ignore` here hints at a real issue: `_parse_value()` returns `PropertyValue` (which includes `bool`, `float`, `datetime`), but `In.values` is typed as `list[int] | list[str]`.

What happens with `field IN (1.5, 2.5)` or `field IN (true, false)`? This would silently produce a list that doesn't match the declared type.

Either:
1. Widen `In.values` to `list[PropertyValue]`, or
2. Add explicit validation in the parser that rejects non-int/str values in IN lists with a clear error message.

src/memmachine/common/filter/filter_parser.py
```diff
@@ -154,43 +191,54 @@ def _parse_primary(self) -> FilterExpr:
...
             self._expect("RPAREN")
             return expr
-        return self._parse_comparison()
+        if self._accept("NOT"):
+            return Not(expr=self._parse_primary())
+        return self._parse_predicate()
 
-    def _parse_comparison(self) -> FilterExpr:
+    def _parse_predicate(self) -> FilterExpr:
         field_tok = self._expect("IDENT")
```
jealous, line 203: Minor: both `NOT field IN (...)` (via `_parse_primary` → `Not`) and `field NOT IN (...)` (here, special case) produce `Not(In(...))`. This is correct, but worth a brief comment noting the two entry points for NOT+IN, since the grammar handles them differently.

src/memmachine/common/filter/sql_filter_util.py
```diff
@@ -1,72 +1,115 @@
...
-def parse_sql_filter(
-    column: MappedColumn[Any] | InstrumentedAttribute[Any] | None,
-    is_metadata: bool,
-    expr: Comparison,
+def _compile_leaf(
+    expr: IsNull | In | Comparison,
+    resolve_field: FieldResolver,
 ) -> ColumnElement[bool] | None:
-    """Parse a FilterExpr comparison into an SQLAlchemy boolean expression."""
+    column, is_json = resolve_field(expr.field)
```
jealous, line 56:
```python
return column != column  # always false
```
This works but is surprising to read. Consider `sqlalchemy.false()` or `sqlalchemy.literal(False)` for clarity — both express "always false" more directly.

src/memmachine/episodic_memory/short_term_memory/short_term_memory.py
```diff
@@ -305,78 +312,65 @@ def _safe_compare(
...
-                )
-            logger.warning("Unsupported filter field: %s", filters.field)
+        if isinstance(expr, Comparison):
+            value = self._resolve_field(episode, expr.field)
+            return self._compare(value, expr.op, expr.value)
+        if isinstance(expr, In):
+            value = self._resolve_field(episode, expr.field)
+            return value in expr.values
+        if isinstance(expr, IsNull):
+            value = self._resolve_field(episode, expr.field)
```
jealous, line 326: This changes from reading `episode.metadata[key]` to `episode.filterable_metadata.get(key)`. The test update (`metadata=` → `filterable_metadata=`) confirms this is intentional.

But is this a breaking change for existing in-memory episodes that only have `metadata` set (not `filterable_metadata`)? `episodic_memory.py` auto-copies eligible values from `metadata` to `filterable_metadata` on add, but episodes already in the STM buffer before this change would only have `metadata` populated. Worth noting in the PR description if this is expected.

tests/memmachine/semantic_memory/storage/in_memory_semantic_storage.py
```diff
@@ -604,97 +615,60 @@ def _evaluate_filter_expr(
...
+        "<": lambda lhs, rhs: lhs < rhs,
+        ">=": lambda lhs, rhs: lhs >= rhs,
+        "<=": lambda lhs, rhs: lhs <= rhs,
+    }
+
     def _evaluate_comparison(
         self,
         entry: _FeatureEntry,
         comparison: FilterComparison,
     ) -> bool:
```
jealous, line 651: Good: removing `_normalize_metadata_value` (the string coercion) is correct since the bug was that everything was compared as strings. Worth noting in the PR description that this is an intentional behavior change for the in-memory test storage — metadata values are now compared with native types instead of string-coerced.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
```diff
@@ -896,16 +904,20 @@ async def search_directional_nodes(
...
-                            "=",
-                            f"$starting_at[{index}]",
-                            starting_at[index],
-                        )
-                        for index, sanitized_by_property in enumerate(
-                            sanitized_by_properties,
+                    + (
+                        " AND ".join(
+                            render_comparison(
+                                f"n.{sanitized_by_property}",
```
jealous, line 910: The `or "TRUE"` fallback handles the case where all `starting_at` values are `None` (so the generator yields nothing and `" AND ".join(...)` produces an empty string). This is a subtle edge case — a brief inline comment explaining why would help future readers.

jealous (commented):
Solid refactor overall — the new AST nodes (`In`, `IsNull`, `Not`), centralized `compile_sql_filter` + `FieldResolver` pattern, and `map_filter_fields` significantly reduce duplication and improve type safety. The JSON metadata numeric comparison fix (`_cast_json_column`) addresses a real bug, and test coverage is strong (+312 lines parser tests, +258 lines SQL filter integration tests). Approve with minor comments below.

src/memmachine/episodic_memory/short_term_memory/short_term_memory.py
edwinyyyu: It should be non-breaking since STM is not persisted.

src/memmachine/common/vector_graph_store/neo4j_vector_graph_store.py
edwinyyyu: Added comment.

src/memmachine/common/filter/sql_filter_util.py
edwinyyyu: now returns false

src/memmachine/common/filter/filter_parser.py
edwinyyyu: comments added

src/memmachine/common/filter/filter_parser.py
edwinyyyu: now checks types

PR #1128, opened 2026-02-20
fix: skip invalid episode_ids in semantic ingestion instead of crashing

### LINKED ISSUE #1102: [Bug]: Background semantic ingestion crashes/restarts when semantic storage references missing episode_ids
### Describe the bug

Reported error - https://github.com/MemMachine/MemMachine/actions/runs/21885425744/job/63179096417

The job is failing because the ingestion process encountered invalid episode_ids for set_id user-b, specifically ['4'], resulting in this error:

ValueError: Failed to retrieve messages. Invalid episode_ids exist for set_id user-b: ['4']
This error is raised in src/memmachine/semantic_memory/semantic_ingestion.py at line 110, when messages cannot be retrieved for the given episode_ids.

### Steps to reproduce

Review any Pull Request that failed the test.

### Expected behavior

Unit tests should pass.

### Environment

- OS: Mac OS X
- MemMachine: current commit

## Proposed fix

Investigate why episode_id '4' is invalid or missing from the episode storage. Check if it's being deleted prematurely or not added correctly.

If an episode was deleted or failed to save, ensure all references to that episode_id are removed or handled gracefully.

You can modify the ingestion logic to skip invalid episode_ids rather than fail entirely.
In src/memmachine/semantic_memory/semantic_ingestion.py, update the _process_single_set method as follows:

```python
# Replace
if len(none_h_ids) != 0:
    raise ValueError(
        "Failed to retrieve messages. Invalid episode_ids exist for set_id "
        f"{set_id}: {none_h_ids}"
    )
# With
if len(none_h_ids) != 0:
    logger.warning(f"Skipping invalid episode_ids for set_id {set_id}: {none_h_ids}")
    raw_messages = [m for m in raw_messages if m is not None]
    # Optionally, mark as ingested or delete references to prevent repeated failures
```

Additionally, audit your data insertion and deletion routines to prevent inconsistent episode references for set_id user-b.

This change will prevent complete ingestion failure and allow processing of valid messages.

haosenwang1018:
Closes #1102

When episode_ids referenced in semantic storage have been deleted or are otherwise invalid, the ingestion process crashes with a `ValueError`. This change gracefully skips invalid episode_ids with a warning log instead of failing entirely, allowing valid messages to still be processed.

As suggested in the issue discussion.

### DIFF
--- a/src/memmachine/semantic_memory/semantic_ingestion.py
+++ b/src/memmachine/semantic_memory/semantic_ingestion.py
@@ -114,6 +114,9 @@ class IngestionService:
             )
 
         raw_messages = [m for m in raw_messages if m is not None]
+        if not raw_messages:
+            logger.warning("No valid messages for set_id %s after filtering, skipping", set_id)
+            return
         messages = TypeAdapter(list[Episode]).validate_python(raw_messages)
 
         logger.info("Processing %d messages for set %s", len(messages), set_id)


haosenwang1018:
Closing per review/backlog policy.

PR #1141, opened 2026-02-25
remove container-names from compose

### LINKED ISSUE #1140: [Bug]: Explicit network and container name
### Describe the bug

Current compose uses explicit container names and network names which makes it impossible to start two stacks side by side.

### Steps to reproduce

Start one stack. Create a worktree to start a different stack.
```
$ git worktree list
~/src/github.com/ChristianKniep/MemMachine          f72ff3e [main]
~/src/github.com/ChristianKniep/MemMachine-cleanup  4481a17 [knowledge_graph]
```
The second stack changes the ports using the `.env` file.

It will not start b/c the container name is already used. Second issue is going to be that the network name is used as well.

### Expected behavior

Starting a stack without an explicit network defined will put all containers in a dedicated network anyways. 
Instead of using `docker exec` which kinda needs a stable name, one can use `docker-compose exec $service_name`.

### Environment

compose

### Additional context

_No response_

ChristianKniep:
### Purpose of the change

Removing `container-name` and explicit network from the docker-compose file and the startup script to not box the stack in.

### Description

I wanted to start two compose stacks from two branches on my laptop to show the difference of both. The `container-name` and explicit network name

### Fixes/Closes

Fixes #1140 

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

Change the `docker exec` use in the memmachine-compose script.

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### Screenshots/Gifs

### Further comments

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,6 @@ services:
   # PostgreSQL with pgvector extension for profile memory storage
   postgres:
     image: pgvector/pgvector:pg16
-    container_name: memmachine-postgres
     restart: unless-stopped
     ports:
       - "${POSTGRES_PORT:-5432}:5432"
@@ -25,7 +24,6 @@ services:
   # Neo4j database for graph storage
   neo4j:
     image: neo4j:5.23-community
-    container_name: memmachine-neo4j
     restart: unless-stopped
     ports:
       - "${NEO4J_HTTP_PORT:-7474}:7474"  # HTTP
@@ -60,7 +58,6 @@ services:
   memmachine:
     image: ${MEMMACHINE_IMAGE:-memmachine/memmachine}
     pull_policy: ${PULL_POLICY:-always}
-    container_name: memmachine-app
     restart: unless-stopped
     ports:
       - "${MEMORY_SERVER_PORT:-8080}:8080"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,9 +18,7 @@ services:
       timeout: 5s
       retries: 5
       start_period: 30s
-    networks:
-      - memmachine-network
-
+    
   # Neo4j database for graph storage
   neo4j:
     image: neo4j:5.23-community
@@ -51,8 +49,6 @@ services:
       timeout: 10s
       retries: 5
       start_period: 30s
-    networks:
-      - memmachine-network
 
   # MemMachine application
   memmachine:
@@ -99,8 +95,6 @@ services:
       timeout: 10s
       retries: 3
       start_period: 60s
-    networks:
-      - memmachine-network
     extra_hosts:
       - "host.docker.internal:host-gateway"
 
@@ -127,7 +121,3 @@ volumes:
   memmachine_logs:
     driver: local
 
-networks:
-  memmachine-network:
-    driver: bridge
-    name: memmachine-network
--- a/memmachine-compose.sh
+++ b/memmachine-compose.sh
@@ -101,7 +101,10 @@ find_docker_compose() {
         print_error "Docker Compose is not installed. Please install Docker Compose first."
         exit 1
     fi
-    
+    if [[ -z "$COMPOSE_PROJECT" ]]; then
+        COMPOSE_PROJECT="$(basename "$(pwd)")"
+    fi
+    COMPOSE_CMD="$COMPOSE_CMD -p $COMPOSE_PROJECT"
     print_success "Docker and Docker Compose are available"
 }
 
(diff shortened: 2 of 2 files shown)


ChristianKniep:
Two bridges are created with two stacks side-by-side:
```
$ docker network ls
NETWORK ID     NAME                         DRIVER    SCOPE
48e53306bfef   memmachine-cleanup_default   bridge    local
615a8a558eb7   memmachine-main_default      bridge    local
```

sscargal (commented):
This is a good proposal, though it's a less common use case.

This cleanly supports multiple deployments on the same host. Without hard-coded names, we can run: `memmachine-prod` and `memmachine-canary`, or `memmachine-staging` and `memmachine-prod` simultaneously using Compose project names (-p) and/or different directories.

By default, the project name is derived from the directory name you run Compose in (unless overridden). You can override it with:

- `docker compose -p <project> …`
- `COMPOSE_PROJECT_NAME=<project>`
- A `.env` value for `COMPOSE_PROJECT_NAME`.

**Default container & network names** 

Without a fixed `container_name:`, Compose names containers like:

- `<project>-postgres-1`
- `<project>-neo4j-1`
- `<project>-memmachine-1`

When the service is scaled, the suffix increments (…-2, …-3, etc.).

Because the explicit `networks:` block is removed, Compose will create a single default bridge network:

- `<project>_default`

All services are automatically attached to this network.

The compose file declares named volumes (e.g., postgres_data, neo4j_data, etc.). When they don’t have an explicit name: override, Compose will scope them to the project automatically:

- `<project>_postgres_data`
- `<project>_neo4j_data`
- `<project>_neo4j_logs`
- `<project>_memmachine_data`
- `<project>_memmachine_logs`

## Challenges

With this approach, we need to find a way to ensure the service ports don't cause conflicts, which they will in this PR.

Docker Compose only auto-assigns a host port if you omit the host port and publish container ports like this:

ports: ["5432"] (instead of "5432:5432")

Then Docker picks an ephemeral available host port (random-ish), which is:

- not predictable,
- not “instance 2 = base+1”,
- and harder for scripts/docs unless you query it (docker compose port <service> <containerPort>).

If we want predictable A/B stacks (8080 and 8081, 5432 and 5433, etc.), Compose won’t do that for us. If predictable increments are desired, they must be handled by memmachine-compose.sh (or by user-provided .env files).

1. User supplies ports per instance (simplest, most transparent)

    - stack A: defaults
    - stack B: POSTGRES_PORT=5433 MEMORY_SERVER_PORT=8081 ...

2. Script computes ports from an “instance number”
3. Use random ephemeral ports and print them. This is viable, but less friendly because every run can differ; you must query and display the assigned ports.

## TODO

Here are some ideas to further improve this PR

-  Add explicit Compose project naming support to `memmachine-compose.sh`

    e.g., support `--project <project_name>`, `-p <project_name>`, or honoring COMPOSE_PROJECT_NAME (less desirable) so developers/ops can run multiple stacks without relying on “directory name” defaults.

- Figure out a solution (default and user selectable) for assigning port numbers to the services.

ChristianKniep:
In my case I used different directories (using `git worktree`). I added an environment variable to define the `COMPOSE_PROJECT`. In case it's already up I error out:
```
$ ./memmachine-compose.sh 
[ERROR] Existing Docker Compose project detected with name: memmachinemain - please set a different project name (COMPOSE_PROJECT) or remove existing containers to avoid conflicts.
```
The ports can already be overwritten using there environment vars
- `POSTGRES_PORT`
- `NEO4J_HTTP_PORT`
- `NEO4J_HTTPS_PORT`
- `NEO4J_PORT`
- `MEMORY_SERVER_PORT`
- `DOCS_PORT`

Thus, when using a single stack nothing will change. 
Only difference is that you can not run `./memmachine-compose.sh` when the stack is already running (,which would have ran `docker-compose up -d` again).

sscargal:
`./memmachine-compose.sh` is the primary interface, so we'll need to find a solution that keeps existing behavior. The proposed `--project,-p` option is one idea. We should have a default `memmachine` project name to maintain the current behavior.

jgong:
The proposal is a good idea to allow multi-environments on one host, this is great! And thanks to the detailed review comments  @sscargal @ChristianKniep , i understand how it behaves better after the change is applied.

I think it is important to keep existing behavior, set default `memmachine` project sounds great to me!
And this also reminds me one feedback from pengfei earlier. He unzips the code downloaded from release into a new folder, the script `memmachine-compose.sh start` fails to load previous volumes because the project namespace has changed. i once asked him to set `COMPOSE_PROJECT_NAME=memmachine`

And I checked qa test code, specific container names are used in tests to check container statuses. Note that there will be effort expected to update test code to align with new naming of containers. Thanks.

PR #1154, opened 2026-03-01
fix(compose): remove explicit container and network names for parallel stacks

### LINKED ISSUE #1140: [Bug]: Explicit network and container name

haosenwang1018:
Fixes #1140

## Summary

Remove hardcoded `container_name` and network `name` directives from `docker-compose.yml` so multiple stacks can run side by side.

## Problem

The current compose file sets explicit names:
- `container_name: memmachine-postgres`
- `container_name: memmachine-neo4j`
- `container_name: memmachine-app`
- `name: memmachine-network`

This prevents running two stacks simultaneously (e.g. main branch + feature branch worktree), because Docker requires container and network names to be unique.

## Fix

Remove all `container_name` directives and the explicit network `name`. Docker Compose automatically scopes names using the project name (directory name or `COMPOSE_PROJECT_NAME`), so:
- Stack in `~/MemMachine/` → containers named `memmachine-postgres-1`, etc.
- Stack in `~/MemMachine-cleanup/` → containers named `memmachine-cleanup-postgres-1`, etc.

Users who need a specific prefix can set `COMPOSE_PROJECT_NAME` in `.env`.

## Note

`memmachine-compose.sh` also uses `docker exec memmachine-postgres` (lines 801, 810). A follow-up PR could migrate those to `docker compose exec postgres` for full compatibility with parallel stacks.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,6 @@ services:
   # PostgreSQL with pgvector extension for profile memory storage
   postgres:
     image: pgvector/pgvector:pg16
-    container_name: memmachine-postgres
     restart: unless-stopped
     ports:
       - "${POSTGRES_PORT:-5432}:5432"
@@ -25,7 +24,6 @@ services:
   # Neo4j database for graph storage
   neo4j:
     image: neo4j:5.23-community
-    container_name: memmachine-neo4j
     restart: unless-stopped
     ports:
       - "${NEO4J_HTTP_PORT:-7474}:7474"  # HTTP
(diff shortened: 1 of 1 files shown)


sscargal:
@haosenwang1018 This is a duplicate PR of #1141. 

I'm curious how prevalent this requirement is, as it complicates our existing QA/Test setup that runs outside GitHub. 

Q) Do you use or need parallel environments yourself?

PR #1173, opened 2026-03-03
Fix broken paths

edwinyyyu:
### Purpose of the change

Fix missed broken paths during refactor.

https://github.com/astral-sh/uv/issues/6685 for change in Dockerfile from locked to frozen

### DIFF
--- a/packages/server/pyproject.toml
+++ b/packages/server/pyproject.toml
@@ -69,6 +69,9 @@ Issues = "https://github.com/MemMachine/MemMachine/issues"
 [tool.setuptools.packages.find]
 where = ["src"]
 
+[tool.setuptools.package-data]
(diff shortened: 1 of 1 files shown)


edwinyyyu:
@jgong I think this should fix the Dockerfile?

jgong:
> @jgong I think this should fix the Dockerfile?

Thank you @edwinyyyu , I will check and update you

edwinyyyu:
pyproject.toml is missing
```
[tool.setuptools.package-data]
"*" = ["*.md", "*.txt", "*.yml", "*.yaml", "*.sql"]
```
from original.

jgong:
Hi @edwinyyyu , thanks for the fix, built image from latest change `1635a54`,  `memmachine-app` container can be started successfully. CC @sscargal

jgong (approved):
Thanks for fixing the Dockerfile!

PR #1199, opened 2026-03-06
New episodic memory (currently named EventMemory) based on VectorStore and (new) SegmentStore

edwinyyyu:
### Purpose of the change

Motivation:
- current DeclarativeMemory does not handle multimodal content
- current DeclarativeMemory does not handle large messages or text items well (no chunking).
- current DeclarativeMemory produces high fan out in number of Neo4j queries
- current DeclarativeMemory operations do not tolerate failures
- VectorGraphStore is difficult to implement using other databases
- we have a new VectorStore interface defined previously to move toward a solution
- tangential: current DeclarativeMemory may face problems with top-k redundancy from vector search
- current DeclarativeMemory has no way of distinguishing name from identity
  - top-level API from server does not provide enough information as there is no concept of temporal context
- add encryption support

### Description

Based on #1375.
#1205 based on this.

Changes:
- define new extensible data models to support multimodal content and chunking
- define new APIs to allow for more efficient and atomic operations
- API: limit is now vector limit for transparency and to avoid throwing away computations -- it is trivial for client to limit/threshold and transform into usable context and this allows much more flexible iterative expansion without requerying
- uses new API to avoid breaking existing memory for now -- although I would prefer an overhaul of the server due to its many problems #1297
- introduce derivative eviction system for duplicate-heavy online-ingested data (cannot pre-dedup, need to maintain index quality)

each segment comes from exactly one episode
each derivative comes from exactly one segment

alternative considered: many-to-many segments-to-derivatives
- slows growth of vector index
- handles HNSW index degradation due to ((potentially) near) duplicates
- rejected for complexity and inability to filter in vector store
- consolidation is about as effective as increasing the search limit proportionally
- requires reference counting, ownership transfer, or similar

Approach to deletion:
- purging state necessary as a lock to allow deleting from external DB (vector DB)

Scoping:
- by partition key: should allow sharding and horizontal scaling

- this memory will be parallel to `DeclarativeMemory`

Decisions to make:
- naming

POC:
- Current Neo4j implementation on a single MacBook Pro (M3 Pro, 18GB) cannot handle even 30 concurrent search queries with search limit 100 (latency explodes).
- Qdrant + PostgreSQL can handle 50+ concurrent search queries with acceptable latency.

### Type of change

Will eventually be breaking or require new API.

- [ ] Documentation update

### How Has This Been Tested?

APIs and data models not tested.

- [ ] Unit Test
- [ ] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code


malatewang (commented):
Let put vector store to a separate PR

edwinyyyu:
Need to wire in. Need to handle partition deletion.

edwinyyyu:
Updating properties processing to allow external system-defined fields.
Also need to define what field types are allowed for Context.

edwinyyyu:
Maybe we should remove filtering by context fields so that we can support encryption more easily and safely.

The application consuming the memory can then decide whether it wants to filter on context fields, and add equivalents to user metadata.

edwinyyyu:
https://github.com/MemMachine/MemMachine/commit/729f994056b883c716130d9823c8e8ad9808e3e9
is before rebase on key_manager

packages/server/src/memmachine_server/episodic_memory/event_memory/segment_store/segment_store.py
```diff
@@ -0,0 +1,228 @@
...
+        """Startup."""
+        raise NotImplementedError
+
+    @abstractmethod
+    async def shutdown(self) -> None:
+        """Shutdown."""
+        raise NotImplementedError
+
+    @abstractmethod
+    async def create_partition(
```
malatewang, line 144: What is the purpose of partition here?

packages/server/src/memmachine_server/episodic_memory/event_memory/data_types.py
```diff
@@ -0,0 +1,240 @@
...
+        self, v: dict[str, PropertyValue]
+    ) -> dict[str, dict[str, bool | int | float | str]]:
+        return encode_properties(v)
+
+    def __hash__(self) -> int:
+        """Hash an event by its UUID."""
+        return hash(self.uuid)
+
+
+class Segment(BaseModel):
```
malatewang, line 146: Is chunk a better name than segment?

packages/server/src/memmachine_server/episodic_memory/event_memory/data_types.py
```diff
@@ -0,0 +1,240 @@
...
+    Text,
+    Field(discriminator="type"),
+]
+
+
+class MessageContext(BaseModel):
+    """The content is communicated by a source."""
+
+    type: Literal["message"] = "message"
+    source: str
```
malatewang, line 47: The source is the actual data, id or something else?

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
```diff
@@ -0,0 +1,967 @@
...
+                    segments.extend(
+                        Segment(
+                            uuid=uuid4(),
+                            event_uuid=event.uuid,
+                            index=index,
+                            offset=offset,
+                            timestamp=event.timestamp,
+                            block=Text(text=chunk),
+                            context=context,
+                            properties=event.properties,
```
malatewang, line 383: Why only inherent properties, not metadata from events?

packages/server/src/memmachine_server/episodic_memory/event_memory/segment_store/segment_store.py
```diff
@@ -0,0 +1,228 @@
+"""
```
malatewang, line 1: There in no concrete implementation yet?

malatewang (changes requested):

packages/server/src/memmachine_server/episodic_memory/event_memory/segment_store/segment_store.py
edwinyyyu: It makes data isolation and sharding easier. Partitions can be deleted in O(1) time instead of requiring deleting all rows.

packages/server/src/memmachine_server/episodic_memory/event_memory/data_types.py
edwinyyyu: Chunk is more standard for text, but to me it feels weird for temporal types like audio or video. I am borrowing from terms like "image segmentation".

packages/server/src/memmachine_server/episodic_memory/event_memory/data_types.py
edwinyyyu: This is LLM-visible source, so it is a readable name. I want to separate this from the concept of producer id or similar that we currently have, which may be used in filtering.

The separation is important for privacy/security and LLM-friendliness (we want multiple tenants to be able to provide the same source/name).

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
edwinyyyu: Metadata from events would belong in context under this model.

It is not filterable because it is meant to be human/LLM-readable so it may be free-form, and because I have concerns about supporting encryption with it -- it can be private data (actual user name, or a sensitive topic they are reading a book on, etc.)

At a minimum, I think there needs to be the following separation:
- properties -- filterable, non-private, for non-readable ids and isolation and things like that
  - if a user wants to add filterable metadata, they are acknowledging the additional risk (need warning or something) -- especially if user-defined schemas are shared between tenants
- context -- things that are useful for retrieval (visible to LLM), may be private, structured so that the retrieval system doesn't have to work too hard for things that are typically easily available as metadata
- block/body -- the main content, unstructured

packages/server/src/memmachine_server/episodic_memory/event_memory/segment_store/segment_store.py
edwinyyyu: It's in #1205

edwinyyyu:
Review effort moved to #1205

PR #1201, opened 2026-03-09
Improve vector store API

edwinyyyu:
Motivation:
- Update APIs to be easier to read, work with, and extend.
- Support more efficient operations.
- Define concurrency models for horizontal scaling.

Changes:
- query now returns an Iterable with one item per query vector.
- Add collection startup and shutdown for cases where collection must be loaded.
- Support multitenancy via sharding and define concurrency guarantees.
- Fix default booleans to match docstring comments.
- Naming constraints.

Details:
- Most vector databases recommend against one collection per tenant, and empirical tests show that many collections induces significant overhead.
- We guarantee safe concurrent usage on a single process.
- We guarantee safe concurrent sharded usage (one node per logical collection) on multiple processes.

Naming constraints determined by lowest common denominators for support.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -63,7 +63,6 @@
               "install_guide/integrate/crewai",
               "install_guide/integrate/n8n",
               "install_guide/integrate/dify",
-              "install_guide/integrate/Google_ADK",
               "install_guide/integrate/LlamaIndex",
               "install_guide/integrate/nebula_graph_integration_guide"
             ]
--- a/docs/getting_started/quickstart.mdx
+++ b/docs/getting_started/quickstart.mdx
@@ -100,10 +100,10 @@ The `memmachine-compose.sh` script is your main tool for managing the server.
 ```
 
 Common commands:
-- `./memmachine-compose.sh start - Start MemMachine services
-- `./memmachine-compose.sh stop - Stop MemMachine services
-- `./memmachine-compose.sh restart - Restart MemMachine services
-- `./memmachine-compose.sh logs - Show service logs
-- `./memmachine-compose.sh clean - Remove all services and data
-- `./memmachine-compose.sh build - Build a custom MemMachine image
-- `./memmachine-compose.sh help - Show help message
+- `./memmachine-compose.sh start` - Start MemMachine services
+- `./memmachine-compose.sh stop` - Stop MemMachine services
+- `./memmachine-compose.sh restart` - Restart MemMachine services
+- `./memmachine-compose.sh logs` - Show service logs
+- `./memmachine-compose.sh clean` - Remove all services and data
+- `./memmachine-compose.sh build` - Build a custom MemMachine image
+- `./memmachine-compose.sh help` - Show help message


edwinyyyu:
Blocks #1199, #1205

packages/server/src/memmachine_server/common/vector_store/vector_store.py
```diff
@@ -19,16 +19,29 @@
...
+    @abstractmethod
+    async def shutdown(self) -> None:
+        """Shutdown."""
+        raise NotImplementedError
+
     @abstractmethod
     async def upsert(
         self,
         *,
+        partition_key: str,
```
malatewang: Why do we need partition here?

packages/server/src/memmachine_server/common/vector_store/vector_store.py
```diff
@@ -140,54 +188,121 @@ async def shutdown(self) -> None:
...
         *,
+        namespace: str,
+        name: str,
         vector_dimensions: int,
         similarity_metric: SimilarityMetric = SimilarityMetric.COSINE,
         properties_schema: Mapping[str, type[PropertyValue]] | None = None,
     ) -> None:
         """
-        Create a collection in the vector store.
+        Create a logical collection in the vector store and return a handle to it.
```
malatewang, line 171: The return type should be Collection

packages/server/src/memmachine_server/common/vector_store/vector_store.py
```diff
@@ -140,54 +188,121 @@ async def shutdown(self) -> None:
...
 
+        Raises:
+            CollectionConfigurationMismatchError: If a collection with the same
+                (namespace, name) already exists with a different configuration.
         """
         raise NotImplementedError
 
     @abstractmethod
-    async def delete_collection(self, collection_name: str) -> None:
+    async def open_collection(self, *, namespace: str, name: str) -> Collection:
```
malatewang: return type should be Collection | None

malatewang (changes requested):

packages/server/src/memmachine_server/common/vector_store/vector_store.py
edwinyyyu: In open_or_create_collection.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
edwinyyyu: Fixed.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
edwinyyyu: Fixed.

edwinyyyu:
Before merge, need to check that (de)serialization is deterministic.

edwinyyyu:
> Before merge, need to check that (de)serialization is deterministic.

Tests work now on Qdrant implementation branch.

PR #1205, opened 2026-03-10
New long-term episodic memory (named EventMemory to differentiate) based on VectorStore and (new) SegmentStore

edwinyyyu:
### Purpose of the change

Motivation:
- current DeclarativeMemory does not handle multimodal content
- current DeclarativeMemory does not handle large messages or text items well (no chunking).
- current DeclarativeMemory produces high fan out in number of Neo4j queries
- current DeclarativeMemory operations do not tolerate failures
- VectorGraphStore is difficult to implement using other databases
- we have a new VectorStore interface defined previously to move toward a solution
- tangential: current DeclarativeMemory may face problems with top-k redundancy from vector search
- current DeclarativeMemory has no way of distinguishing name from identity
  - top-level API from server does not provide enough information as there is no concept of temporal context
- add encryption support (initial infrastructure)

### Description

For wiring into server, #1304 is required.

Choices:
- JSON/JSONB allows faster upserts. Filters are already scoped to queries so queries are fast. GIN indexes are not so useful right now. Another alternative is EAV properties table which requires more complexity but can be better for range filters.

Changes:
- define new extensible data models to support multimodal content and chunking
- define new APIs to allow for more efficient and atomic operations
- API: limit is now vector limit for transparency and to avoid throwing away computations -- it is trivial for client to limit/threshold and transform into usable context and this allows much more flexible iterative expansion without requerying
- uses new API to avoid breaking existing memory for now -- although I would prefer an overhaul of the server due to its many problems #1297
~- introduce derivative eviction system for duplicate-heavy online-ingested data (cannot pre-dedup, need to maintain index quality)~

each segment comes from exactly one episode
each derivative comes from exactly one segment

alternative considered: many-to-many segments-to-derivatives
- slows growth of vector index
- handles HNSW index degradation due to ((potentially) near) duplicates
- rejected for complexity and inability to filter in vector store
- consolidation is about as effective as increasing the search limit proportionally
- requires reference counting, ownership transfer, or similar

Approach to deletion:
- purging state necessary as a lock to allow deleting from external DB (vector DB)

Scoping:
- by partition key: should allow sharding and horizontal scaling

- this memory will be parallel to `DeclarativeMemory`

Decisions to make:
- naming

POC:
- Current Neo4j implementation on a single MacBook Pro (M3 Pro, 18GB) cannot handle even 30 concurrent search queries with search limit 100 (latency explodes).
- Qdrant + PostgreSQL can handle 50+ concurrent search queries with acceptable latency.

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code


edwinyyyu:
Still needs some performance fixes.

edwinyyyu:
Maybe lazy ownership transfer is more performant.

edwinyyyu:
Needs wiring still.

edwinyyyu:
More work is required for encryption.

edwinyyyu:
https://github.com/edwinyyyu/MemMachine/tree/sqlalchemy_segment_store for version with encryption support.

packages/server/src/memmachine_server/episodic_memory/event_memory/data_types.py
```diff
@@ -0,0 +1,240 @@
...
+Block = Annotated[
+    Text,
+    Field(discriminator="type"),
+]
+
+
+class MessageContext(BaseModel):
+    """The content is communicated by a source."""
+
+    type: Literal["message"] = "message"
```
malatewang: It is clearer to use messagecontext

packages/server/src/memmachine_server/episodic_memory/event_memory/data_types.py
```diff
@@ -0,0 +1,240 @@
...
+
+class Segment(BaseModel):
+    """Snapshot of an event, representing a smaller unit of content."""
+
+    uuid: UUID
+    event_uuid: UUID
+    index: int
+    offset: int
+    timestamp: datetime
+    context: Context = Field(default_factory=NullContext)
```
malatewang, line 134: Why the context only in segment? I think it should belong to event

packages/server/src/memmachine_server/episodic_memory/event_memory/data_types.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+class Derivative(BaseModel):
+    """Information derived from a segment."""
+
+    uuid: UUID
+    segment_uuid: UUID
+    timestamp: datetime
+    context: Context = Field(default_factory=NullContext)
+    text: str
```
malatewang: It is better to use block instead of plain str here

edwinyyyu:
Wiring for SQLAlchemySegmentStore will be added in #1304 if this is merged.

packages/server/src/memmachine_server/episodic_memory/event_memory/segment_store/sqlalchemy_segment_store.py
```diff
@@ -0,0 +1,1024 @@
...
+        self._payload_codec = payload_codec
+        self._create_session = async_sessionmaker(engine, expire_on_commit=False)
+        self._is_sqlite = engine.dialect.name == "sqlite"
+
+    @override
+    @property
+    def config(self) -> SegmentStorePartitionConfig:
+        return self._config
+
+    async def _lock_partition_for_write(self, session: AsyncSession) -> None:
```
malatewang, line 223: It should be safe to assume that one partition managed by one instance. It may be enough to use in memory lock.

packages/server/src/memmachine_server/episodic_memory/event_memory/segment_store/sqlalchemy_segment_store.py
```diff
@@ -0,0 +1,1024 @@
...
+                SQLAlchemySegmentStorePartition._resolve_segment_field,
+            )
+            if property_filter is not None
+            else None
+        )
+
+        for seed_uuid, seed_row in seed_rows_by_uuid.items():
+            seed_ordering_values = tuple_(
+                literal(seed_row.timestamp),
+                literal(seed_row.event_uuid),
```
malatewang, line 542: I think the uuid is just randome ID?

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
```diff
@@ -0,0 +1,967 @@
...
+        ...,
+        description="Embedder instance for creating embeddings",
+    )
+    reranker: InstanceOf[Reranker] | None = Field(
+        None,
+        description="Reranker instance for scoring search results. "
+        "If None, embedding similarity scores are used instead",
+    )
+    derive_sentences: bool = Field(
+        False,
```
malatewang: It is better to pass an instance of text chunker and derivative generator.

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
```diff
@@ -0,0 +1,967 @@
...
+            for field in event.properties
+            if EventMemory._is_reserved_field(field)
+        }
+        if reserved_fields:
+            raise ValueError(
+                f"Event properties must not contain reserved fields: "
+                f"{', '.join(sorted(reserved_fields))}"
+            )
+
+        missing_fields = (
```
malatewang: This has been checked in the constructor

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
```diff
@@ -0,0 +1,967 @@
...
+                or if the collection schema is missing fields required by any event's Context type.
+        """
+        t_start = time.monotonic()
+
+        events = list(events)
+        self._validate_events(events)
+
+        events = sorted(
+            events,
+            key=lambda event: (event.timestamp, event.uuid),
```
malatewang: Since the UUID is random, there is no sense to sort by uuid here. Or do we require special UUID generator?

packages/server/src/memmachine_server/common/vector_store/vector_store.py
```diff
@@ -25,6 +25,12 @@ class VectorStoreCollection(ABC):
 
     Identified by a (namespace, name) pair.
     All data operations are scoped to this logical collection.
+
+    Implementations must support storing, filtering on, and returning
+    record properties not declared in the configured indexed properties schema.
+
+    The schema exists to support indexing on fixed-type record properties.
+    Record properties not declared in the schema may have mixed-type values.
```
malatewang, line 33: Looks like one API is missing: build index

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
```diff
@@ -0,0 +1,967 @@
...
+        500,
+        description="Max code-point length for text chunking in segment creation",
+    )
+
+
+class EventMemory:
+    """Event memory system."""
+
+    # System-defined metadata field names. Reserved.
+    _SEGMENT_UUID_FIELD_NAME = "_segment_uuid"
```
malatewang, line 114: Are they enough? Event source should be indexed

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
```diff
@@ -0,0 +1,967 @@
...
+        "If None, embedding similarity scores are used instead",
+    )
+    derive_sentences: bool = Field(
+        False,
+        description="Whether to derive sentence-level derivatives from content",
+    )
+    max_text_chunk_length: int = Field(
+        500,
+        description="Max code-point length for text chunking in segment creation",
+    )
```
malatewang, line 107: Should add metrics to the parameter

malatewang (changes requested):

packages/server/src/memmachine_server/episodic_memory/event_memory/data_types.py
edwinyyyu: The original intent was for Derivatives to be text-only since multimodal embedding models are uncommon. The solution for supporting multimodal embeddings is probably to make Derivatives text if the embedding reports no support for multimodality.

packages/server/src/memmachine_server/episodic_memory/event_memory/data_types.py
edwinyyyu: Updated to be ProducerContext and context_type: "producer". Removed CitationContext (no use case yet).

Having "context" in the value for the discriminator is redundant and more error-prone than "context" in the discriminator key.

packages/server/src/memmachine_server/episodic_memory/event_memory/data_types.py
edwinyyyu: Removed body type union. Now Event gets context and blocks directly instead of a union of different body types.

packages/server/src/memmachine_server/episodic_memory/event_memory/segment_store/sqlalchemy_segment_store.py
edwinyyyu: The current implementation is safe already and performance is fine so far due to the lock being low granularity and shared. It also allows for safer multi-process usage in a case that does not use the REST client/server.

packages/server/src/memmachine_server/episodic_memory/event_memory/segment_store/sqlalchemy_segment_store.py
edwinyyyu: It's there so that segments from the same event are grouped together instead of randomly interleaved, and for consistent ordering (one use case is retrieving temporally close context dynamically without extra vector searches).

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
edwinyyyu: This checks that input events for ingestion are not missing fields, which cannot be done at construction time.

Unless you mean it is checked at Event construction time? In that case I think it's safer to check close to where the condition is required.

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
edwinyyyu: The implied behavior is that UUID can be used to order events that have the same timestamp, so behavior with UUIDv7 vs. UUIDv4 works differently.

This is not documented, so I can remove it.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
edwinyyyu: The index creation is automatic. I don't know of a consumer that wants to manually build the index. It can be added when there is one.

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
edwinyyyu: Event source (along with all Context fields) is LLM-friendly private data that should not be indexed in a multi-tenant system where we may have encryption. If the user/consumer feels that they need to filter on this and it is not sensitive information, they may add it to the properties schema.

It is safer to completely prohibit indexing anything in Context because Context can reveal things like the user's name, interests (if adding content from something like a book), sensitive information like location and time they were there (which may reveal where they live), etc in plaintext (which may be required for most/all filters to work).

If it needs to be system-defined, the application may add _producer_id as a non-LLM-friendly system-reserved field in properties (EventMemory is not responsible for reserved prefix validation, and it allows for creating upper-level system-reserved fields.). In this proposed case, producer_id is a property that exposes no sensitive information (it is just a user or agent id like user_123 or a UUID or something).

Another challenge that we avoid by not indexing this is that semantically different fields (potentially with different types) with the same name do not collide. Because Context is a discriminated union, enforcing this would be complicated and each Context implementation would have to consider what other implementations have already reserved.

Originally I agreed with this idea. But the above considerations made me reconsider. See https://github.com/MemMachine/MemMachine/pull/1205/commits/41f2fc1fabab523fd21718c5acd7b7a73ef29e2f, https://github.com/MemMachine/MemMachine/pull/1205/commits/7582233c2b33b28fae18df8af222dd1075d3dee4 for changes. Also the code was a lot more complicated when supporting filtering by context.

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
edwinyyyu: I think the original motivation for sorting at all was to make temporal derivative eviction easier -- simulating the order of occurrence.

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
edwinyyyu: Done.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
malatewang: The vector store can filter search based on metadata. Qdrant requires to build payload index manually.

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
malatewang: I do not think the code is working that way. It does not check the property of the incoming events.

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
```diff
@@ -0,0 +1,926 @@
...
+            "embed=%.3fs "
+            "seg_store=%.3fs "
+            "v_store=%.3fs "
+            "total=%.3fs",
+            t_segment - t_start,
+            t_derive - t_segment,
+            t_embed - t_derive,
+            t_seg_store - t_embed,
+            t_v_store - t_seg_store,
+            t_v_store - t_start,
```
malatewang: Add the stats to metrics

malatewang (changes requested):

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
edwinyyyu: Sorry, I think that (or checking that the schema includes all fields required by events) was the behavior at some point but it changed and it was not deduplicated. I will remove.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
edwinyyyu: I mean that if a vector store backend supports payload indexing, the indexes should be created at native collection creation time (automatically from the perspective of the consuming application).

Consider QdrantVectorStore, which creates indexes in a loop:
```
for prop_name, prop_type in config.indexed_properties_schema.items():
    index_type = QdrantVectorStore._PROPERTY_TYPE_TO_INDEX_TYPE.get(
        prop_type
    )
    if index_type is not None:
        await self._client.create_payload_index(
            collection_name=native_collection_name,
            field_name=prop_name,
            field_schema=index_type,
        )
```

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
edwinyyyu: Done. Renamed/reorganized.

packages/server/src/memmachine_server/episodic_memory/event_memory/event_memory.py
```diff
@@ -0,0 +1,941 @@
...
+                "Time spent in each phase of encode_events",
+                label_names=("phase",),
+            )
+            self._query_phase_seconds = params.metrics_factory.get_histogram(
+                "event_memory_query_phase_seconds",
+                "Time spent in each phase of query",
+                label_names=("phase",),
+            )
+
+        self._text_splitter = RecursiveCharacterTextSplitter(
```
malatewang: In our discussion, I think we agreed to move this out of the event memory

edwinyyyu: Done.

malatewang (approved):

edwinyyyu:
Before merge, I would like to do another manual verification run on LoCoMo/LongMemEval, since there have been quite a few changes since last runs.

edwinyyyu:
LoCoMo score is good enough. Used SQLite for both segment store and vector store (sqlite-vec). Happy path works.

edwinyyyu:
No breaking changes since 0.3.7 so I'm changing the milestone to 0.3.8. Wiring determines if 0.4.0 is needed.

PR #1258, opened 2026-03-23
feat: implement SQLitePersistence layer for local memory storage

jhawpetoss6-collab:
This PR introduces a `SQLitePersistence` layer to MemMachine, providing a robust local storage alternative to in-memory or cloud-based solutions.

### Changes:
- Added `SQLitePersistence` in `memmachine/storage/sqlite/`.
- Support for structured memory storage with metadata.
- Enables persistent agent memory across restarts without external dependencies.

/claim #persistence

### DIFF
--- a/memmachine/storage/sqlite/persistence.py
+++ b/memmachine/storage/sqlite/persistence.py
@@ -1,0 +1,37 @@
+import sqlite3
+import json
+
+class SQLitePersistence:
+    """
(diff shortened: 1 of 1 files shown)


edwinyyyu:
I don't see how this is meant to be integrated into the system as a whole? Closing since stale.
There are #1292 and #1205 that will allow using SQLite.

PR #1259, opened 2026-03-24
fix: stop semantic ingestion retry loop on context_length_exceeded

### LINKED ISSUE #1252: Semantic ingestion retries forever on context_length_exceeded errors
## Summary

When a message exceeds the LLM's context window, the background ingestion task logs the error but retries immediately in a tight loop — causing high CPU and unbounded database cost on Aurora Serverless v2.

## Impact

On our prod environment, 177 uningested messages that are too large for `gpt-4o-mini` cause a tight retry loop that drives Aurora ACU to 12-16 (the configured max) and writer CPU to 100%.

## Root Cause

In `semantic_memory.py:772-787`, the background ingestion loop:

```python
while not self._is_shutting_down:
    dirty_sets = await self._semantic_storage.get_history_set_ids(...)
    if len(dirty_sets) == 0:
        await asyncio.sleep(self._background_ingestion_interval_sec)
        continue
    try:
        await ingestion_service.process_set_ids(dirty_sets)
    except Exception:
        logger.exception("background task crashed, restarting")
```

When `process_set_ids` fails (e.g., OpenAI returns `context_length_exceeded`), the exception is caught, logged, and the loop restarts **immediately with no backoff**. The same messages are found again, the same error occurs, and the cycle repeats as fast as the database can serve the polling query.

The OpenAI error from prod logs:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': 'Your input exceeds the context window 
of this model. Please adjust your input and try again.', 'type': 'invalid_request_error', 
'param': 'input', 'code': 'context_length_exceeded'}}
```

## Suggested Fixes

1. **Add exponential backoff on failure** — sleep after exceptions, increasing the delay on consecutive failures (e.g., 2s → 4s → 8s → ... capped at 60s).

2. **Skip permanently failing messages** — after N retries for the same set_id, mark its messages as ingested (or add a `failed` state) so they don't block the loop forever.

3. **Handle `context_length_exceeded` specifically** — truncate or chunk oversized inputs before sending to the LLM, or skip them with a warning rather than raising an exception.

4. **Always sleep between iterations** — even when `dirty_sets` is not empty, sleep briefly between processing rounds to prevent CPU saturation on repeated failures.

## Related

- MemMachine/MemMachine#1251 — the polling query itself is also O(n²), compounding this issue

o-love:
### Purpose of the change

Prevent semantic ingestion from retrying indefinitely on messages that can never be processed because they exceed the model context window.

### Description

This change treats `context_length_exceeded` failures as non-retryable during semantic ingestion. When that error is detected (including wrapped exception chains), the message is skipped and marked ingested so it does not remain in the dirty set forever.

It also adds a regression test that reproduces a wrapped external API error with underlying `context_length_exceeded` and verifies the message is marked ingested.

### Fixes/Closes

Fixes #1252

### Type of change

### How Has This Been Tested?

- [x] Unit Test

**Test Results:**

- `uv run ruff check packages/server/src/memmachine_server/semantic_memory/semantic_ingestion.py packages/server/server_tests/memmachine_server/semantic_memory/test_semantic_ingestion.py`
- `uv run pytest packages/server/server_tests/memmachine_server/semantic_memory/test_semantic_ingestion.py packages/server/server_tests/memmachine_server/semantic_memory/test_semantic_memory_background.py -q`

### Checklist

### Screenshots/Gifs

N/A

### Further comments

None

### DIFF
--- a/packages/server/src/memmachine_server/semantic_memory/semantic_ingestion.py
+++ b/packages/server/src/memmachine_server/semantic_memory/semantic_ingestion.py
@@ -30,6 +30,27 @@ from memmachine_server.semantic_memory.storage.storage_base import SemanticStora
 logger = logging.getLogger(__name__)
 
 
+def _is_context_length_exceeded_error(error: Exception) -> bool:
+    seen: set[int] = set()
+    current: BaseException | None = error
+    while current is not None and id(current) not in seen:
+        seen.add(id(current))
+        code = getattr(current, "code", None)
+        if isinstance(code, str) and code.lower() == "context_length_exceeded":
+            return True
+
+        message = str(current).lower()
+        if (
+            "context_length_exceeded" in message
+            or "exceeds the context window" in message
+        ):
+            return True
+
+        current = current.__cause__ or current.__context__
+
+    return False
+
+
 class IngestionService:
     """
     Processes un-ingested history for each set_id and updates semantic features.
@@ -177,7 +198,17 @@ class IngestionService:
                         model=resources.language_model,
                         update_prompt=semantic_category.prompt.update_prompt,
                     )
-                except Exception:
+                except Exception as err:
+                    if _is_context_length_exceeded_error(err):
+                        logger.warning(
+                            "Skipping message %s for semantic type %s due to non-retryable context length error",
+                            message.uid,
+                            semantic_category.name,
+                        )
+                        if message.uid not in mark_messages:
+                            mark_messages.append(message.uid)
+                        continue
+
                     logger.exception(
                         "Failed to process message %s for semantic type %s",
                         message.uid,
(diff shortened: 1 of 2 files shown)


packages/server/src/memmachine_server/semantic_memory/semantic_ingestion.py
```diff
@@ -30,6 +30,27 @@
...
+            return True
+
+        message = str(current).lower()
+        if (
+            "context_length_exceeded" in message
+            or "exceeds the context window" in message
+        ):
+            return True
+
+        current = current.__cause__ or current.__context__
```
edwinyyyu, line 49: Can both of these be truthy simultaneously?

edwinyyyu (approved):

PR #1291, opened 2026-04-01
 Add SQL filter compilation for properties JSON

### LINKED ISSUE #1380: [Feat]: SQL filter compilation improvement

Current filter compilation logic does not apply filters intuitively and is unsafe.

Claude review 1:
```
  compile_sql_filter then has explicit None-aware combinators:

  - And(left, right): if either side is None, returns the other side. Both None → None. → unknown leaf behaves like TRUE (no
  constraint).
  - Or(left, right): same code path — if either side is None, returns the other side. → unknown leaf behaves like FALSE (sibling
  alone carries the disjunction).
  - Not(inner): if inner is None, returns None. → NOT unknown again falls out as "no constraint".

  If the entire expression collapses to None, compile_sql_filter returns None and the caller (e.g. _apply_episode_filter at
  episode_sqlalchemy_store.py:271-274) treats that as "no filter clause" and the WHERE is omitted.

  Net effect / sharp edges

  - An unknown leaf inside an AND widens the result (drops the constraint).
  - An unknown leaf inside an OR narrows the result — A OR unknown becomes just A, not TRUE. So unknown is implicitly "true under
  AND, false under OR" — semantically inconsistent but matches the intuition that "the field doesn't exist, so no row satisfies it."
  - NOT unknown evaluates to "no constraint" too (because ~None would crash, the code short-circuits to None), which contradicts the
  OR reading.
  - The only feedback to the caller is a logger warning; the API does not surface a typed error for misspelled fields.
```

Claude review 2:
```
❯ does the logic make sense under a filtering system? Is it consistent with intuition?

⏺ Short answer: no — it's neither algebraically consistent nor a defensible filter semantics.

  Why it isn't algebraically consistent

  There's no single truth assignment for "unknown" that reproduces all the cases. Trace it:

  ┌────────────────────┬─────────────────┬────────────────┬──────────────────────────┐
  │        Tree        │    Compiled     │ Effective rows │ Implied truth of unknown │
  ├────────────────────┼─────────────────┼────────────────┼──────────────────────────┤
  │ unknown            │ None → no WHERE │ all rows       │ TRUE                     │
  ├────────────────────┼─────────────────┼────────────────┼──────────────────────────┤
  │ A AND unknown      │ A               │ rows where A   │ TRUE (A ∧ T = A)         │
  ├────────────────────┼─────────────────┼────────────────┼──────────────────────────┤
  │ A OR unknown       │ A               │ rows where A   │ FALSE (A ∨ F = A)        │
  ├────────────────────┼─────────────────┼────────────────┼──────────────────────────┤
  │ NOT unknown        │ None → no WHERE │ all rows       │ FALSE (¬F = T)           │
  ├────────────────────┼─────────────────┼────────────────┼──────────────────────────┤
  │ A AND NOT unknown  │ A               │ rows where A   │ FALSE                    │
  ├────────────────────┼─────────────────┼────────────────┼──────────────────────────┤
  │ unknown OR unknown │ None → no WHERE │ all rows       │ T or F (either)          │
  └────────────────────┴─────────────────┴────────────────┴──────────────────────────┘

  It flips between TRUE-like and FALSE-like depending on whether a sibling survived. It also isn't 3-valued (NULL) logic: in SQL,
  WHERE UNKNOWN excludes the row, so A AND UNKNOWN returns nothing — the opposite of what this code does.

  Where it violates intuition

  - A user expects misspellings to be loudly wrong (error) or uniformly conservative (match no rows). This code instead silently
  shifts the meaning of the whole expression depending on tree shape.
  - The bare-unknown case matches all rows. Intuitively, if I write WHERE bogus = 'x', I expect zero rows or an error, not "ignore
  the predicate."
  - NOT unknown returning "no constraint" is especially surprising: I asked to exclude something and got no exclusion.

  Concrete failure modes

  These aren't hypothetical — they're sharp:

  1. Auth/tenant scoping silently opens up. If a caller filters tenant_id = X AND role = admin and typos tennant_id, the predicate
  collapses to role = admin across all tenants. If both are unknown, every row matches. Any filter used as an access-control gate is
  unsafe under this policy.
  2. Negation inverts intent. NOT internal_only with internal_only misspelled returns all rows including internal ones.
  3. OR narrows instead of widens. flagged OR archived with archived misspelled returns only flagged rows — the user's "either"
  became "only."
  4. Schema drift hides. Renaming a column makes existing filters silently match different row sets. There's only a logger warning at
   WARN level — easy to miss in production.

  What "consistent" filter semantics look like

  Pick one and stick to it:

  - Strict (the segment-store branch): unknown → ValueError. Fails fast, surfaces typos at the API boundary. Best default for a typed
   filter language.
  - Safe-deny: unknown leaf compiles to false(), and NOT unknown to true(). Then unknown matches nothing, A AND unknown matches
  nothing, A OR unknown matches A. Defensible and predictable. Used by some search DSLs that need to tolerate evolving schemas
  without breaking clients.
  - 3-valued (SQL NULL): harder to implement correctly but matches DB intuition; rarely worth it for a filter compiler.

  The current "drop the leaf and re-knit the tree" rule is none of these. It looks like it was written to make the tree compile under
   partial schema knowledge, but it does so by changing the semantics of the surviving expression in a non-uniform way.

  Recommendation

  For a filtering system that's also used for authorization-shaped queries (tenant scoping, role gates, soft-delete filters), the
  strict policy is the only one I'd defend. If permissive/forward-compatible behavior is genuinely needed for a subset of fields,
  gate that explicitly — e.g. resolver returns ("ignore", False) and the compiler maps that to true() — so the call site opts in and
  the algebra stays consistent.
```

https://github.com/MemMachine/MemMachine/pull/1291
# Filter Compilation: Design

Backend-agnostic description of the filter-compilation contract: what was wrong with the previous design and what the new design guarantees. Talks about `FilterExpr → predicate` compilation in the abstract; storage-layer details live in `SPEC.md`.

## Problems with the previous design

1. **Unknown fields had no consistent boolean meaning.** The resolver could signal "I don't recognize this name" by returning a sentinel, and the compiler would drop the leaf and re-knit the surrounding boolean tree. The result was that the implicit truth value of an unrecognized leaf depended on its parent — TRUE-shaped under `AND`, FALSE-shaped under `OR`, "no constraint" at the top level or under `NOT`. No single truth assignment to "unknown" reproduced all the cases, and the behavior matched neither classical two-valued logic nor SQL's three-valued logic.

2. **Typos and schema drift were invisible.** Unrecognized fields produced no observable error at the API boundary. Misspelled names and renamed columns compiled successfully and silently changed the meaning of the surrounding query.

3. **Storage encoding was a binary flag.** Resolvers returned `(column, is_json)`, which conflated "is this a JSON path" with "every encoding the system might want." Adding a third encoding required either changing every resolver's signature or smuggling extra information through the column expression.

4. **Boolean combinators carried implicit conventions.** Sentinel-handling for "leaf disappeared" was pushed into `And`, `Or`, and `Not`, so the meaning of a subexpression depended on what was around it. The algebra was hard to reason about and the implementation hard to extend.

## New behaviors

1. **Unknown fields fail loud.** The resolver returns either a resolved field (column expression plus encoding tag) or raises a typed error naming the offending field. The compiler does not catch the error. Misspellings, removed columns, and schema drift all surface before the query runs. Programmer errors are routed through the exception channel; data conditions are not.

2. **Boolean combinators behave as classical two-valued operators.** `And`, `Or`, and `Not` recurse over their children with no sentinel propagation, leaf-dropping, or tree re-knitting. The truth value of any subexpression depends only on its own structure and its operands' values, never on what's around it.

3. **Encoding is a closed tag, not a boolean.** The resolver returns one of a small enumeration of encoding tags alongside the column. Adding a new encoding is an additive change: define a new tag value and a new leaf compiler; existing call sites do not change.

4. **Leaves are type-aware.** Each encoding's leaf compiler picks the right cast for the comparison value's type. Encodings that store typed values additionally check that the stored type matches the comparison-value type before comparing, so a comparison whose value type doesn't match the storage type excludes the row rather than spuriously matching or erroring.

5. **One short-circuit, ordered after resolution.** An empty `IN` set compiles to "match no rows" — but only after the field has resolved successfully. An empty `IN` against an unknown field still raises. This is the only deviation from "every leaf goes through the resolver and emits a comparison," and it exists only because most backends treat empty value lists as a syntax error.

## Field naming and namespaces

Field names outside the resolver's recognized set raise. To support sparse user-defined keys without losing the strict-on-typo guarantee, resolvers may carve out a namespace prefix that always resolves (canonically `m.` or `metadata.`). Keys under that prefix compile successfully even when no record has ever stored them; missing-key behavior then falls through to ordinary "comparison against absent value excludes the row" semantics at query time.

This split is the central design choice: the closed schema is treated as a contract (typos error), and the open keyspace is treated as data (missing keys exclude). The compiler does not need to know which fields are which — it only sees encoding tags — but the contract makes both halves predictable for callers.

## In short

The previous design tried to be helpful by tolerating unknown fields silently. The new design refuses to. Anything that looks like a programmer error — typo, schema drift, type mismatch on a typed field — raises at compile time. Anything that looks like a data condition — missing key, absent value — follows ordinary backend semantics at query time.

Three-valued logic (rejected):
```
3VL is the right tool for partial data (a row whose salary is NULL), not partial queries (a caller who typed salray instead of
  salary) — the first is a real-world condition the database should reason about, the second is a bug that should fail loudly. Under
  3VL-at-compile-time, WHERE salray = 100000 silently returns zero rows, identical to "nobody earns that"; under strict mode it
  raises and you fix the typo before it ships. Same shape for schema drift: rename a column from created_at to created_ts, and every
  old WHERE created_at > ... quietly returns empty under 3VL but errors loudly under strict — the latter is the conversation you
  actually want.
```
A 3VL-shaped reading of "unknown field name" — `A AND unknown = U`, etc. — is rejected because schema errors are programmer errors, not data conditions. Surfacing them as `ValueError` is more useful than smearing them into row-inclusion behavior.

### Additional context

_No response_

edwinyyyu:
### Purpose of the change

Needed for SQLAlchemySegmentStore, SQLite VectorStore implementations.

### Description

Made based on #1302 for easier rebasing, but this does not need it.

Add unified, well-tested way to encode/decode properties as SQL JSON to avoid repeated reimplementation.

Consumers of existing behavior are modified to use raw JSON encoding -- they may be moved to properties JSON in the future if appropriate. In particular, existing consumers of existing compile_sql_filter do not handle datetimes (no way of handling type collisions -- the new method uses a discriminated union).

### Type of change

Breaking: filters will no longer ignore unknown keys.

It's confusing and potentially dangerous (especially with more complex filters including NOT, AND, OR) what treating unknown keys as None is supposed to mean. New behavior: If a filter is meant to be sent to multiple memories simultaneously, then all memories must support those fields.

It maybe would be okay if it obeyed SQL 3VL, but there is no case where a filter on one memory's field should affect the filter for another memory.

- [ ] Documentation update

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### DIFF
--- a/packages/client/client_tests/test_integration_complete.py
+++ b/packages/client/client_tests/test_integration_complete.py
@@ -353,7 +353,7 @@ class TestMemMachineIntegration:
         try:
             results = memory.search(
                 "What do I like to drink?",
-                filter_dict={"time": "morning"},
+                filter_dict={"m.time": "morning"},
                 limit=10,
             )
             assert isinstance(results, SearchResult)
@@ -488,7 +488,7 @@ class TestMemMachineIntegration:
         )
 
         # Search with custom filters only - built-in filters are automatically merged
-        custom_filters = {"category": "profession"}
+        custom_filters = {"m.category": "profession"}
         try:
             results = memory.search(
                 "What is my profession?",
--- a/packages/server/server_tests/memmachine_server/common/filter/test_sql_filter_util.py
+++ b/packages/server/server_tests/memmachine_server/common/filter/test_sql_filter_util.py
@@ -3,21 +3,43 @@
 These tests verify that JSON metadata filtering handles numeric types correctly
 (integer ordering, float ordering, boolean equality) rather than falling back
 to string/lexicographic comparison.
+
+Properties-JSON tests additionally verify the type-tagged {"t": …, "v": …} format
+against both SQLite and PostgreSQL (via testcontainers).
 """
 
+from datetime import UTC, datetime, timedelta, timezone
+from typing import Any, cast
+
 import pytest
+import pytest_asyncio
 from sqlalchemy import JSON, Column, Integer, String, create_engine, select
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
 from sqlalchemy.orm import DeclarativeBase, Session
 
-from memmachine_server.common.filter.filter_parser import parse_filter
+from memmachine_server.common.filter.filter_parser import (
+    And,
+    Comparison,
+    In,
+    IsNull,
+    Not,
+    Or,
+    parse_filter,
+)
 from memmachine_server.common.filter.sql_filter_util import compile_sql_filter
+from memmachine_server.common.properties_json import encode_properties
+
+# ============================================================================
+# Raw JSON ("json" kind) + direct column ("column" kind) tests — sync SQLite
+# ============================================================================
 
 
-class Base(DeclarativeBase):
+class _JsonBase(DeclarativeBase):
     pass
 
 
-class Item(Base):
+class _Item(_JsonBase):
     __tablename__ = "items"
 
     id = Column(Integer, primary_key=True, autoincrement=True)
@@ -25,32 +47,32 @@ class Item(Base):
     json_metadata = Column(JSON, nullable=True)
 
 
-def _resolve_field(field: str):
-    """Field resolver for the Item model."""
+def _resolve_json_field(field: str):
+    """Field resolver for the Item model: column + json kinds."""
     normalized = field.lower()
     field_mapping = {
-        "id": Item.id,
-        "name": Item.name,
+        "id": _Item.id.expression,
+        "name": _Item.name.expression,
     }
     if normalized in field_mapping:
-        return field_mapping[normalized], False
+        return field_mapping[normalized], "column"
 
     if normalized.startswith(("m.", "metadata.")):
         key = normalized.split(".", 1)[1]
-        return Item.json_metadata[key], True
+        return _Item.json_metadata[key], "json"
 
-    return None, False
+    raise ValueError(f"Unknown filter field: {field!r}")
 
 
 @pytest.fixture
-def session():
+def json_session():
     """Create an in-memory SQLite database with seeded rows."""
     engine = create_engine("sqlite:///:memory:")
-    Base.metadata.create_all(engine)
+    _JsonBase.metadata.create_all(engine)
     with Session(engine) as s:
         s.add_all(
             [
-                Item(
+                _Item(
                     name="alpha",
                     json_metadata={
                         "count": 5,
@@ -59,7 +81,7 @@ def session():
                         "tag": "a",
                     },
                 ),
-                Item(
+                _Item(
                     name="beta",
                     json_metadata={
                         "count": 10,
@@ -68,7 +90,7 @@ def session():
                         "tag": "b",
                     },
                 ),
-                Item(
+                _Item(
                     name="gamma",
                     json_metadata={
                         "count": 15,
@@ -77,7 +99,7 @@ def session():
                         "tag": "c",
                     },
                 ),
-                Item(
+                _Item(
                     name="delta",
                     json_metadata={
                         "count": 20,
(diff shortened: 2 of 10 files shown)


edwinyyyu:
Need to support system-defined metadata too. Will combine with existing compile_sql_filter.

edwinyyyu:
Discovered client tests were wrong all along -- filters silently failed before. Need to fix.

sscargal:
Breaking Changes should be delayed or moved to 0.4.0, IMHO. We're not strictly following SemVer and are currently using MINOR versions for breaking changes.

sscargal:
Reviewed these changes with CoPilot. 

## What Changed

The breaking change is in `compile_sql_filter` and all `_resolve_field` callbacks across the codebase.

### The Core Change: Unknown Filter Fields Now Raise Instead of Silently Returning `None`

**Before** — unknown filter fields were silently ignored. The resolver returned `(None, False)`, and `compile_sql_filter` would swallow it, treating a filter on an unknown field as if it didn't exist (returning `None`):

```python name=sql_filter_util.py (before)
# resolve_field returned (None, False) for unknown fields
def _resolve_episode_field(field: str) -> tuple[Any, bool] | tuple[None, bool]:
    ...
    return None, False  # ← unknown field: silently ignored

# compile_sql_filter then treated None as "no filter"
def compile_sql_filter(expr, resolve_field) -> ColumnElement[bool] | None:
    if isinstance(expr, And):
        left = compile_sql_filter(expr.left, resolve_field)
        right = compile_sql_filter(expr.right, resolve_field)
        if left is None:   # ← if one side was unknown, just skip it
            return right
        if right is None:
            return left
        return and_(left, right)
```

**After** — unknown fields raise `ValueError`, and `compile_sql_filter` always returns a valid expression (never `None`):

```python name=sql_filter_util.py (after)
# resolve_field now raises for unknown fields
def _resolve_episode_field(field: str) -> tuple[ColumnElement, FieldEncoding]:
    ...
    raise ValueError(f"Unknown filter field: {field!r}")  # ← explicit error

# compile_sql_filter now always returns a clause, never None
def compile_sql_filter(expr, resolve_field) -> ColumnElement[bool]:
    if isinstance(expr, And):
        return and_(
            compile_sql_filter(expr.left, resolve_field),
            compile_sql_filter(expr.right, resolve_field),
        )
```

---

## Before vs. After: Concrete Code Example

### Scenario: Filtering on an unknown field

```python name=before_example.py
# BEFORE: No error. The unknown field is silently dropped.
# If you have two memories and one doesn't know about "category",
# the filter just vanishes — you get ALL records instead of an error.

results = memory.search(
    "What is my profession?",
    filter_dict={"category": "profession"},  # ← "category" unknown to this memory
    limit=10,
)
# Silently returns everything — dangerous, especially with NOT/AND/OR!
# e.g. NOT(unknown_field = x) would have matched ALL rows
```

```python name=after_example.py
# AFTER: Raises ValueError immediately — no silent data leak.

results = memory.search(
    "What is my profession?",
    filter_dict={"category": "profession"},  # ← unknown field
    limit=10,
)
# Raises: ValueError: Unknown filter field: 'category'
```

The integration tests in the PR demonstrate this directly:

```python name=test_semantic_storage.py url=https://github.com/MemMachine/MemMachine/pull/1291
# BEFORE: Test verified NO error occurred
await semantic_storage.get_feature_set(
    filter_expr=_expr("missing_column IN (foo)"),
)

# AFTER: Test now asserts an error IS raised
with pytest.raises(ValueError, match="Unknown filter field"):
    await semantic_storage.get_feature_set(
        filter_expr=_expr("missing_column IN (foo)"),
    )
```

---

## Does the Breaking Change Require API Changes?

**Yes — callers using `filter_dict` with unqualified or unrecognized field names will break.** The PR already includes one such API-visible fix in the integration test client:

```python name=test_integration_complete.py url=https://github.com/MemMachine/MemMachine/pull/1291
# BEFORE: bare key
filter_dict={"time": "morning"}
filter_dict={"category": "profession"}

# AFTER: must be prefixed with "m." to target the JSON metadata column
filter_dict={"m.time": "morning"}
filter_dict={"m.category": "profession"}
```

The `m.` (or `metadata.`) prefix is required for **user-defined metadata fields**. Without it, the resolver doesn't recognize the field and now raises a `ValueError` instead of silently passing through.

### Summary of Required API Changes for Callers

| Scenario | Before | After | Action Required |
|---|---|---|---|
| Filter on user metadata field | `{"time": "morning"}` | `{"m.time": "morning"}` | Add `m.` prefix |
| Filter on unknown field shared across memories | Silently ignored | `ValueError` raised | Fix or remove the filter |
| `NOT`/`AND`/`OR` with an unknown field | Unpredictably dropped | `ValueError` raised | Fix the field name |
| Filter on known system fields (`created_at`, `tag`, etc.) | Works | Works | No change needed |

The change is **externally visible** to any code that builds `filter_dict` queries, whether through the Python SDK or the REST API's filter parameters. Any filter using a bare key for metadata (without the `m.` prefix) will now fail at runtime instead of silently misbehaving.

sscargal:
Given this is a breaking change, we need to ensure all downstream components are updated:

- [ ] Update the Python Client SDK code (#1305)
- [ ] Update the Typescript Client SDK code 
- [ ] Update Examples
- [ ] Update documentation 
- [ ] Ensure all SDK Docstrings contain the updated information so the auto-generated `openapi.json` used by the documentation is successfully updated and accurate (#1306)
- [ ] Server, Client, Examples, and Documentation must be merged in the same Release to avoid breaking the front-end and back-end (causing incompatibility issues)

edwinyyyu:
> Given this is a breaking change, we need to ensure all downstream components are updated:
> 
>     * [ ]  Update the Python Client SDK code ([Update client libraries to support strict filter field validation and metadata prefixing for PR #1291 #1305](https://github.com/MemMachine/MemMachine/issues/1305))
> 
>     * [ ]  Update the Typescript Client SDK code
> 
>     * [ ]  Update Examples
> 
>     * [ ]  Update documentation
> 
>     * [ ]  Ensure all SDK Docstrings contain the updated information so the auto-generated `openapi.json` used by the documentation is successfully updated and accurate ([Update OpenAPI documentation and docstrings for strict filter field and prefixing changes (PR #1291) #1306](https://github.com/MemMachine/MemMachine/issues/1306))
> 
>     * [ ]  Server, Client, Examples, and Documentation must be merged in the same Release to avoid breaking the front-end and back-end (causing incompatibility issues)

The TypeScript client needs no updates, but users may need to update their filters. However, the original logic was unintuitive (not consistent with any common boolean logic system) and filters are not in any of the documentation, such that we can assume anyone dependent on the logic is depending on undefined behavior. The filter is submitted to the server as a string.

More changes may be necessary to bring the client/server to a beneficial state, so I am not confident that documentation changes would be beneficial at the moment.

The Python client does not even support the filter expressions, according to Claude. Only the TypeScript client supports them.

sscargal (approved):
LGTM

malatewang (approved):
The comments from copilot should be fixed

PR #1292, opened 2026-04-01
Embedded VectorStore Implementation(s)

edwinyyyu:
### Purpose of the change

Add embedded implementations of VectorStore.

### Description

Based on #1291.

One is based on SQLite + sqlite-vec (https://github.com/asg017/sqlite-vec).
- sqlite-vec only works with SQLite that allows loading extensions, and does not support ANN yet.

One is based on SQLite + Vector Search Engine e.g.
- hnswlib (https://github.com/nmslib/hnswlib)
- USearch (https://github.com/unum-cloud/usearch)
- also considered FAISS but it conflicts with PyTorch and it does not support deletion well 

Notes:
- sqlite-vec is not usable OotB when the Python installation does not support loading SQLite extensions. This may be hard to debug for consumers.

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code


edwinyyyu:
The SQLiteVectorStore using VectorSearchEngine can probably be generalized to SQLAlchemy in general, although that would maybe not be so useful since then it's not completely local-only.

edwinyyyu:
SEGFAULT caused by duplicate OpenMP runtimes, according to Claude.
Found: https://github.com/facebookresearch/faiss/issues/3401

edwinyyyu:
Needs wiring still.

edwinyyyu:
https://github.com/MemMachine/MemMachine/actions/runs/24857340725/job/72917682977
Could be due to uv caching of compiled artifacts.
AMD and Intel may support different optional SIMD/vector extensions.

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/hnswlib_engine.py
```diff
@@ -0,0 +1,305 @@
...
+
+        self._num_dimensions = num_dimensions
+        self._similarity_metric = similarity_metric
+
+        self._space = hnswlib_space
+        self._ef_search = ef_search
+
+        self._index = hnswlib.Index(space=hnswlib_space, dim=num_dimensions)
+        self._index.init_index(
+            max_elements=initial_capacity, ef_construction=ef_construction, M=m
```
malatewang: allow_replace_deleted should be set here.

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/vector_search_engine.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Result of a nearest-neighbor search for a single query vector.
+
+    Matches are ordered from best to worst.
+    """
+
+    matches: list[SearchMatch]
+
+
+class VectorSearchEngine(ABC):
```
malatewang, line 38: We may need a rebuild function for accuracy

packages/server/src/memmachine_server/common/vector_store/sqlite_vec_vector_store.py
```diff
@@ -0,0 +1,716 @@
...
+        limit: int | None = None,
+        property_filter: FilterExpr | None = None,
+        return_vector: bool = False,
+        return_properties: bool = True,
+    ) -> list[QueryResult]:
+        query_vectors = list(query_vectors)
+        if not query_vectors:
+            return []
+
+        effective_limit = min(limit, 4096) if limit is not None else 4096
```
malatewang: Why to use 4096 arbitrary number

packages/server/src/memmachine_server/common/vector_store/sqlite_vec_vector_store.py
```diff
@@ -0,0 +1,716 @@
...
+            for query_vector in query_vectors:
+                query_blob = self._serialize_vector(query_vector)
+
+                knn_rows = (
+                    await session.execute(
+                        text(
+                            f"SELECT rowid, distance FROM [{self._vector_table_name}] "
+                            f"WHERE vector MATCH :query AND k = :k "
+                            f"ORDER BY distance"
+                        ),
```
malatewang, line 229: should the property filter be applied here?

malatewang (changes requested):

packages/server/src/memmachine_server/common/vector_store/sqlite_vec_vector_store.py
edwinyyyu: sqlite-vec does not support all of our filters here.

packages/server/src/memmachine_server/common/vector_store/sqlite_vec_vector_store.py
edwinyyyu: I just made it a class variable. sqlite-vec hard-caps at 4096. I think it is better to clamp the limit rather than error.

packages/server/src/memmachine_server/common/vector_store/sqlite_vec_vector_store.py
edwinyyyu: From Claude:

Also filters are mutually exclusive with ANN indexing (not released yet) in sqlite-vec codebase.
- aside: sqlite-vec partitions are also mutually exclusive with ANN indexing

And filtering in the vec0 table requires a fixed schema with only up to 16 columns.

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/hnswlib_engine.py
edwinyyyu: Setting that option causes index corruption if IDs are reused (bug or unsupported edge case in hnswlib). I'm implementing it, but it requires relying on weird hnswlib API behavior if we don't want to make the VectorSearchEngine API very ugly. I prefer to keep the API sane instead of making the API worse for the caller but easier to implement for a particular engine.

It's also ~10x slower when replacing deleted.

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/vector_search_engine.py
edwinyyyu: I don't see any evidence that this is needed.

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/hnswlib_engine.py
```diff
@@ -0,0 +1,365 @@
...
+        vectors: Iterable[Sequence[float]],
+        *,
+        limit: int,
+        allowed_keys: Container[int] | None = None,
+    ) -> list[SearchResult]:
+        vectors = list(vectors)
+        if self._index.element_count == 0 or not vectors:
+            return [SearchResult(matches=[]) for _ in vectors]
+
+        async with self._lock:
```
malatewang: For performance, read-write lock should be used here.

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/hnswlib_engine.py
```diff
@@ -0,0 +1,365 @@
...
+        k: int,
+        filter_fn: Callable[[int], bool] | None,
+    ) -> list[SearchResult]:
+        # Fast path: requested k works for the full batch.
+        result = self._try_knn_query(query, k, filter_fn)
+        if result is not None:
+            return self._build_search_results_from_knn(result)
+
+        # Binary search with a single probe to find max fillable k.
+        max_k = self._find_max_k(query[0:1], k, filter_fn)
```
malatewang: Why do you think the first vector is good enough in the fallback code path

packages/server/src/memmachine_server/common/vector_store/sqlite_vector_store.py
```diff
@@ -0,0 +1,1004 @@
...
+    ) -> VectorSearchEngine:
+        cache_key = (namespace, name)
+        if cache_key not in self._search_engines:
+            search_engine = self._vector_search_engine_factory(
+                config.vector_dimensions, config.similarity_metric
+            )
+
+            path = self._index_path(namespace, name)
+            if path is not None and path.exists():
+                await search_engine.load(str(path))
```
malatewang: Need to handle error condition here: file missing, corruption

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/usearch_engine.py
```diff
@@ -0,0 +1,192 @@
...
+        vectors: Iterable[Sequence[float]],
+        *,
+        limit: int,
+        allowed_keys: Container[int] | None = None,
+    ) -> list[SearchResult]:
+        vectors = list(vectors)
+        if self._index.size == 0 or not vectors:
+            return [SearchResult(matches=[]) for _ in vectors]
+
+        async with self._lock:
```
malatewang: Use read-write lock for performance

malatewang (changes requested):

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/hnswlib_engine.py
edwinyyyu: The same results (but different ordering per query vector) will be returned for every single vector if the limit is higher than the number of vectors, because the binary search finds the max fillable k. hnswlib is different from most other engines as it tries to fill the k value exactly, and it errors if it cannot fill k.

edwinyyyu:
Wiring will result in merge conflicts with other VectorStore implementations. I would like to do it in a later PR since there are no consumers yet.

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/usearch_engine.py
edwinyyyu: Done.

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/hnswlib_engine.py
edwinyyyu: Done.

packages/server/src/memmachine_server/common/vector_store/sqlite_vector_store.py
edwinyyyu: I've updated it to crash on missing or corrupted index since those are not recoverable while maintaining the same search performance. I don't think we should store an extra copy of the vectors in SQLite because corruption there also faces the same story.

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/hnswlib_engine.py
edwinyyyu: Corrections:
- Claude can engineer adversarial inputs with ingestion only to disconnect the bottom-level graph
  - methods include:
    - simplex (many unique near-equidistant vectors)
    - many near duplicate vectors (possible in benign input in extreme use cases)
- with allow_replace_deleted=False, disconnection cannot occur on deletion
- with allow_replace_deleted=True, disconnection can occur on deletion
- binary search assumes connected bottom-level graph, which is not true in adversarial cases

edwinyyyu:
- there is no index corruption under allow_replace_deleted=False from what Claude can tell
- index quality degrades faster with add/delete churn when allow_replace_deleted=True with synthetic randomly generated vectors
  - does not matter much for real embeddings

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/hnswlib_engine.py
edwinyyyu: The original code would work but would be wrong in the sense that it could produce unexpected results.

edwinyyyu: Rebuilding index is not likely to fix the problem because the disconnection depends on geometry instead of ingestion history.

packages/server/src/memmachine_server/common/vector_store/vector_search_engine/hnswlib_engine.py
```diff
@@ -0,0 +1,334 @@
...
+        for labels, distances in zip(all_labels, all_distances, strict=True):
+            matches = [
+                SearchMatch(key=int(label), score=self._distance_to_score(float(dist)))
+                for label, dist in zip(labels, distances, strict=True)
+                if int(label) >= 0
+            ]
+            results.append(SearchResult(matches=matches))
+        return results
+
+    def _find_max_k(
```
malatewang: This function should return the knn_query data. So the caller can just use the result directly without performing another knn_query.

malatewang (approved):

PR #1327, opened 2026-04-15
fix(semantic-memory): break ingestion retry loop on <think>-prefixed LLM responses

### LINKED ISSUE #1270: [Bug]: Dead loop to trigger memmachine_server.semantic_memory.semantic_ingestion when LLM reponse is invalid
### Describe the bug

This is a related issue with #1265 

When the issue in #1265 happened, we could see semantic_ingestion keep called repeatly and report errors.
Does this waste user's token? If yes, it is a critical issue.

[memmachine-output.log](https://github.com/user-attachments/files/26261542/memmachine-output.log)

### Steps to reproduce

- Configure to use qwen3.5-plus or qwen3.5-flash in configuration.yml
- add some memory to generate semantic memory and you will see such errors in docker logs.

### Expected behavior

MemMachine add/search memory should compatible with qwen3.5 LLM response.

### Environment

memmachine 0.3.2
qwen3.5-plus, qwen3.5-flash

### Additional context

_No response_

sscargal:
### Purpose of the change

Qwen3.5 (and other reasoning-mode models) prefix responses with <think>...</think> blocks. The OpenAI Responses client returned this text verbatim, pydantic rejected it in llm_feature_update, and the generic except branch in _process_single_set re-queued the message on every 2-second ingestion tick — burning LLM tokens indefinitely for any user on a reasoning-mode model.

### Description

Fix A (packages/server/src/memmachine_server/common/language_model/openai_responses_language_model.py): strip leading <think>...</think> blocks at the LLM client boundary. Applied at both output_text return sites in _generate_response; added a fallback in generate_parsed_response that strips and parses via json_repair + TypeAdapter when the SDK's output_parsed is None.

Fix B (packages/server/src/memmachine_server/semantic_memory/semantic_ingestion.py): catch pydantic.ValidationError in process_semantic_type and mark the message uid as ingested, mirroring the existing context_length_exceeded skip path. Narrow to ValidationError only — architect ruled out a broad (ValueError, TypeError) catch because those leak from the embedder and storage call sites. Generic except Exception and the _debug_fail_loudly re-raise are preserved so unknown errors stay loud.

Fix A closes the immediate trigger; Fix B is defense-in-depth against the next permanent-failure shape (provider schema drift, truncated responses) that will surface the same way.

Tests:
- test_openai_responses_language_model.py: 3 new cases (leading <think>, multi-line <think>, output_parsed=None fallback).
- test_semantic_ingestion.py: ValidationError marks message as ingested, and a guard test confirming generic Exception does NOT mark it — defends against future over-broad widening of the catch.

### Fixes/Closes

Fixes #1270

### Type of change

### How Has This Been Tested?

- [x] Unit Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### Screenshots/Gifs

N/A

### Further comments

None

### DIFF
--- a/packages/server/src/memmachine_server/common/language_model/openai_responses_language_model.py
+++ b/packages/server/src/memmachine_server/common/language_model/openai_responses_language_model.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import logging
+import re
 from typing import Any, TypeVar, cast
 from uuid import uuid4
 
@@ -13,7 +14,7 @@ from openai.types.responses import (
     ResponseInputParam,
     ToolParam,
 )
-from pydantic import BaseModel, Field, InstanceOf
+from pydantic import BaseModel, Field, InstanceOf, TypeAdapter
 
 from memmachine_server.common.data_types import ExternalServiceAPIError
 from memmachine_server.common.metrics_factory import MetricsFactory, OperationTracker
@@ -23,6 +24,12 @@ from .language_model import LanguageModel
 T = TypeVar("T")
 
 logger = logging.getLogger(__name__)
+
+_THINK_BLOCK_RE = re.compile(r"<think>.*?</think>", re.DOTALL)
+
+
+def _strip_think_blocks(text: str) -> str:
+    return _THINK_BLOCK_RE.sub("", text).strip()
 
 
 class OpenAIResponsesLanguageModelParams(BaseModel):
@@ -173,7 +180,18 @@ class OpenAIResponsesLanguageModel(LanguageModel):
 
             self._collect_usage_metrics(response)
 
-            return response.output_parsed
+            if response.output_parsed is not None:
+                return response.output_parsed
+
+            raw_text = response.output_text or ""
+            if raw_text:
+                raw = _strip_think_blocks(raw_text)
+                if raw:
+                    return TypeAdapter(output_format).validate_python(
+                        json_repair.loads(raw)
+                    )
+
+            return None
 
     async def generate_response(
         self,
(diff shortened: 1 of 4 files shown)


packages/server/src/memmachine_server/common/language_model/openai_responses_language_model.py
```diff
@@ -24,6 +25,12 @@
 
 logger = logging.getLogger(__name__)
 
+_THINK_BLOCK_RE = re.compile(r"<think>.*?</think>", re.DOTALL)
+
+
+def _strip_think_blocks(text: str) -> str:
+    return _THINK_BLOCK_RE.sub("", text).strip()
```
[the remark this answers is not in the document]
sscargal: For now, we'll remove all `think` blocks. We can revisit this if necessary.

edwinyyyu (commented):
What if I want the output to have a `<think></think>` block on purpose? What about another model that uses e.g. `<reason></reason>` blocks instead?

And would this apply to that model no matter which API it's going through? If so, I think some generalized optional sanitization utility would be preferable, that has many checkbox-like options for which sanitization operations to apply.

packages/server/src/memmachine_server/common/language_model/openai_responses_language_model.py
```diff
@@ -24,6 +25,12 @@
 
 logger = logging.getLogger(__name__)
 
+_THINK_BLOCK_RE = re.compile(r"<think>.*?</think>", re.DOTALL)
```
malatewang, line 28: I think the response API is only supported by openAI, which does not return the "think" block.

edwinyyyu: I think the issue is using third-party models with the OpenAI API. OpenAI does a good job of avoiding this problem.

By baking the solution into the API wrapper instead of a sanitization layer, it can ruin intended outputs from a model that follows instructions well, and we have to play whack-a-mole on every single API which a model that does not follow instructions well may use.

Issue #1297, opened 2026-04-03
[Feat]: Server tech debt resolution wishlist


edwinyyyu:
### Is your feature request related to a problem?

- horizontal scalability
Claude:
```
❯ Rate horizontal scalability of each component.

⏺ Based on what I've read in the codebase:

  ┌──────────────────────────────────┬────────┬───────────────────────────────────────────────────────────────────────────────────┐
  │            Component             │ Rating │                                    Bottleneck                                     │
  ├──────────────────────────────────┼────────┼───────────────────────────────────────────────────────────────────────────────────┤
  │ MemMachine                       │ B      │ Mostly stateless orchestrator, but holds a singleton _retrieval_agent with a      │
  │                                  │        │ process-local lock. Multiple instances work, minor duplication.                   │
  ├──────────────────────────────────┼────────┼───────────────────────────────────────────────────────────────────────────────────┤
  │ EpisodeStorage                   │ B+     │ Postgres via SQLAlchemy. Reads scale with replicas. Writes are single-primary.    │
  │                                  │        │ Session-scoped data is naturally shardable.                                       │
  ├──────────────────────────────────┼────────┼───────────────────────────────────────────────────────────────────────────────────┤
  │ SessionDataManager               │ B+     │ Same as EpisodeStorage — Postgres-backed, shardable by session_key.               │
  ├──────────────────────────────────┼────────┼───────────────────────────────────────────────────────────────────────────────────┤
  │                                  │        │ Process-local LRU cache with ref counting. No cross-instance cache coherence.     │
  │ EpisodicMemoryManager            │ D      │ Each server instance builds its own cache, wastes memory, risks stale state. This │
  │                                  │        │  is the main scaling bottleneck.                                                  │
  ├──────────────────────────────────┼────────┼───────────────────────────────────────────────────────────────────────────────────┤
  │ ShortTermMemory                  │ C      │ Summary state persisted in DB (good), but the in-memory message buffer is         │
  │                                  │        │ process-local. Requires the same instance to serve a session's lifetime.          │
  ├──────────────────────────────────┼────────┼───────────────────────────────────────────────────────────────────────────────────┤
  │ LongTermMemory/DeclarativeMemory │ A-     │ Stateless facade. All state in external VectorGraphStore. Scales as well as the   │
  │                                  │        │ backing store does.                                                               │
  ├──────────────────────────────────┼────────┼───────────────────────────────────────────────────────────────────────────────────┤
[...]

PR #1329, opened 2026-04-15
Semantic memory bugs, performance, scalability, unification of storage backends, and usability outside of server

edwinyyyu:
### Purpose of the change

Having ids assigned by database and returned by database is not commonly supported by many database types, and it increases ingestion latency for any consumers dependent on a canonical id for linking as a foreign key.

There is now a VectorStore interface that can wrap many different vector store backends with a unified interface.

Problems with using auto-incremented integer ids:
- Makes horizontal scalability more difficult. Using auto-incremented integer ids requires a centralized source of truth and makes moving shards non-trivial.
- Makes database migration more difficult to get correct. Migration strategy would depend on network topology and interaction with other databases, especially if changing id type.
- Ids for different systems are easier to conflate. If every database has an id 42, then it's possible to mistake where that id comes from.
- Auto-increment ids reveal data about usage that could be used by attackers.

Other problems:
- naming of hierarchy levels has been inherited from prior implementations over time and makes little sense now (try to ask someone or an LLM unfamiliar with the system what a "set", "feature", or "tag" is and what the hierarchy should be)
- existing semantic memory spans ~10 classes
- existing semantic memory system has ~170 public methods (deduplicated, still would be ~80)
- existing semantic memory partitions (set_id -- confusing name) did not have meaningful semantics beyond what generic metadata filtering could provide
  - essentially reserves a pre-defined generic metadata key for no reason
    - allows cross-set_id search, making any isolation concepts meaningless
    - cross-set_id search requires weird orchestration logic
  - new design will have partitions with actually meaningful semantics and should provide better isolation safety and performance boosts
    - will have actual enforced isolation -- an application that needs cross-partition logic needs to take on the responsibility itself -- existing use case replaced by generic metadata filtering -- more flexible and intuitive for the application
    - allows O(1) partition drops for scalability
- existing bugs
  - before calling the LLM, _features_to_llm_format() converted the old profile to dict[tag][feature] = value
    - so if multiple rows shared the same (tag, feature), later ones overwrote earlier ones in the prompt payload
- semantic memory has been causing CI failures for a long time
- semantic memory exposes features as single values to LLM but as multiple values internally
- we have no idea if clustering is beneficial

More problems identified by AI:
```
  5. DELETE command: prompt vs schema mismatch

  prompts.py:67-72 shows the LLM "delete all values for an attribute" form as a 3-key object (no value):
  {"command": "delete", "category": "Language Preferences", "attribute": "format"}
  But Command.value: str is required (data_types.py:177). If the LLM follows this example, structured-output parsing fails. If it
  always supplies a value, the "delete-all" semantics in the prompt are unreachable. Either drop that example from the prompt or make
   value: str | None = None and update _match_uuids to skip the value clause when value is None.

  6. value: true in prompt example contradicts schema

  prompts.py:60-64 — the first add example uses "value": true (boolean). Command.value is str. Same problem as #5 — the prompt
  teaches a shape the schema rejects.

The prompt/schema mismatch is present in the legacy semantic_memory code too, not just the attribute_memory rewrite.

  - Legacy semantic_memory has the same bad examples in packages/server/src/memmachine_server/semantic_memory/util/
    semantic_prompt_template.py:33-64: one add example uses "value": true, and the delete-all example omits value. But packages/
    server/src/memmachine_server/semantic_memory/semantic_model.py:28-34 still requires value: str, so structured parsing can fail
    there too.
  - The legacy delete path does not filter on value when applying deletes; packages/server/src/memmachine_server/semantic_memory/
    semantic_ingestion.py:306-324 only matches set_id, category_name, feature, and tag. So in legacy, the prompt/schema mismatch is
    real, but delete behavior itself is not value-dependent.
  - The attribute_memory rewrite has the same prompt/schema mismatch in packages/server/src/memmachine_server/semantic_memory/
    attribute_memory/prompts.py:57-87 and packages/server/src/memmachine_server/semantic_memory/attribute_memory/data_types.py:166-
    177.
  - In the rewrite, the delete path is also value-dependent: packages/server/src/memmachine_server/semantic_memory/attribute_memory/
    attribute_memory.py:649-660 matches category, attribute, and value. So the delete-all example is not just a schema mismatch
    there; it also misdescribes how deletion works.
```
^ Explains CI failures on main I guess (some tests are disabled for PRs)...

### Description

WIP making semantic memory horizontally scalable and allow configuring different storage backends more easily. Initial work done using Claude Code.

Change all ids to UUIDs.

Remove custom message queue and make it a generic interface.

Cut number of public methods from ~170 to ~30.

### Fixes/Closes

Partially addresses #1297

### Type of change

- [ ] Documentation update
- [x] Security (improves security without changing functionality)

### How Has This Been Tested?

- [ ] Unit Test
- [ ] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code


edwinyyyu:
May be better to design from scratch and just make a new parallel system if we want to solve some of these problems.

Issue #1247, opened 2026-03-20
[Bug]: Context window exceeded error not triggering retry logic during episodic summarization (original error message missing in ExternalServiceAPIError)


junttang:
### Describe the bug

When the LLM returns a context length exceeded error, MemMachine’s retry logic fails to activate. The root cause is that ExternalServiceAPIError is raised without including the original error message from the LLM. As a result, _is_exceed_context_window_error() cannot detect keywords like “context length” or “token limit”, skipping the split-and-retry mechanism.
This leads to silent failure with “Summarization failed due to unexpected error”, risking data loss.

```
2026-03-18 19:43:14,833 [ERROR] memmachine_server.common.language_model.openai_chat_completions_language_model - [call uuid: e3758a9d-e2ef-481c-b14c-f381949b03e1] Giving up generating response after failed attempt 1 due to non-retryable BadRequestError
Traceback (most recent call last):
File "/path/to/mm/packages/server/src/memmachine_server/common/language_model/openai_chat_completions_language_model.py", line 219, in _generate_response
response = await self._client.chat.completions.create(**args)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/path/to/mm/.venv/lib/python3.12/site-packages/openai/resources/chat/completions/completions.py", line 2676, in create
return await self._post(
^^^^^^^^^^^^^^^^^
File "/path/to/mm/.venv/lib/python3.12/site-packages/openai/_base_client.py", line 1884, in post
return await self.request(cast_to, opts, stream=stream, stream_cls=stream_cls)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/path/to/mm/.venv/lib/python3.12/site-packages/openai/_base_client.py", line 1669, in request
raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "You passed 202753 input tokens and requested 0 output tokens. However, the model's context length is only 202752 tokens, resulting in a maximum input length of 202752 tokens. Please reduce the length of the input prompt. (parameter=input_tokens, value=202753)", 'type': 'BadRequestError', 'param': 'input_tokens', 'code': 400}}
[...]

PR #1362, opened 2026-04-21
fix(server): preserve original LLM error in ExternalServiceAPIError

### LINKED ISSUE #1309: [Bug]: Context length errors from LLM are not detected, preventing split-and-retry mechanism for episodic summary
### Describe the bug

I am reporting that the issue raised three weeks ago in #1247 still persists in version v0.3.3.
Please refer to the original issue (#1247) for the detailed bug description, as the context remains the same.

Although a fix was previously merged via PR #1254, my testing indicates that the merged code does not resolve the problem.
The root cause is as follows:
> When generating a summary, if the LLM throws a context length error, this error message should ideally be detected by the `_is_exceed_context_window_error()` function. 
> However, for some reason, the detection fails. Consequently, the split-and-retry mechanism for summaries is never triggered, causing the process to fail outright.

As mentioned in #1247, this occurs in a specific environment where:
- Chat history data is being injected rapidly for benchmarking purposes.
- An on-prem LLM is running (vLLM).

While this specific combination might trigger the issue, this bug poses a risk for possible features involving "bulk chat history loading". 
Furthermore, it will affect any users deploying MemMachine with on-prem LLMs. 
Therefore, I believe this requires re-evaluation.

---

**Proposed Solution:**
At the end of the thread in #1247, I shared a local workaround that completely resolved the issue for me. The solution involves explicitly returning the error message from the LLM response handler, ensuring it is correctly caught by the detection logic.

Could you please review this persistent issue and consider the proposed fix?

### Steps to reproduce

Please refer to #1247.

### Expected behavior

Please refer to #1247.

### Environment

Please refer to #1247.

### Additional context

_No response_

sscargal:
### Purpose of the change

preserve original LLM error in ExternalServiceAPIError

### Description

  The non-retryable branch in OpenAIChatCompletionsLanguageModel wrapped the
  OpenAI error with a generic "due to non-retryable BadRequestError" message
  and dropped str(e). That stripped the keywords ("context length", "token
  limit", …) that ShortTermMemoryConsolidator._is_exceed_context_window_error
  relies on, so the split-and-retry path never triggered and episodic
  summaries silently failed on oversized prompts.

  Append str(e) to the wrapper message so the detector sees the original
  text and recursion kicks in as designed. Adds a regression test that
  raises openai.BadRequestError("…context length is only 202752 tokens…")
  and asserts the wrapped ExternalServiceAPIError still matches "context
  length".

### Fixes/Closes

Fixes #1309 
Refs: #1247

### Type of change

### How Has This Been Tested?

- [x] Unit Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### Screenshots/Gifs

N/A

### Further comments

None

### DIFF
--- a/packages/server/server_tests/memmachine_server/common/language_model/test_openai_chat_completions_language_model.py
+++ b/packages/server/server_tests/memmachine_server/common/language_model/test_openai_chat_completions_language_model.py
@@ -820,6 +820,25 @@ async def test_generate_response_runtime_exception_mapping(
 
 
 @pytest.mark.asyncio
+async def test_generate_response_preserves_original_error_message(
+    mock_async_openai,
+    minimal_config,
+):
+    # Regression for #1309: the wrapped ExternalServiceAPIError must carry the
+    # original LLM error text so _is_exceed_context_window_error can key off it.
+    mock_client = mock_async_openai.return_value
+    mock_client.chat.completions.create.side_effect = openai.BadRequestError(
+        "This model's maximum context length is 202752 tokens...",
+        response=MagicMock(),
+        body=None,
+    )
+
+    lm = OpenAIChatCompletionsLanguageModel(minimal_config)
+    with pytest.raises(ExternalServiceAPIError, match="context length"):
+        await lm.generate_response()
+
+
+@pytest.mark.asyncio
 async def test_metrics_collection(mock_async_openai, full_config):
     """Test that metrics are collected on a successful call."""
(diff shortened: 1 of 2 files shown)


edwinyyyu (approved):
Longer-term fix should probably define a specific error type for context exceeded case.

jealous (approved):

junttang:
Thank you for the fix!

PR #1365, opened 2026-04-22
Add encryption infrastructure

edwinyyyu:
### Purpose of the change

For supporting sensitive data encryption.
The approach to encrypting in SQL will be to store codec config (including wrapped DEK) in database.

### Description

Based on #1375 for easier rebasing, but really only creates stuff for encryption.
DOES NOT HANDLE key creation/rotation.

Alternative is encryption at rest but database sees the plaintext data.

KMSCryptoClient may get implementations with the following backends:
- OpenBao
- HashiCorp Vault
- cloud KMS such as AWS, Azure, GCP

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

- [ ] Unit Test
- [ ] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code


edwinyyyu:
May need changes to support BYOK.

jealous (approved):

edwinyyyu:
Needs a design doc.

packages/server/src/memmachine_server/common/payload_codec/aes_gcm_payload_codec.py
```diff
@@ -0,0 +1,43 @@
...
+    @override
+    def encode(self, value: bytes) -> bytes:
+        nonce = token_bytes(self._nonce_size)
+        ciphertext = self._aead.encrypt(nonce, value, self._associated_data)
+        return nonce + ciphertext
+
+    @override
+    def decode(self, value: bytes) -> bytes:
+        if len(value) < self._nonce_size:
+            raise ValueError("Encrypted payload is too short")
```
sscargal: The length check guards against a missing nonce, but AES-GCM also requires a 16-byte auth tag. A 12–27 byte buffer passes this check and then surfaces as an opaque `InvalidTag` from `AESGCM.decrypt`, which conflates *truncated ciphertext* with *active tampering* in logs/Sentry.

Suggest:

```python
if len(value) < self._nonce_size + 16:
    raise ValueError("Encrypted payload is too short (missing nonce or auth tag)")
```

The GCM tag is fixed at 16 bytes for AES-128/192/256.

packages/server/src/memmachine_server/common/payload_codec/kms_envelope_payload_codec_loader.py
```diff
@@ -0,0 +1,48 @@
...
+                )
+                return AESGCMPayloadCodec(
+                    dek,
+                    nonce_size=config.nonce_size,
+                    associated_data=config.associated_data,
+                )
+            case _:
+                raise NotImplementedError(
+                    f"Unsupported KMS envelope payload codec config: "
+                    f"{type(config).__name__}"
```
sscargal, line 47: `NotImplementedError` is conventionally for unimplemented abstract methods / unfinished features. An unsupported runtime config is a caller/data error — `TypeError` (matches Python's own behavior for unmatched `match` over types) or `ValueError` is more appropriate, so callers can distinguish *the loader is incomplete* from *you passed a bogus config*.

The test at `test_kms_envelope_payload_codec_loader.py:88-89` would need a matching update.

packages/server/src/memmachine_server/common/payload_codec/payload_codec_config.py
```diff
@@ -17,7 +20,52 @@ class PlaintextPayloadCodecConfig(BaseModel):
...
+    wrapped_dek: bytes
+    associated_data: bytes | None = None
+
+    @field_validator("wrapped_dek", mode="before")
+    @classmethod
+    def _decode_wrapped_dek(cls, value: object) -> bytes:
+        if isinstance(value, bytes):
+            return value
+        if isinstance(value, str):
+            return urlsafe_b64decode(value.encode("ascii"))
```
sscargal: Complements the Copilot note above: when `urlsafe_b64decode` rejects malformed input it raises `binascii.Error`, which is currently re-raised unwrapped. `AGENTS.md` L94 requires `raise ... from err` chaining so the original cause is preserved. Suggest:

```python
import binascii

if isinstance(value, str):
    try:
        return urlsafe_b64decode(value.encode("ascii"))
    except (binascii.Error, ValueError) as e:
        raise ValueError("wrapped_dek is not valid base64url") from e
```

Same pattern needed for `associated_data` at line 51.

packages/server/src/memmachine_server/common/payload_codec/payload_codec_config.py
```diff
@@ -17,7 +20,52 @@ class PlaintextPayloadCodecConfig(BaseModel):
     type: Literal["plaintext"] = "plaintext"
 
 
-PayloadCodecConfigUnion = PlaintextPayloadCodecConfig
+class KMSEnvelopePayloadCodecConfig(BaseModel):
```
sscargal: `KMSEnvelopePayloadCodecConfig` is intended as an abstract base (no `type` discriminator) but is freely instantiable — `test_kms_envelope_payload_codec_loader.py:80-96` confirms this by constructing an `UnknownKMSEnvelopePayloadCodecConfig` directly.

A couple of options:

- **If genuinely abstract:** make it `ABC` (or a `typing.Protocol`) and move the base64 validators into a shared mixin. Bare instantiation should fail.
- **If purely structural:** consider composition — give each concrete config an `envelope: KMSEnvelopeParams` field instead of inheritance. That flattens the discriminated union and removes the *abstract base in a union* hazard.

Either way, consider `model_config = ConfigDict(frozen=True)` — config objects holding (wrapped) key material should be immutable.

packages/server/server_tests/memmachine_server/common/payload_codec/test_payload_codec.py
```diff
@@ -14,3 +20,21 @@ def test_plaintext_payload_codec_round_trip() -> None:
...
+    decoded = codec.decode(encoded)
+
+    assert encoded != value
+    assert decoded == value
+
+    wrong_codec = AESGCMPayloadCodec(
+        b"0" * 32,
+        associated_data=b"partition:block",
+    )
+    with pytest.raises(InvalidTag):
```
sscargal, line 39: Three defensive branches the codec adds are currently untested — easy wins:

- **`nonce_size <= 0` constructor guard** (codec line 24-25)
- **short-payload `ValueError` branch** (codec line 39-40)
- **tampered ciphertext distinct from AD mismatch** — the existing `wrong_codec` test only varies the AAD, so a regression that silently disabled GCM auth-tag verification while leaving AAD intact would still pass. Flipping a byte in the ciphertext (`buf = bytearray(encoded); buf[20] ^= 1`) and asserting `InvalidTag` would catch it.

packages/server/server_tests/memmachine_server/common/payload_codec/test_kms_envelope_payload_codec_loader.py
```diff
@@ -0,0 +1,97 @@
...
+
+@pytest.mark.asyncio
+async def test_loader_returns_aes_gcm_codec() -> None:
+    kms_client = FakeKMSCryptoClient(b"0" * 32)
+    loader = KMSEnvelopePayloadCodecLoader(kms_client)
+    config = AESGCMPayloadCodecConfig(
+        key_ref="partition_key",
+        wrapped_dek=b"wrapped-dek-bytes",
+        nonce_size=12,
+        associated_data=b"partition:context",
```
sscargal: `associated_data=None` is never exercised in the codec, config, or loader tests. Both `_decode_associated_data` / `_serialize_associated_data` `None` branches and `AESGCM.encrypt(...)` with `aad=None` are unverified. A second loader test with `associated_data=None` would close this — it's the path most production callers will hit first if they forget to pass AAD.

sscargal (changes requested):
## Summary

Nice, well-scoped infrastructure addition — clean ABCs, sensible AAD propagation, and the loader's `match` pattern leaves room for future codec types. CI is green across all platforms and the happy-path tests are correct.

Requesting changes for a few items not already covered by @Copilot's notes. Copilot has already flagged the `nonce_size` validation gap (twice) and the permissive `urlsafe_b64decode` behavior; those still apply and aren't repeated here.

## Blocking

- **GCM auth tag length check missing in `decode`** — see inline on `aes_gcm_payload_codec.py:39`. Currently conflates truncated buffers with active tampering.
- **`NotImplementedError` is the wrong exception type** for an unsupported runtime config — see inline on `kms_envelope_payload_codec_loader.py:45`.
- **`binascii.Error` from `urlsafe_b64decode` is not wrapped with `from err`** — see inlines on `payload_codec_config.py:36` and `:51`. `AGENTS.md` L94 requires `raise ... from err` chaining.

## Suggestions

- `KMSEnvelopePayloadCodecConfig` is meant as an abstract base but is instantiable and has no `type` discriminator; the test at `test_kms_envelope_payload_codec_loader.py:80` shows this leaks. See inline on `payload_codec_config.py:23`.
- Test gaps for defensive branches the author explicitly added (short-payload guard, tampered-ciphertext distinct from AD mismatch, `associated_data=None` path). See inlines on the test files.

## Tests

Existing tests cover round-trip + AD-mismatch + unsupported-config rejection — the right baseline shape. The PR checklist has *I have added unit tests* unticked despite tests existing — likely just an oversight.

## Local checks

Skipped local `ruff` / `ty` re-run. CI ran both across Python 3.12/3.13/3.14 on Linux/macOS/Windows and all 40+ checks passed.

packages/server/src/memmachine_server/common/payload_codec/aes_gcm_payload_codec.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(
+        self,
+        key: bytes,
+        *,
+        nonce_size: int = 12,
+        associated_data: bytes | None = None,
+    ) -> None:
+        """Initialize with a raw AES key."""
+        if nonce_size <= 0:
+            raise ValueError("nonce_size must be positive")
```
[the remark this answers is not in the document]
sscargal: This recommendation fixes the nonce at 12, whereas in `packages/server/src/memmachine_server/common/payload_codec/payload_codec_config.py` CoPilot suggests a default value of 12 - a safe minimum. This check should validate the safe minimum or higher, not a fixed value.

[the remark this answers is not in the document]
edwinyyyu: The actual cryptography library allows 8-128, so it would have raised anyway (but not at init time). Changing the validated range.

packages/server/src/memmachine_server/common/payload_codec/payload_codec_config.py
```diff
@@ -17,7 +20,52 @@ class PlaintextPayloadCodecConfig(BaseModel):
...
+        if value is None:
+            return None
+        return urlsafe_b64encode(value).decode("ascii")
+
+
+class AESGCMPayloadCodecConfig(KMSEnvelopePayloadCodecConfig):
+    """Codec config for AES-GCM payload encryption."""
+
+    type: Literal["aes_gcm"] = "aes_gcm"
+    nonce_size: int = 12
```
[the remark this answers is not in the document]
edwinyyyu: Done.

packages/server/src/memmachine_server/common/payload_codec/aes_gcm_payload_codec.py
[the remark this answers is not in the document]
edwinyyyu: Done. There's no helper or constants to import so it's just hardcoded for us.

packages/server/src/memmachine_server/common/payload_codec/payload_codec_config.py
```diff
@@ -17,7 +20,52 @@ class PlaintextPayloadCodecConfig(BaseModel):
...
+    @field_validator("associated_data", mode="before")
+    @classmethod
+    def _decode_associated_data(cls, value: object) -> bytes | None:
+        if value is None:
+            return None
+        if isinstance(value, bytes):
+            return value
+        if isinstance(value, str):
+            return urlsafe_b64decode(value.encode("ascii"))
+        raise TypeError("associated_data must be bytes, base64url string, or None")
```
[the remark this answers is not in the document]
edwinyyyu: Done.

packages/server/src/memmachine_server/common/payload_codec/payload_codec_config.py
```diff
@@ -17,7 +20,52 @@ class PlaintextPayloadCodecConfig(BaseModel):
...
+    associated_data: bytes | None = None
+
+    @field_validator("wrapped_dek", mode="before")
+    @classmethod
+    def _decode_wrapped_dek(cls, value: object) -> bytes:
+        if isinstance(value, bytes):
+            return value
+        if isinstance(value, str):
+            return urlsafe_b64decode(value.encode("ascii"))
+        raise TypeError("wrapped_dek must be bytes or a base64url string")
```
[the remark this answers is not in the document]
edwinyyyu: Done.

packages/server/src/memmachine_server/common/payload_codec/aes_gcm_payload_codec.py
edwinyyyu: Done.

packages/server/server_tests/memmachine_server/common/payload_codec/test_payload_codec.py
edwinyyyu: Done.

packages/server/src/memmachine_server/common/payload_codec/payload_codec_config.py
edwinyyyu: Done.

packages/server/src/memmachine_server/common/payload_codec/payload_codec_config.py
edwinyyyu: Composition is probably better than inheritance here from a design pattern perspective, but it will make config processing more complicated.

edwinyyyu: Changed.

packages/server/src/memmachine_server/common/payload_codec/kms_envelope_payload_codec_loader.py
edwinyyyu: It represents an unfinished feature, assuming that the type passed in matches the type hint. In Python, runtime type checks are excessive. If anyone cares about typing, they can use a static type checker.

PR #1378, opened 2026-04-28
Don't ensure ASCII in LLM-visible JSON strings

edwinyyyu:
### Purpose of the change

Text in non-ASCII languages is rendered poorly. A good-enough LLM can mostly deal with it but it's not as readable for humans and it costs more tokens.

### Description

Only affects Python json.dumps()

Set ensure_ascii=False in LLM-visible JSON strings.

```
import json

string = "π世界"
print(json.dumps(string)) # Old.
print(json.dumps(string, ensure_ascii=False)) # New.
```

Results in:
```
"\u03c0\u4e16\u754c"
"π世界"
```
### Type of change

### DIFF
--- a/packages/client/client_tests/test_format.py
+++ b/packages/client/client_tests/test_format.py
@@ -99,6 +99,59 @@ class TestFormatEpisodes:
         result = format_episodes([ep])
         assert json.dumps('She said "hello"') in result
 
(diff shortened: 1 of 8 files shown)


sscargal (approved):
LGTM

malatewang (approved):

PR #1394, opened 2026-05-08
Segmenter and Deriver should be async

edwinyyyu:
### Purpose of the change

Future Segmenter and Deriver implementations should not block.


malatewang (approved):

sscargal (approved):
LGTM

PR #1460, opened 2026-06-18
[sqlite store fixes 1/7] Publish vector index files atomically (but not durably)

edwinyyyu:
## Problem

`SQLiteVectorStore` persists each collection's index by calling the search engine's `save()`, which writes directly to the final path. A crash partway through leaves a truncated file on disk.

The torn file is only half of it. `_save_collection_index` trims the applied `_PendingOperationRow`s as soon as `save()` returns, and that log is the **only** other copy of those vectors — the records table has no vector column. So the trim is safe at exactly one instant: when the index at the published path holds them. Getting that instant wrong is silent, because a missing vector is indistinguishable from a vector that simply doesn't match.

## Fix

A shared `index_persistence` helper writes the index to a sibling temp file, flushes it, and swaps it into place with `os.replace`; both engines adopt it and neither gains anything else. `load` clears a temp left by an interrupted save, so a crash leaks at most one stale file per index.

After this, a save either publishes a complete index or leaves the previous one exactly where it was. Nothing else changes: no schema change, no migration, no rename of `index_path`, and no new file at the base path.

## What a save does not promise

It does not promise that the publication is **durable**. That is a decision rather than an oversight, so the reasoning is below.

A rename changes a **directory entry**, not file data, and the two have very different guarantees:

| | make it durable | available |
|---|---|---|
| file contents | `fsync` / `F_FULLFSYNC` / `FlushFileBuffers` | everywhere, documented |
| directory entry | `fsync` on a directory file descriptor | POSIX only, and not on every filesystem |

- **POSIX** closes the gap with a parent-directory fsync, but only best-effort — network and FUSE filesystems are where that bites.
- **Windows** has no equivalent at all. `os.fsync` is `_commit`, which is `FlushFileBuffers`, which is for file data; you cannot open a directory to fsync it. `MOVEFILE_WRITE_THROUGH` does not help either: its documented guarantee is scoped to "a move performed as a copy and delete operation", the cross-volume path, and says nothing about flushing NTFS metadata for a same-volume rename. The decisive evidence is SQLite's — its VFS threads a directory-sync flag through every commit-relevant directory operation, `unixDelete` honors it, and `winDelete` declares the same parameter `/* Not used on win32 */`. SQLite cares about this more than almost any software and makes no attempt on Windows.

So a power failure can roll the rename back after `save` returned, while the trim behind it stays committed.

## Why that gap stays open

Closing it means never using a directory operation as the commit point: two index slots plus a generation record written into a file that already exists, in the spirit of SQLite's `PERSIST` journal mode. An earlier revision of this PR implemented exactly that, and it worked. It was still the wrong trade, for four reasons.

**The window is narrow, and only one kind of failure lands in it.** A clean shutdown, an unhandled exception, an OOM kill, a `SIGKILL` — none of these lose a rename, because the kernel still owns the page cache and writes it back. Those failures are already covered: the pending log holds everything the engine has not been checkpointed with, and startup replays it. What is left is the machine itself dying between the rename and writeback — a power cut, a host failure, a panic — plus the filesystems where a directory fsync is best-effort or unavailable anyway.

**What it costs is recall, not correctness.** The records table is the authority on what exists, and every query hit resolves through it, so a reverted publication cannot produce a wrong or stale result — only fewer results. Concretely: the records survive, `get` still returns them, `query` stops finding them, and the exposure is bounded by the operations applied since the previous checkpoint, i.e. at most `save_threshold`. Re-upserting an affected record repairs it. That is the direction this store already tolerates, and it is repairable by the same ingest path that created the record.

**Detecting it is not cheap enough to be worth it.** Comparing the index's size against the record count is the obvious check and it does not work: `row_id`s are not reused (#1469 makes that unconditional; before it only the highest row's id can come back), so deleting a record and upserting the same uuid again moves it to a new id. A rollback spanning that pair leaves the index holding the old id and missing the new one — one extra, one missing, identical count — and a re-embedding workload produces that shape constantly. A check that does work needs an id-set digest maintained by every engine, or a full reconcile scan at every boot, which is most expensive precisely where the index is large. Neither earns its keep against a bounded recall gap.

**And the guarantee is not free to hold.** It cannot be delegated to a rename, so it has to become a layout: two slots and a generation record per index, which every engine has to implement and every future engine has to be audited against. Engines that persist by writing one file — which is what both engines here do, and what most ANN libraries expose — stop being usable as they ship. Existing indexes stop loading until they are cleared and re-ingested, which is a migration this revision no longer asks anyone to perform. Fewer guarantees, less machinery to keep correct, and a wider set of engines that can plug in unmodified.

The direction that *would* cost — a published index that will not parse — stays closed: the atomic swap prevents it, the pre-swap `fsync` narrows the window where a rename outlives the data behind it, and a saved-but-unloadable index remains a loud `IndexLoadError` rather than a silently empty rebuild.

Nothing here detects a lost publication for the caller. A deployment that needs every record searchable after a power failure must be able to re-ingest.

## Where the contract is stated

The guarantee is only useful if callers can find it, so it is written where each of them looks:

- `VectorSearchEngine.save` — returning means a later `load` reads this index or the one it replaced, never a mixture; it does **not** mean the publication survives a power failure.
- `index_persistence` — the mechanism, and why a stronger guarantee is not portable.
- `sqlite_vector_store` — what survives a process crash, what a power failure costs, and that re-ingest is the repair.
- `_save_collection_index` — why the save-then-trim order is the whole protocol.

One consumer needed more than a docstring. `VectorStoreSemanticStorage.update_feature` reads a feature's stored embedding back when the caller updates the feature without supplying one — the only place in the server that depends on the index still holding a vector. It reported a lost vector as `Vector record not found`, naming the feature, which points at the wrong thing and hides the repair; it now separates a record that is genuinely absent from one whose embedding is gone, and the second says to pass a fresh embedding.

## Tests

- `test_index_persistence.py`: the swap publishes completely or not at all, a failed write leaves the previous index intact and removes the temp, and a stale temp is cleared on load.
- Per-engine (hnswlib + usearch): a save leaves no temp file behind, and a load clears a stale temp file and still finds every key.
- Store-level: `test_a_reverted_publication_costs_search_not_records` reconstructs a lost publication deterministically — restore the previous index bytes after the trim has committed — and pins the direction it fails in: the record still resolves by uuid, and only search loses it. Missing and corrupt indexes each still raise `IndexLoadError` once `index_saved` is set.

`vector_store` suites pass (284 passed, 140 integration deselected) and `semantic_memory` suites pass (345 passed, 2 skipped, 999 integration deselected); `ruff` and `ty` clean.

## Changed from the earlier revision

This PR previously shipped the two-slot generation-record protocol and its breaking on-disk change. That guarantee has been dropped in favor of the atomic swap plus an explicit contract, so **the breaking-change notice no longer applies**: indexes written by the current code keep loading, and there is nothing to migrate.

### Port

This branch carries this change as merged into `speedkick` (#1588, `21105d6e6`), the one commit `d7a6042b2`; the branch's earlier commits are replaced. Adaptation: the commit sits above [vector store scale-out 5/5], so its docstrings and its test speak of the one-collection store's partitions; the code is #1588's.

### Stack

| # | PR | change |
|---|---|---|

| # | PR | change |
|---|---|---|
| [sqlite store fixes 1/7] | #1460 **(this PR)** | Publish vector index files atomically (but not durably) (as merged into `speedkick`, #1588) |

This PR's own change is its one commit, `d7a6042b2`; the rest of its diff is the PRs under it. Stacked on #1627; #1469 is stacked on it.

### Verification

At every commit from [sqlite store fixes 1/7] (`d7a6042b2`) to the tip `545ec401e`, on 2026-09-21: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`); 2035 tests at this PR's head `d7a6042b2`. Those commits were verified before the re-order that moved #1460 into the stage, which left their trees unchanged (`git diff` against the verified commits is empty). The 22 commits below `d7a6042b2` (the scale-out stage and the independent PRs' copies in its history), whose trees the re-order changed only by the absence of #1460's files, are being re-verified as this is written; this paragraph is replaced with the result.

### DIFF
--- a/packages/server/server_tests/memmachine_server/common/vector_store/vector_search_engine/test_usearch_engine.py
+++ b/packages/server/server_tests/memmachine_server/common/vector_store/vector_search_engine/test_usearch_engine.py
@@ -296,6 +296,42 @@ class TestPersistence:
         assert result.matches[0].key == 1
         assert result.matches[0].score == pytest.approx(1.0, abs=0.01)
 
+    @pytest.mark.asyncio
+    async def test_save_leaves_no_temp_file(self, tmp_path: Path):
+        engine = USearchVectorSearchEngine(
+            num_dimensions=NDIM, similarity_metric=SimilarityMetric.COSINE
+        )
+        await engine.add({1: _normalize([1, 0, 0])})
+
+        path = tmp_path / "test.idx"
+        await engine.save(str(path))
+
+        assert path.exists()
+        assert not (tmp_path / "test.idx.tmp").exists()
+
+    @pytest.mark.asyncio
+    async def test_load_clears_stale_temp_file(self, tmp_path: Path):
+        engine = USearchVectorSearchEngine(
+            num_dimensions=NDIM, similarity_metric=SimilarityMetric.COSINE
+        )
+        await engine.add({1: _normalize([1, 0, 0]), 2: _normalize([0, 1, 0])})
+
+        path = tmp_path / "test.idx"
+        await engine.save(str(path))
+
+        # A temp file left behind by a previously interrupted save.
+        stale_temp = tmp_path / "test.idx.tmp"
+        stale_temp.write_text("STALE")
+
+        engine2 = USearchVectorSearchEngine(
+            num_dimensions=NDIM, similarity_metric=SimilarityMetric.COSINE
+        )
+        await engine2.load(str(path))
+
+        assert not stale_temp.exists()
+        result = await _search_one(engine2, _normalize([1, 0, 0]), limit=2)
+        assert {m.key for m in result.matches} == {1, 2}
+
 
 # -- SearchResult types --
 
--- a/packages/server/server_tests/memmachine_server/common/vector_store/vector_search_engine/test_hnswlib_engine.py
+++ b/packages/server/server_tests/memmachine_server/common/vector_store/vector_search_engine/test_hnswlib_engine.py
@@ -362,6 +362,42 @@ class TestPersistence:
         result = await _search_one(engine2, _normalize([1, 0, 0]), limit=3)
         assert {m.key for m in result.matches} == {1, 3}
 
+    @pytest.mark.asyncio
+    async def test_save_leaves_no_temp_file(self, tmp_path: Path):
+        engine = HnswlibVectorSearchEngine(
+            num_dimensions=NDIM, similarity_metric=SimilarityMetric.COSINE
+        )
+        await engine.add({1: _normalize([1, 0, 0])})
+
+        path = tmp_path / "test.hnswlib"
+        await engine.save(str(path))
+
+        assert path.exists()
+        assert not (tmp_path / "test.hnswlib.tmp").exists()
+
+    @pytest.mark.asyncio
+    async def test_load_clears_stale_temp_file(self, tmp_path: Path):
+        engine = HnswlibVectorSearchEngine(
+            num_dimensions=NDIM, similarity_metric=SimilarityMetric.COSINE
+        )
+        await engine.add({1: _normalize([1, 0, 0]), 2: _normalize([0, 1, 0])})
+
+        path = tmp_path / "test.hnswlib"
+        await engine.save(str(path))
+
+        # A temp file left behind by a previously interrupted save.
+        stale_temp = tmp_path / "test.hnswlib.tmp"
+        stale_temp.write_text("STALE")
+
+        engine2 = HnswlibVectorSearchEngine(
+            num_dimensions=NDIM, similarity_metric=SimilarityMetric.COSINE
+        )
+        await engine2.load(str(path))
+
+        assert not stale_temp.exists()
+        result = await _search_one(engine2, _normalize([1, 0, 0]), limit=2)
+        assert {m.key for m in result.matches} == {1, 2}
+
 
 # -- allow_replace_deleted=True (slot reclamation) --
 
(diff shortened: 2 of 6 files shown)


packages/server/server_tests/memmachine_server/common/vector_store/vector_search_engine/test_usearch_engine.py
```diff
@@ -296,6 +296,44 @@ async def test_save_and_load(self, tmp_path: Path):
         assert result.matches[0].key == 1
         assert result.matches[0].score == pytest.approx(1.0, abs=0.01)
 
+    # -- SearchResult types --
```
edwinyyyu: Duplicate `# -- SearchResult types --` inside `TestPersistence`; the real header is at 338. Cherry-pick artifact.

edwinyyyu (commented):
This branch is not "#1588 as merged into speedkick": two of 21105d6e6's ten files are absent, packages/server/server_tests/memmachine_server/common/vector_store/test_sqlite_vector_store.py (+51, the store-level `test_a_reverted_publication_costs_search_not_records`) and packages/server/src/memmachine_server/semantic_memory/storage/vector_store_semantic_storage.py (+17/-1, the `_get_existing_vector_record` split that reports a lost embedding as lost), while the description still lists both. Nothing on main forced the drop: both targets exist unchanged on ff58f38c. The commit message also carries "* Report a lost embedding as lost, not as a missing feature" for a change the commit does not contain, and its trailer names c4b6c4d4, a local stack-main3 commit, as the cherry-pick source. The description's "`row_id`s are `AUTOINCREMENT` and never reused" is also not true at this base: `sqlite_autoincrement=True` arrives with #1469.

PR #1462, opened 2026-06-19
Fix timezone-aware datetime roundtrip on SQLite

edwinyyyu:
## Problem

`SQLAlchemySegmentStore` (and a couple of other SQLAlchemy-backed stores) can corrupt timezone-aware timestamps when running on **SQLite**.

SQLite's `DateTime(timezone=True)` does not store timezone information. SQLAlchemy's SQLite dialect discards `tzinfo` and persists the datetime's *wall-clock fields* verbatim, reading them back **naive**. So when a store wrote a timezone-aware datetime *without first normalizing it to UTC*, the stored wall-clock was the local time of the original zone. On read the value was treated as UTC and then shifted to the recorded offset — corrupting the instant by the original UTC offset.

Concretely, a segment timestamp of `2024-01-01 13:30:45-08:00` came back as `2024-01-01 05:30:45-08:00`.

PostgreSQL is **unaffected** — `timestamptz` stores a true instant — so this change is a no-op there.

### Proof

```
original                     : 2026-06-19 13:26:43-08:00 -> instant 2026-06-19 21:26:43+00:00
read back (current write)    : datetime.datetime(2026, 6, 19, 13, 26, 43) tzinfo=None
reconstructed (current)      : 2026-06-19 05:26:43-08:00   CORRECT? False
reconstructed (UTC-norm)     : 2026-06-19 13:26:43-08:00   CORRECT? True
```

## Fix

Normalize to UTC (`ensure_tz_aware(...).astimezone(UTC)`) *before* persisting wherever an app-supplied timezone-aware datetime is written to a `DateTime(timezone=True)` column:

- **`SQLAlchemySegmentStore.timestamp`** — the primary report. The original offset is already stored separately (`timestamp_timezone_offset`) and reapplied on read; only the write was missing UTC normalization.
- **`ClusterStateStorageSqlAlchemy`** — `last_ts` and pending `created_at`.
- **`SqlAlchemyEpisodeStore`** — `created_at`, plus the `start_time`/`end_time` filter bounds so range comparisons stay consistent with the stored UTC instant on SQLite.

Server-generated columns (`server_default=func.now()`) are always UTC and roundtrip fine, so they are left unchanged. No schema changes; no migration (existing SQLite rows are not rewritten — the read path already assumes UTC, which new writes now honor).

## Tests

Added SQLite regression tests covering non-UTC timezones (`-08:00` and `+05:30`) for the segment store, cluster store, and episode store. Each was confirmed to **fail before** the fix and **pass after**. Existing tests for the affected modules continue to pass (952 passed locally, non-integration).

### DIFF
--- a/packages/server/server_tests/memmachine_server/common/episode_store/test_episode_storage.py
+++ b/packages/server/server_tests/memmachine_server/common/episode_store/test_episode_storage.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import asyncio
-from datetime import UTC, datetime, timedelta
+from datetime import UTC, datetime, timedelta, timezone
 
 import pytest
 import pytest_asyncio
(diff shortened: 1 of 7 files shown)


edwinyyyu:
Pushed a follow-up commit: the comparison side needed the same normalization as the write side.

Storing normalizes to UTC because SQLite's `DateTime(timezone=True)` does not persist tzinfo. Filtering did not: an aware value bound as-is is rendered with its own offset and compared against the stored text lexically, so the **wall clock** decides. `timestamp <= date('2024-01-01T08:00:30+08:00')` excluded a row stored at `2024-01-01T00:00:30Z` — thirty seconds *earlier* than the bound.

`_compile_column_leaf` now normalizes aware datetimes to UTC, the same conversion the write path and the typed-JSON comparison path already do. PostgreSQL compares `TIMESTAMPTZ` by instant and is unaffected, so this makes the two backends agree rather than changing Postgres.

Test asserts that one instant spelled in three different zones selects the same rows, parametrized over both backends.

edwinyyyu:
Superseded by #1661 (folded in).

PR #1469, opened 2026-07-01
[sqlite store fixes 2/7] Never reuse a row id in SQLiteVectorStore

### LINKED ISSUE #1468: SQLiteVectorStore: row_id reuse + concurrent async writes silently lose a record's vector
## Summary

In `SQLiteVectorStore`, a `delete()` racing an `upsert()` on the same collection can silently lose the upserted record's vector when SQLite reuses the deleted record's `row_id`. The record's row then exists and its pending operation is marked `applied`, but the search engine holds no vector for it — the record is silently unsearchable, and the next index save persists its absence. Deterministic reproduction below (plus a control showing the identical interleaving with a fresh `row_id` loses nothing).

## Mechanism

Two ingredients:

1. **`row_id` values are reused.** The per-collection records table declares `Column("row_id", Integer, primary_key=True, autoincrement=True)` without `sqlite_autoincrement=True` on the `Table`, so SQLAlchemy emits a plain `row_id INTEGER PRIMARY KEY` — a rowid alias. SQLite assigns `max(rowid) + 1`, which means deleting the row holding the current maximum frees that id for the very next insert (an emptied table restarts at 1).

2. **The SQL commit and the engine apply are separate steps with an `await` between them.** `delete()` commits its transaction (records row deleted → `row_id` freed; pending `delete` op staged) and only then `await`s `self._search_engine.remove(record_row_ids)` (an `asyncio.to_thread` in the usearch/hnswlib engines). At that await, the event loop can run other tasks.

Interleaving (single event loop, two tasks):

```
task 1: delete(A)   — A holds row_id N (the max rowid)
  ├─ tx: stage pending (N, 'delete'), DELETE records row  → COMMIT (N is now free)
  ├─ await engine.remove([N])          ← suspended here
task 2: upsert(B)   — runs to completion in that window
  ├─ tx: INSERT records row for B → SQLite REUSES row_id N
  │      pending row (ns, name, N) overwritten to ('upsert', vec_B, applied=False) → COMMIT
  ├─ engine.remove([N]); engine.add({N: vec_B})
  └─ mark pending (N) applied=True
task 1 (resumes):
  ├─ engine.remove([N])                ← deletes B's vector
  └─ mark applied WHERE row_id IN (N) AND applied IS False → no-op (already True)
```

Durable end state: B's records row exists; the pending log says `(N, 'upsert', applied=True)`; the engine holds nothing at `N`. `get()` returns B, `query()` never will, and `_save_collection_index` will trim the pending op and persist an index without B. Nothing ever heals it (startup replay skips it too, since the op is present but the vector was never re-applied — replay does re-apply all rows, but only after a restart; within the running process the record is simply gone from search).

The same root cause also affects the read path: a `query()` holds no lock between `engine.search` and `_build_matches`, so a hit on `N` scored against the old record's vector can reverse-map to the *new* record that reused `N`, attributing a stale score to the wrong record.

## Why this is within the documented contract

The engine ABC explicitly blesses concurrency: *"Safe for concurrent use from async tasks (single event loop)."* Neither `VectorStore` nor `VectorStoreCollection` documents any requirement that writes to a collection be serialized by the caller. So two concurrent tasks calling `delete` and `upsert` on one collection appear to be supported usage.

## Reproduction

Runnable script (usearch engine). The `GatedEngine` wrapper only widens the already-awaitable window at `engine.remove` to make the interleaving deterministic — `remove` is already an `await` point in the stock engines.

```python
import asyncio
import os
import sqlite3
import tempfile
from uuid import uuid4

from memmachine_server.common.data_types import SimilarityMetric
from memmachine_server.common.vector_store.data_types import (
    Record,
    VectorStoreCollectionConfig,
)
from memmachine_server.common.vector_store.sqlite_vector_store import (
    SQLiteVectorStore,
    SQLiteVectorStoreParams,
)
from memmachine_server.common.vector_store.vector_search_engine.usearch_engine import (
    USearchVectorSearchEngine,
)
from memmachine_server.common.vector_store.vector_search_engine.vector_search_engine import (
    VectorSearchEngine,
)
from sqlalchemy.ext.asyncio import create_async_engine

DIMS = 8
VEC_A = [1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
VEC_B = [0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
VEC_C = [0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0]


class GatedEngine(VectorSearchEngine):
    """Delegates to a real engine; the next `remove` waits on a gate first."""

    def __init__(self, inner: VectorSearchEngine) -> None:
        self.inner = inner
        self.gate: asyncio.Event | None = None

    async def add(self, vectors):
        return await self.inner.add(vectors)

    async def remove(self, keys):
        if self.gate is not None:
            gate, self.gate = self.gate, None
            await gate.wait()
        return await self.inner.remove(keys)

    async def search(self, vectors, *, limit, allowed_keys=None):
        return await self.inner.search(vectors, limit=limit, allowed_keys=allowed_keys)

    async def get_vectors(self, keys):
        return await self.inner.get_vectors(keys)

    async def save(self, path):
        return await self.inner.save(path)

    async def load(self, path):
        return await self.inner.load(path)


def records_table_rows(db_path: str, namespace: str, name: str):
    table = f"vector_store_sqlite_{len(namespace)}_{namespace}_{len(name)}_{name}_rc"
    conn = sqlite3.connect(db_path)
    try:
        return conn.execute(f"SELECT row_id, hex(uuid) FROM [{table}]").fetchall()
    finally:
        conn.close()


def pending_rows(db_path: str):
    conn = sqlite3.connect(db_path)
    try:
        return conn.execute(
            "SELECT record_row_id, operation_type, applied"
            " FROM vector_store_sqlite_pd_op"
        ).fetchall()
    finally:
        conn.close()


async def run_scenario(*, with_reuse: bool) -> dict:
    tmp = tempfile.mkdtemp()
    db_path = os.path.join(tmp, "store.db")
    holder: dict[str, GatedEngine] = {}

    def factory(ndim: int, metric: SimilarityMetric) -> VectorSearchEngine:
        engine = GatedEngine(
            USearchVectorSearchEngine(num_dimensions=ndim, similarity_metric=metric)
        )
        holder["engine"] = engine
        return engine

    store = SQLiteVectorStore(
        SQLiteVectorStoreParams(
            sqlalchemy_engine=create_async_engine(f"sqlite+aiosqlite:///{db_path}"),
            vector_search_engine_factory=factory,
            index_directory=None,
            save_threshold=1000,
        )
    )
    await store.startup()
    collection = await store.open_or_create_collection(
        namespace="ns",
        name="c",
        config=VectorStoreCollectionConfig(vector_dimensions=DIMS),
    )

    uuid_a, uuid_b, uuid_c = uuid4(), uuid4(), uuid4()

    # Seed A. In the control, also seed C so A's row_id is NOT the max rowid
    # and B's insert cannot reuse it.
    await collection.upsert(records=[Record(uuid=uuid_a, vector=VEC_A, properties={})])
    if not with_reuse:
        await collection.upsert(
            records=[Record(uuid=uuid_c, vector=VEC_C, properties={})]
        )

    # Concurrent writers: delete(A) commits, then blocks at its engine.remove;
    # upsert(B) runs to completion in that window.
    gate = asyncio.Event()
    holder["engine"].gate = gate
    delete_task = asyncio.create_task(collection.delete(record_uuids=[uuid_a]))
    await asyncio.sleep(0.2)  # let delete commit and reach the gated remove
    await collection.upsert(records=[Record(uuid=uuid_b, vector=VEC_B, properties={})])
    gate.set()
    await delete_task

    got_b = await collection.get(record_uuids=[uuid_b])
    results = await collection.query(query_vectors=[VEC_B], limit=5)
    b_searchable = any(
        match.record.uuid == uuid_b for result in results for match in result.matches
    )
    return {
        "records rows (row_id, uuid)": records_table_rows(db_path, "ns", "c"),
        "pending rows (row_id, op, applied)": pending_rows(db_path),
        "B record exists": len(got_b) == 1,
        "B searchable": b_searchable,
    }


async def main() -> None:
    print("=== scenario 1: id REUSE (delete frees the max rowid; B reuses it) ===")
    for key, value in (await run_scenario(with_reuse=True)).items():
        print(f"  {key}: {value}")
    print("=== scenario 2: control, NO reuse (B gets a fresh rowid) ===")
    for key, value in (await run_scenario(with_reuse=False)).items():
        print(f"  {key}: {value}")


if __name__ == "__main__":
    asyncio.run(main())
```

Observed output:

```
=== scenario 1: id REUSE (delete frees the max rowid; B reuses it) ===
  records rows (row_id, uuid): [(1, '3665...')]        <- B reused row_id 1
  pending rows (row_id, op, applied): [(1, 'upsert', 1)]
  B record exists: True
  B searchable: False                                  <- LOST
=== scenario 2: control, NO reuse (B gets a fresh rowid) ===
  records rows (row_id, uuid): [(3, '...'), (2, '...')]
  pending rows (row_id, op, applied): [(1, 'delete', 1), (2, 'upsert', 1), (3, 'upsert', 1)]
  B record exists: True
  B searchable: True                                   <- identical interleaving, no reuse, no loss
```

## Severity / masking

Real but narrow: it needs (a) the deleted record to hold the maximum `row_id`, (b) a concurrent insert landing in the commit→engine-remove window, and (c) callers that actually issue concurrent writes to one collection. If callers serialize writes per collection in practice, it is unreachable — but that requirement is not documented anywhere.

## Possible fixes

1. **`sqlite_autoincrement=True` on the records table** — never-reused `row_id`s make unrelated writes id-disjoint, so the delete's `remove` can only ever target the dead id. Smallest change; costs SQLite's `sqlite_sequence` bookkeeping write per insert. (The read-path stale-score case also degrades gracefully: a stale engine key maps to no row and is dropped instead of mis-attributed.)
2. **Document the contract**: writes to a collection must be serialized by the caller (matching the engine ABC's single-loop framing), making the current schema sound by convention.
3. **Serialize commit→apply per collection** (e.g. a per-collection async lock spanning the transaction and the engine apply), closing the window itself.

---

*Filed by Claude Code (AI agent) on behalf of @edwinyyyu.*

## Update: a second defect in the same window, which fix 1 cannot reach

Ingredient 2 above -- the commit and the engine apply being separate steps with an `await` between them -- is a defect on its own, independent of which `row_id` a record gets. `AUTOINCREMENT` makes *unrelated* writes id-disjoint, but two writes addressing **one uuid** share a `row_id` by design: `upsert` is `ON CONFLICT DO UPDATE`, so re-upserting an existing uuid keeps its row. Nothing about the id policy orders them.

**Inversion of same-uuid writes.** `upsert(U)` commits (pending row `(N, 'upsert', vec, applied=False)`) and suspends at its engine apply. `delete(U)` then runs its whole transaction -- overwriting that pending row to `('delete', applied=False)` and deleting the records row -- commits, removes `N` from the engine, and marks the row applied. The upsert resumes and re-adds `vec` at `N`.

Durable end state: no records row, pending row `(N, 'delete', applied=True)`, and a vector at `N` in the engine that belongs to nothing. It resolves to no row, so every result it wins is silently dropped -- `query(limit=k)` returns fewer than `k` matches while the engine found `k` -- and the next save publishes it into the index and trims the log row that would have repaired it on replay. Two concurrent upserts of one uuid invert the same way, leaving the index serving the older vector while SQLite records the newer.

**A save in the same window loses a committed write.** `_save_collection_index` writes the index and then deletes every pending row marked applied, because the log holds the only other copy of those vectors. A write that applies to the engine after `search_engine.save(path)` returns but before that trim commits is in neither: not in the file just written, and no longer in the log. It is live in the running engine and absent from disk, so a **process** crash loses it -- not just a power failure. Same shape with a delete: the vector stays in the published index with no row and no log row to remove it.

Neither needs `row_id` reuse, so neither is fixed by option 1. Both are fixed by option 3, and option 3 does not remove the need for option 1: `query()` scores keys in the engine and resolves them to records in a second step with nothing held in between (readers are deliberately not serialized against writers), so a reused id would still hand back a record that was never scored, wearing the score of the record that was.

So the fix is 1 **and** 3, which is what [#1469](https://github.com/MemMachine/MemMachine/pull/1469) now does, with a regression test per defect.

---

*Update filed by Claude Code (AI agent) on behalf of @edwinyyyu.*

edwinyyyu:
This branch carries #1589, the first half of #1468 as merged into `speedkick` (`658a5164c`), on `main`: the same change on the one-collection store of #1627 below. The tests build their store through the module's `_params` helper and open partitions by key, and the gated test engine implements `get_vectors`, which `main`'s engines have and `speedkick`'s no longer do. The ordering half is #1607, ported as #1673.

### Purpose of the change

`SQLiteVectorStore` keeps records in SQLite and vectors in a search engine keyed by the record's row id. The records table used a plain `INTEGER PRIMARY KEY`, which *is* SQLite's rowid, assigned as `max(rowid) + 1`: deleting the highest row freed its id for the very next insert, so a new record inherited the id an old one was being served under.

`query` scores keys in the search engine, then resolves them to rows in a second step, holding nothing in between, because readers are not made to wait on writers. A reused id lets a record that was never scored come back wearing the score of the record that was. Nothing about that result looks wrong from outside: the record exists, the score is in range, and no row or vector is left dangling to find the mix-up by.

`sqlite_autoincrement=True`, so ids are never reused. A stale engine key then matches no row and is dropped.

First half of #1468. The other half, concurrent writers reaching the engine out of commit order, is #1607, stacked on this.

### Tests

Both fail without the flag. `test_row_ids_are_never_reused` pins the id policy directly. `test_a_query_cannot_return_a_record_it_never_scored` gates the engine so a query parks between scoring and row lookup, deletes the scored record and inserts another in that window, and asserts the query returns nothing.

### Port

This branch carries this change as merged into `speedkick` (#1589, `658a5164c`), the one commit `f335891fc`; the branch's earlier commits are replaced.

### Stack

| # | PR | change |
|---|---|---|

| # | PR | change |
|---|---|---|
| [sqlite store fixes 2/7] | #1469 **(this PR)** | Never reuse a row id in SQLiteVectorStore (as merged into `speedkick`, #1589) |

This PR's own change is its one commit, `f335891fc`; the rest of its diff is the PRs under it. Stacked on #1460; #1672 is stacked on it.

### Verification

At every commit from [sqlite store fixes 1/7] (`d7a6042b2`) to the tip `545ec401e`, on 2026-09-21: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`); 2037 tests at this PR's head `f335891fc`. Those commits were verified before the re-order that moved #1460 into the stage, which left their trees unchanged (`git diff` against the verified commits is empty). The 22 commits below `d7a6042b2` (the scale-out stage and the independent PRs' copies in its history), whose trees the re-order changed only by the absence of #1460's files, are being re-verified as this is written; this paragraph is replaced with the result.

### DIFF
--- a/packages/server/server_tests/memmachine_server/common/vector_store/test_sqlite_vector_store.py
+++ b/packages/server/server_tests/memmachine_server/common/vector_store/test_sqlite_vector_store.py
@@ -1,5 +1,6 @@
 """Tests for SQLiteVectorStore."""
 
+import asyncio
 import math
 from datetime import UTC, datetime, timedelta, timezone
 from uuid import UUID, uuid4
@@ -33,6 +34,9 @@ from memmachine_server.common.vector_store.sqlite_vector_store import (
 )
 from memmachine_server.common.vector_store.vector_search_engine.usearch_engine import (
     USearchVectorSearchEngine,
+)
+from memmachine_server.common.vector_store.vector_search_engine.vector_search_engine import (
+    VectorSearchEngine,
 )
 
 NAMESPACE = "test_namespace"
@@ -1072,6 +1076,150 @@ class TestConcurrentAsync:
             await collection.upsert(records=more_records)
 
         await asyncio.gather(query_loop(), upsert_more())
+
+
+# ── row_id reuse (issue #1468) ──
+
+
+class _GatedRemoveEngine(VectorSearchEngine):
+    """Delegates to a real engine; the next remove() waits on `gate` first.
+
+    delete() already suspends at its engine remove() (an await point after
+    the SQL commit); the gate only widens that window so the interleaving
+    with a concurrent upsert() is deterministic.
+    """
+
+    def __init__(self, inner: VectorSearchEngine) -> None:
+        self.inner = inner
+        self.gate: asyncio.Event | None = None
+        self.gate_reached = asyncio.Event()
+
+    async def add(self, vectors):
+        await self.inner.add(vectors)
+
+    async def remove(self, keys):
+        if self.gate is not None:
+            gate, self.gate = self.gate, None
+            self.gate_reached.set()
+            await gate.wait()
+        await self.inner.remove(keys)
+
+    async def search(self, vectors, *, limit, allowed_keys=None):
+        return await self.inner.search(vectors, limit=limit, allowed_keys=allowed_keys)
+
+    async def get_vectors(self, keys):
+        return await self.inner.get_vectors(keys)
+
+    async def save(self, path):
+        await self.inner.save(path)
+
+    async def load(self, path):
+        await self.inner.load(path)
+
+
+class TestRowIdReuse:
+    """Regression tests for row_id reuse (issue #1468).
+
+    Without AUTOINCREMENT, SQLite assigns max(rowid) + 1, so deleting the
+    record holding the maximum row_id frees that id for the very next
+    insert. A delete() suspended between its SQL commit and its engine
+    remove() can then erase the vector of a concurrently upserted record
+    that reused the id.
+    """
+
+    async def _row_id_of(self, collection, record_uuid):
+        async with collection._create_session() as session:
+            return (
+                await session.execute(
+                    select(collection._records_table.c.row_id).where(
+                        collection._records_table.c.uuid == record_uuid
+                    )
+                )
+            ).scalar_one()
+
+    @pytest.mark.asyncio
+    async def test_row_ids_are_never_reused(self, collection):
+        """A new record must not be assigned a previously deleted row_id."""
+        record_a = _make_record(vector=_normalize([1.0, 0.0, 0.0]))
+        await collection.upsert(records=[record_a])
+        row_id_a = await self._row_id_of(collection, record_a.uuid)
+
+        await collection.delete(record_uuids=[record_a.uuid])
+
+        record_b = _make_record(vector=_normalize([0.0, 1.0, 0.0]))
+        await collection.upsert(records=[record_b])
+        row_id_b = await self._row_id_of(collection, record_b.uuid)
+
+        assert row_id_b != row_id_a
+
+    @pytest.mark.asyncio
+    async def test_concurrent_delete_and_upsert_keeps_upserted_vector(self, tmp_path):
+        """delete() racing upsert() must not lose the upserted record's vector.
+
+        Interleaving: delete(A) commits its SQL transaction and suspends at
+        its engine remove(); upsert(B) runs to completion in that window;
+        delete(A) resumes and removes A's row_id from the engine. B must
+        remain searchable afterwards.
+        """
+        db_path = tmp_path / "test.db"
+        engine = create_async_engine(f"sqlite+aiosqlite:///{db_path}")
+        gated_engines: list[_GatedRemoveEngine] = []
+
+        def factory(ndim, metric):
+            gated = _GatedRemoveEngine(
+                USearchVectorSearchEngine(num_dimensions=ndim, similarity_metric=metric)
+            )
+            gated_engines.append(gated)
+            return gated
+
+        store = SQLiteVectorStore(
+            SQLiteVectorStoreParams(
+                sqlalchemy_engine=engine,
+                vector_search_engine_factory=factory,
+            )
+        )
+        await store.startup()
+        try:
+            collection = await store.open_or_create_collection(
+                namespace=NAMESPACE,
+                name=NAME,
+                config=VectorStoreCollectionConfig(vector_dimensions=VECTOR_DIM),
+            )
+            (gated_engine,) = gated_engines
+
+            record_a = _make_record(vector=_normalize([1.0, 0.0, 0.0]))
+            await collection.upsert(records=[record_a])
+
+            # delete(A) commits its SQL transaction, then blocks at the
+            # gated engine remove(). Wait for it to consume the gate so
+            # upsert(B)'s own engine remove() passes through unblocked.
+            gate = asyncio.Event()
+            gated_engine.gate = gate
+            delete_task = asyncio.create_task(
+                collection.delete(record_uuids=[record_a.uuid])
+            )
+            await gated_engine.gate_reached.wait()
+
+            record_b = _make_record(vector=_normalize([0.0, 1.0, 0.0]))
+            await collection.upsert(records=[record_b])
+
+            gate.set()
+            await delete_task
+
+            fetched = await collection.get(record_uuids=[record_b.uuid])
+            assert len(fetched) == 1
+
+            results = await collection.query(
+                query_vectors=[_normalize([0.0, 1.0, 0.0])], limit=5
+            )
+            assert any(
+                match.record.uuid == record_b.uuid
+                for result in results
+                for match in result.matches
+            ), "upserted record lost from the search engine"
+        finally:
+            await store.shutdown()
+            await engine.dispose()
 
 
 # ── Crash recovery & pending operations ──
--- a/packages/server/src/memmachine_server/common/vector_store/sqlite_vector_store.py
+++ b/packages/server/src/memmachine_server/common/vector_store/sqlite_vector_store.py
@@ -974,6 +974,11 @@ class SQLiteVectorStore(VectorStore):
             Column("uuid", Uuid, nullable=False, unique=True),
             Column("properties", JSON, nullable=False, default=dict),
             extend_existing=True,
+            # AUTOINCREMENT so row_ids are never reused. Writes commit their
+            # SQL transaction before awaiting the search engine apply, so a
+            # reused row_id lets a suspended delete() erase the vector of a
+            # concurrently upserted record.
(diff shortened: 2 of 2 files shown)


edwinyyyu (commented):
The description is the pre-split PR's: it describes a per-collection write lock held across commit, apply, mark-applied and save, and three tests (`test_an_upsert_survives_a_delete_of_another_record`, `test_an_upsert_cannot_overtake_a_delete_of_the_same_uuid`, `test_a_save_cannot_trim_a_write_it_did_not_publish`) that are not on this branch; they are #1607's, ported as #1673. This branch is `sqlite_autoincrement=True` plus the two `TestRowIdReuse` tests, i.e. #1589, "First half of #1468". "Fixes #1468" would close the issue on merge with the ordering half still open; #1673 carries the same keyword.

PR #1475, opened 2026-07-13
Feat: Add Hybrid Search (Vector + FTS)

Jeoungahn:
### Purpose of the change

This PR introduces **Hybrid Search** capability to MemMachine, combining vector similarity search with keyword-based Full-Text Search (FTS) to significantly improve retrieval accuracy. The implementation addresses the fundamental limitation of pure vector search where exact keyword matches (error messages, model names, technical terms, identifiers) are often missed due to semantic embedding dilution.

This PR enables clients to opt-in to hybrid retrieval via a simple use_fts flag without changing query patterns.

### Description

**1. API Extension** - Added `use_fts: bool` parameter across the API stack to enable hybrid search mode. Default `false` maintains backward compatibility.

**2. FTS Index Setup** - FTS indexes are auto-created during ingest. The index name follows the pattern `fts_{sanitized_derivative_collection}_content` and is indexed on the Derivative node's content field. No manual index creation is required; the index is ready for use when hybrid search is enabled.

**3. FTS Search Implementation** - New `_search_fts()` method (`packages/server/src/memmachine_server/episodic_memory/long_term_memory/long_term_memory.py:363-460`) executes Neo4j Full-Text Search queries on Derivative nodes using `db.index.fulltext.queryNodes()`. It follows the `DERIVED_FROM` relation to retrieve Episode nodes, returning FTS scores and Episode objects with metadata extracted from Neo4j.

**4. Hybrid Search Orchestration** - New `_search_scored_hybrid()` method (`packages/server/src/memmachine_server/episodic_memory/long_term_memory/long_term_memory.py:295-349`) runs Vector Search and FTS in parallel via `asyncio.gather()` (no latency overhead). Results are merged with UID-based deduplication: Vector results (top-k, RRF-reranked) are preserved in order, and unique FTS results (top-10) are appended. Returns up to `top-k + 10` episodes.

**5. Query Escaping Utility** - Static method `_escape_lucene_query()` (`packages/server/src/memmachine_server/episodic_memory/long_term_memory/long_term_memory.py:351-361`) escapes Lucene special characters (`+ - = && || > < ! ( ) { } [ ] ^ " ~ * ? : \ /`) to prevent query syntax errors on user input containing special characters.

**Motivation:** Vector search excels at semantic similarity but struggles with exact keyword matches (error messages like `CUDA_ERROR_OUT_OF_MEMORY`, model names like `RTX-4090`, acronyms like `API/HTTP`, function/class names). FTS complements vector search by providing precise term matching.

**Dependencies:** No new external dependencies. Uses existing Neo4j FTS indexes (auto-created on ingest), asyncio, and vector search infrastructure.

### Summary of Changes

This PR adds Hybrid Search functionality by integrating Neo4j Full-Text Search (FTS) indexes with existing vector search infrastructure. The key changes include:

### Fixes/Closes

N/A (New feature)

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

**Test Environment:**
- **Benchmark**: LongMemEval_M (500 questions, 896 gold turns)
- **Dataset**: `longmemeval_m_cleaned.json`
- **Configuration**: Qwen3-Embedding-4B + Qwen3.5-397B-A17B-NVFP4

- [ ] Unit Test
- [ ] Integration Test
- [ ] End-to-end Test

**Usage Example:**

When data is ingested, FTS indexes are automatically created. To enable hybrid search, pass `use_fts=True` in the `QueryParam` when calling the search function:

```python
from memmachine_server.retrieval_agent.common.agent_api import QueryParam

# Hybrid search enabled
results = await do_query(
    QueryParam(
        query="CUDA_ERROR_OUT_OF_MEMORY",
        limit=50,
        memory=memory,
        use_fts=True,  # Enable Hybrid Search (Vector + FTS)
    )
)
```

| Method | top-k | Overall Recall | Improvement |
|--------|-------|----------------|-------------|
| Vector (baseline) | 50 | 0.8811 | - |
| Vector | 60 | 0.9006 | +1.95% |
| Vector | 70 | 0.8945 | +1.34% |
| Vector | 80 | 0.9010 | +1.99% |
| **Hybrid (Vector+FTS)** | **50** | **0.9108** | **+2.97%** |

**Key Findings:**
- Hybrid Search achieves **+2.97% improvement** over Vector-only (top-k=50)
- Outperforms Vector top-k=80 (0.9010) while fetching fewer total results
- **Single-session-user queries**: +7.8% improvement (largest gain)
- **Temporal-reasoning queries**: +3.9% improvement

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### Screenshots/Gifs

N/A (Backend feature - no UI changes)

### Further comments

**Known Limitations:**
1. Hybrid search currently only supported on declarative backend (Neo4j - VectorGraphStore), not event backend (VectorStore)
2. FTS returns limited metadata (timestamp/producer_id set to defaults)
3. FTS uses BM25 scores while vector uses cosine/euclidean; threshold semantics differ

### DIFF
--- a/packages/client/src/memmachine_client/memory.py
+++ b/packages/client/src/memmachine_client/memory.py
@@ -444,6 +444,7 @@ class Memory:
             filter=filter_str,
             set_metadata=set_metadata,
             types=[MemoryType.Episodic, MemoryType.Semantic],  # Search both types
+            use_fts=False,
         )
         v2_search_data = spec.model_dump(mode="json", exclude_none=True)
 
--- a/packages/client/src/memmachine_client/memory.py
+++ b/packages/client/src/memmachine_client/memory.py
@@ -36,6 +36,7 @@ from memmachine_common.api.spec import (
     DeleteSemanticTagSpec,
     DisableSemanticCategorySpec,
     EpisodicMemoryConfigEntry,
+    FtsMode,
     GetEpisodicMemoryConfigSpec,
     GetFeatureSpec,
     GetSemanticCategorySetIdsResponse,
@@ -370,6 +371,8 @@ class Memory:
         filter: str | None = None,  # noqa: A002 — matches API field name in `SearchMemoriesSpec.filter`
         set_metadata: dict[str, JsonValue] | None = None,
         agent_mode: bool = False,
+        use_fts: FtsMode = False,
+        append_n: int = 10,
     ) -> SearchResult:
         """
         Search for memories.
@@ -401,6 +404,10 @@ class Memory:
                     `filter_dict` filters, all filters are combined with AND.
             set_metadata: Optional metadata key-value pairs used to select semantic sets.
             agent_mode: Whether to enable top-level retrieval-agent orchestration.
+            use_fts: Full-Text Search (keyword) mode combined with Vector Search.
+                    False: vector only; True or 'rrf': RRF fusion; 'append':
+                    append FTS results after vector results.
+            append_n: Number of FTS results to append (only used when use_fts='append').
 
         Returns:
             SearchResult object containing search results from both episodic and semantic memory
@@ -444,7 +451,8 @@ class Memory:
             filter=filter_str,
             set_metadata=set_metadata,
             types=[MemoryType.Episodic, MemoryType.Semantic],  # Search both types
-            use_fts=False,
+            use_fts=use_fts,
+            append_n=append_n,
         )
         v2_search_data = spec.model_dump(mode="json", exclude_none=True)
 
--- a/packages/common/src/memmachine_common/api/spec.py
+++ b/packages/common/src/memmachine_common/api/spec.py
@@ -621,6 +621,14 @@ class SearchMemoriesSpec(_WithOrgAndProj):
         ),
     ]
 
+    # Flags for using Hybrid Search(Vector + Full-Text Search)
+    use_fts: Annotated[
+        bool,
+        Field(
+            default=False,
+            description="Use Full-Text Search (keyword search) with Vector Search",
+        ),
+    ]
 
 class DeleteMemoriesSpec(_WithOrgAndProj):
     """Specification model for deleting memories."""
--- a/packages/common/src/memmachine_common/api/spec.py
+++ b/packages/common/src/memmachine_common/api/spec.py
@@ -26,6 +26,19 @@ UTC = timezone.utc
 PropertyValue = bool | int | float | str | datetime
 """Type for stored property values (duplicated here to avoid server dependency)."""
 
+# FTS (Full-Text Search) hybrid search configuration.
+# use_fts modes:
+#   False           -> vector search only
+#   True / "rrf"    -> hybrid search fused via Reciprocal Rank Fusion (returns top_k)
+#   "append"        -> hybrid search that appends FTS results after vector results
+FtsMode = Literal["rrf", "append"] | bool
+
+# RRF score formula constant (1 / (RRF_K + rank)); NOT a result-count.
+RRF_K = 60
+
+# Default number of FTS results appended in "append" mode.
+FTS_APPEND_COUNT_DEFAULT = 10
+
 # Canonical type-name set accepted in `properties_schema`. Mirrors
 # `memmachine_server.common.data_types.PROPERTY_TYPE_NAME_TO_PROPERTY_TYPE` but
 # lives here so the API contract validates types at request time rather than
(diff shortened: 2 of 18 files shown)


malatewang:
We will retire the neo4j based episodic memory implementation. The new implementation is based on vector store and sql DB, the code is under https://github.com/MemMachine/MemMachine/tree/main/packages/server/src/memmachine_server/episodic_memory/event_memory. Can you also add the FTS to the new implementation?

edwinyyyu:
> We will retire the neo4j based episodic memory implementation. The new implementation is based on vector store and sql DB, the code is under https://github.com/MemMachine/MemMachine/tree/main/packages/server/src/memmachine_server/episodic_memory/event_memory. Can you also add the FTS to the new implementation?

I think a text index with the new system would require a new ABC or interface. It would not be appropriate to add text search functionality to the VectorStore/Collection API.

Jeoungahn:
Thank you for the thoughtful feedback(@malatewang @edwinyyyu).  
I agree that adding text search functionality directly to Event Memory (Vector Store + SQL DB) is not appropriate. While the existing Neo4j provided text search functionality that enabled integrated implementation with Vector Index, the new Event Memory architecture requires verification of equivalent text search support and may require significant architectural changes.
So, I would like to position the current Hybrid Search PR as a concept/reference contribution. I will leave the PR available for reference should you decide to pursue implementation in the future.

RonNiles:
First let's get CI to pass. Please review the attached gist, which covers most of the issues, and apply it if you find it satisfactory: https://gist.github.com/RonNiles/b299f1826d58a0f33ab91ca01a53b479

Jeoungahn:
> First let's get CI to pass. Please review the attached gist, which covers most of the issues, and apply it if you find it satisfactory: https://gist.github.com/RonNiles/b299f1826d58a0f33ab91ca01a53b479

Thanks @RonNiles. 
I've re-opened the PR and confirmed that your patch works as expected. I also rebased onto the latest main.
It seems we have two possible paths for the remaining issue:

1. Limit the current hybrid search implementation to Neo4j only. This keeps the PR focused, but FTS will not be supported on the Nebula backend.
→ For example, we could work around this with an isinstance check and cast() to bypass the type checker."
2. Build a backend-agnostic Hybrid search (FTS). This supports all backends consistently, but requires architectural refactoring beyond current scope.

For this PR, we could move forward with option 1 first, with option 2 as our final development target. Would you agree with this approach?

RonNiles:
> Thanks @RonNiles. I've re-opened the PR and confirmed that your patch works as expected. I also rebased onto the latest main. It seems we have two possible paths for the remaining issue:
> 
>     1. Limit the current hybrid search implementation to Neo4j only. This keeps the PR focused, but FTS will not be supported on the Nebula backend.
>        → For example, we could work around this with an isinstance check and cast() to bypass the type checker."
> 
>     2. Build a backend-agnostic Hybrid search (FTS). This supports all backends consistently, but requires architectural refactoring beyond current scope.
> 
> 
> For this PR, we could move forward with option 1 first, with option 2 as our final development target. Would you agree with this approach?

I think #1 is the best way because it limits the scope of this PR and the refactoring and expanded backend support required by #2 is best done in a separate PR. Also, we will not retire neo4j right away, and legacy support of neo4j should be possible into the future because many users monitor the neo4j database directly for things like determining whether ingestion is complete before beginning search, so we don't want to break those use cases.

Here is my wish list for the rest of the PR:

1. The API and python client search should expose **use_fts**. I think we should support the various flavors: **append-n**, with **n** default to 10  (best case as studies show but custom **n** can specified), and **rrf** for reciprocal rank fusion which is popular in the industry and works well with small k such as WikiHow benchmark.
2. FTS requires adding a new index to neo4j and currently it will happen on ingestion. It can cause a long pause at that time which may be unexpected. Also the user may want to do a hybrid search without having to add episodes for the sole purpose of triggering the new FTS index. I feel it is best to build the index when MemMachine initializes, and via a new "EnableFtsIndex" API call in case the user doesn't want to restart. Perhaps a good way is to add a key to  **config.yml** such as: **episodic_memory:long_term_memory:fts:enabled** and if this key is present, build the neo4j index at startup. This will also be a good place to put "fts_default: " flag so that if user has a client application that is unaware of the fts flag, and wants to run that client application with fts, the configured default will allow that enhancement without needing to modify the client calls.
3.  Add unit test coverage for the FTS cases.

Jeoungahn:
> > Thanks @RonNiles. I've re-opened the PR and confirmed that your patch works as expected. I also rebased onto the latest main. It seems we have two possible paths for the remaining issue:
> > ```
> > 1. Limit the current hybrid search implementation to Neo4j only. This keeps the PR focused, but FTS will not be supported on the Nebula backend.
> >    → For example, we could work around this with an isinstance check and cast() to bypass the type checker."
> > 
> > 2. Build a backend-agnostic Hybrid search (FTS). This supports all backends consistently, but requires architectural refactoring beyond current scope.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For this PR, we could move forward with option 1 first, with option 2 as our final development target. Would you agree with this approach?
> 
> I think #1 is the best way because it limits the scope of this PR and the refactoring and expanded backend support required by #2 is best done in a separate PR. Also, we will not retire neo4j right away, and legacy support of neo4j should be possible into the future because many users monitor the neo4j database directly for things like determining whether ingestion is complete before beginning search, so we don't want to break those use cases.
> 
> Here is my wish list for the rest of the PR:
> 
> 1. The API and python client search should expose **use_fts**. I think we should support the various flavors: **append-n**, with **n** default to 10  (best case as studies show but custom **n** can specified), and **rrf** for reciprocal rank fusion which is popular in the industry and works well with small k such as WikiHow benchmark.
> 2. FTS requires adding a new index to neo4j and currently it will happen on ingestion. It can cause a long pause at that time which may be unexpected. Also the user may want to do a hybrid search without having to add episodes for the sole purpose of triggering the new FTS index. I feel it is best to build the index when MemMachine initializes, and via a new "EnableFtsIndex" API call in case the user doesn't want to restart. Perhaps a good way is to add a key to  **config.yml** such as: **episodic_memory:long_term_memory:fts:enabled** and if this key is present, build the neo4j index at startup. This will also be a good place to put "fts_default: " flag so that if user has a client application that is unaware of the fts flag, and wants to run that client application with fts, the configured default will allow that enhancement without needing to modify the client calls.
> 3. Add unit test coverage for the FTS cases.

@RonNiles  I've implemented the features as you proposed, and completed the basic and E2E tests.
- Scoped Hybrid search to Neo4j only (other backends fall back to vector-only)
- Added RRF(Reciprocal Rank Fusion) fusion mode and user options for search result fusion (rrf / append modes, with configurable append count)
- Auto-create FTS index at session start, plus manual creation support for existing projects
- Added Hybrid Search unit tests

If there are no further comments, please go ahead and merge. @malatewang, @edwinyyyu

Issue #1002, opened 2026-01-29
[Bug]: LangGraph example demo raises AttributeError when adding memories with v2 APIs


sscargal:
### Describe the bug

The current LangGraph example in the repository does not work and fails during execution when adding memories. The error is raised in the MemMachine tools integration with LangGraph, specifically in the demonstration script using the v2 APIs.

### Steps to reproduce

Run the LangGraph integration demo as follows:

1. Set up the environment according to the documentation.
2. From within the `integrations/langgraph/` directory, execute:
   ```
   python demo.py
   ```
3. You will see the initialization succeed, but when adding memories, the following error is produced (this is repeated):

```
Failed to add memory
Traceback (most recent call last):
  File "/home/leoq/memtest/.venv/lib/python3.12/site-packages/memmachine/rest_client/memory.py", line 274, in add
    message = self._build_memory_message(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/leoq/memtest/.venv/lib/python3.12/site-packages/memmachine/rest_client/memory.py", line 196, in _build_memory_message
    combined_metadata["episode_type"] = episode_type.value
                                        ^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'value'
   ❌ Failed: Error adding memory
```

### Expected behavior

The demo should run without errors and successfully add memories using the v2 APIs with the LangGraph integration.

### Environment

- OS: Mac OS X
- MemMachine Version: v0.2.5
- Development language version: Python 3.12
- MemMachine LangGraph integration (demo.py) from the current repository main branch

### Additional context

The error suggests that `episode_type` is being passed as a `str`, but the code expects an Enum that has a `.value` attribute. Passing the string type results in an `AttributeError`. This breaks compatibility with the intended v2 APIs.

Please review the demo and supporting function calls to ensure correct data types are used with `episode_type` and that the LangGraph example works out-of-the-box with v2 APIs.

PR #1512, opened 2026-08-15
fix(llamaindex): pass valid EpisodeType enum instead of invalid "text" string

connectsudhindra-gif:
## Summary
- `MemMachineMemory.add()` in `integrations/llamaindex/mem_machine_memory.py` called `memory.add(episode_type="text", ...)`, passing a raw string where `Memory.add()` expects an `EpisodeType` enum.
- `Memory._build_memory_message()` calls `episode_type.value` internally, so this raised `AttributeError: 'str' object has no attribute 'value'` on every call to `add()`/`put()`.
- `"text"` was also never a valid `EpisodeType` value in the first place (the enum only defines `MESSAGE = "message"`), so simply string-normalizing it (as was done for the LangGraph integration in #1403) wouldn't have been enough here.
- Fix: pass `EpisodeType.MESSAGE` directly, matching the only valid episode type today.
- This is the same failure mode reported in #1002 for the LangGraph integration, which was already fixed by #1403 but never linked back to that issue.

## Test plan
- Added `integrations/llamaindex/test_mem_machine_memory.py` with two unit tests (mocking the backend `Memory`) that assert `episode_type` is forwarded as a real `EpisodeType.MESSAGE` enum via both `add()` and `put()`.
- Verified the new tests fail against the pre-fix code with the exact same shape of error class this bug produces, and pass after the fix.
- Ran `ruff check` and `ruff format --diff` on both changed files: clean.

### DIFF
--- a/integrations/llamaindex/mem_machine_memory.py
+++ b/integrations/llamaindex/mem_machine_memory.py
@@ -7,6 +7,7 @@ from llama_index.core.base.llms.types import ChatMessage, MessageRole
 from llama_index.core.memory import BaseMemory
 from llama_index.core.memory import Memory as LlamaIndexMemory
 from memmachine_client import MemMachineClient
+from memmachine_common.api import EpisodeType
 
 DEFAULT_INTRO_PREFERENCES = "Below are a set of relevant preferences retrieved from potentially several memory sources:"
 DEFAULT_OUTRO_PREFERENCES = "This is the end of the retrieved preferences."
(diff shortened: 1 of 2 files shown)


connectsudhindra-gif:
This is ready for review whenever a maintainer has time. Let me know if any changes are needed.

connectsudhindra-gif:
@ChristianKniep / @jgong  / @jealous pls review

PR #1513, opened 2026-08-15
fix(semantic-memory): stop unbounded LLM retry storm on persistent extraction failures

### LINKED ISSUE #1453: [Bug]: Excessive background semantic/profile memory LLM calls
### Describe the bug

MemMachine appears to trigger an unexpectedly large number of LLM calls for
semantic/profile memory ingestion after a relatively small amount of user
activity.

I used our internal product which integrates `Memchine` to ask around 40 questions over a small knowledge
set: 4 PDF files, each under 80 pages. I did not start any other jobs or run any
batch ingestion manually.

However, starting around midnight and continuing into the evening, the system
kept calling the LLM in the background and generated about $200 in model
costs.

A captured single LLM request shows that the call was not answering my PDF
question directly. It was a `MemMachine profile/semantic memory extraction cal`:

- system prompt: "Your job is to handle memory extraction for a memory system..."
- user prompt includes:
  - `<OLD_PROFILE>{}</OLD_PROFILE>`
  - `<HISTORY>...previous assistant answer about Apple's gross margin...</HISTORY>`

So the unexpected cost seems related to background semantic/profile memory
ingestion or repeated `add_memory` processing, not normal user-facing QA.

### Possible minimal reproduction path:

1. Enable both episodic and semantic memory.
2. Call `add_memory` repeatedly for the same user/project, passing full
   conversation context each time.
3. Wait for semantic ingestion to run.
4. Observe whether MemMachine repeatedly sends profile extraction prompts to the
   LLM even when there is no new user activity.

### Expected behavior

After the pending memory ingestion backlog is processed, MemMachine should stop
calling the LLM while idle.

LLM calls for profile/semantic memory extraction should be bounded and
predictable. For a small document set and around 40 user questions, background
memory processing should not continue for many hours or generate very high
model costs.

Ideally MemMachine should also provide safeguards such as:

- max background ingestion calls per project/user per time window
- deduplication/idempotency for already-ingested messages
- an option to disable automatic semantic/profile ingestion from MCP
- cost/concurrency limits for background ingestion

### Environment

MemMachine commit: af5a1a38d3b3439cc3579e684f79f7bb225ebd41

### Additional context

[one LLM call.txt](https://github.com/user-attachments/files/28978886/one.LLM.call.txt)

connectsudhindra-gif:
## Summary
Fixes #1453 (reported ~$200 in unexpected LLM cost from background semantic/profile ingestion running for many hours with no user activity).

Root cause, traced in `packages/server/src/memmachine_server/semantic_memory/semantic_ingestion.py`:
- The background ingestion loop (`semantic_memory.py::_background_ingestion_task`) polls every `feature_update_interval_sec` (default **2 seconds**) for any set with un-ingested messages.
- Per-message idempotency already exists via an `is_ingested` flag — once a message is successfully processed (or hits a non-retryable context-length error), it's marked ingested and won't be re-fetched.
- But when `llm_feature_update` failed with any *other* exception (a transient API error, a malformed/unparseable LLM response for that specific content, etc.), the code logged it and `continue`d **without marking the message ingested**.
- That message stays `is_ingested=False` forever, so the very next poll cycle (2 seconds later) fetches it again and calls the LLM again — forever, with no backoff, since the exception is swallowed locally and never reaches the outer loop's error-backoff handling.
- A single message that fails deterministically for the same input (which is plausible — malformed JSON output, a specific content pattern the LLM mishandles, etc.) therefore becomes an unbounded retry storm: thousands of LLM calls over hours from a single stuck message, matching the reporter's symptom exactly.

## Fix
- Added `_try_feature_update`, mirroring the existing `_try_consolidate` bounded-retry pattern already used for the consolidation path in the same file.
- A generic failure is now retried up to `_MAX_FEATURE_UPDATE_ATTEMPTS` (2, matching `_MAX_CONSOLIDATION_ATTEMPTS`) times.
- Once retries are exhausted, the message is marked ingested — same treatment already given to non-retryable context-length errors — so a persistently failing message stops being re-fetched instead of retrying every 2 seconds forever.
- `debug_fail_loudly` still raises instead of silently giving up, unchanged behavior for that flag.
- This does not address the other (larger, separately-scoped) asks in the issue — cost/concurrency budgets, a queue/status API, content-hash dedup for resubmitted history. Those are legitimate follow-ups but a bigger feature than this bug fix.

## Test plan
- Added `test_process_single_set_gives_up_on_persistent_generic_errors`: simulates a message whose feature update fails every time, asserts the LLM is called only `_MAX_FEATURE_UPDATE_ATTEMPTS` times (not once, not forever), that the message ends up marked ingested, and that a second `_process_single_set` call (simulating the next poll cycle) makes zero further LLM calls.
- Added `test_process_single_set_retries_and_recovers_from_transient_error`: fails once then succeeds, asserts the retry recovers and the feature is correctly applied.
- Verified both new tests **fail against the pre-fix code** (confirming they catch the real bug: 1 call instead of the expected bounded retries, no backoff before infinite reprocessing) and pass after the fix.
- Ran the full `semantic_memory` test suite (`uv run pytest packages/server/server_tests/memmachine_server/semantic_memory/`): 349 passed, 2 skipped (pre-existing, unrelated), integration tests requiring Postgres/Neo4j deselected as usual.
- `ruff check` and `ruff format --diff` on both changed files: clean.

### DIFF
--- a/packages/server/server_tests/memmachine_server/semantic_memory/test_semantic_ingestion.py
+++ b/packages/server/server_tests/memmachine_server/semantic_memory/test_semantic_ingestion.py
@@ -277,6 +277,130 @@ async def test_process_single_set_marks_context_length_errors_as_ingested(
 
 
 @pytest.mark.asyncio
+async def test_process_single_set_gives_up_on_persistent_generic_errors(
+    ingestion_service: IngestionService,
+    semantic_storage: SemanticStorage,
+    episode_storage: EpisodeStorage,
+    monkeypatch,
+):
+    """A message whose feature update fails deterministically (e.g. a
+    malformed LLM response for that specific content) must not be re-fetched
+    and re-sent to the LLM forever. Regression test for the runaway
+    background-ingestion cost reported in #1453: previously a generic
+    (non-context-length) failure left the message permanently un-ingested,
+    so every subsequent background poll cycle re-sent it to the LLM with no
+    backoff and no way to ever stop."""
+    message_id = await add_history(
+        episode_storage,
+        content="message that always fails feature extraction",
+    )
+    await semantic_storage.add_history_to_set(
+        set_id="user-poison-message",
+        history_id=message_id,
+    )
+
+    llm_feature_update_mock = AsyncMock(
+        side_effect=ValueError("structured-output validation failed")
+    )
+    monkeypatch.setattr(
+        "memmachine_server.semantic_memory.semantic_ingestion.llm_feature_update",
+        llm_feature_update_mock,
+    )
+
+    await ingestion_service._process_single_set("user-poison-message")
+
+    # Retried a bounded number of times, not once and not forever.
+    from memmachine_server.semantic_memory.semantic_ingestion import (
+        _MAX_FEATURE_UPDATE_ATTEMPTS,
+    )
+
+    assert llm_feature_update_mock.await_count == _MAX_FEATURE_UPDATE_ATTEMPTS
+
+    # Given up and marked ingested so it stops showing up as pending work.
+    assert (
+        await _collect(
+            semantic_storage.get_history_messages(
+                set_ids=["user-poison-message"],
+                is_ingested=False,
+            )
+        )
+        == []
+    )
+    assert await _collect(
+        semantic_storage.get_history_messages(
+            set_ids=["user-poison-message"],
+            is_ingested=True,
+        )
+    ) == [message_id]
+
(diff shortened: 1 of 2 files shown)


connectsudhindra-gif:
This is ready for review whenever a maintainer has time. Let me know if any changes are needed.

connectsudhindra-gif:
@ChristianKniep / @jgong  / @jealous pls review

PR #1523, opened 2026-08-25
fix(metrics): make Prometheus metrics correct with multiple workers, and wire the trackers up

wanghy73:
Two fixes to metrics that are wrong in ways nothing reports. Both were found while
trying to attribute request latency on a capacity run and discovering the numbers
could not be trusted.

## 1. Metrics are not aggregated across uvicorn workers

`/metrics` called bare `generate_latest()`. With `MEMMACHINE_WORKERS > 1` uvicorn
forks a worker per process and each keeps its own registry, so a scrape returned
whichever worker happened to answer — about 1/N of the traffic, a different 1/N
each time. Consecutive scrapes look like counter resets, and every rate, histogram
quantile and calls-per-request figure derived from them is wrong in an unbounded
direction.

Build a fresh registry per scrape and attach `MultiProcessCollector` when
`PROMETHEUS_MULTIPROC_DIR` is set. The single-worker path keeps the default
registry, so nothing changes for the default deployment.

**Measured**, on an 8-tenant capacity run over one rate ladder:

| workers | counter resets before | after |
|---|---|---|
| 4 | 70 | **0** |
| 8 | 87 | **0** |

Five consecutive scrapes of a loaded server returned identical values where they
previously diverged.

This needs `PROMETHEUS_MULTIPROC_DIR` set and a writable directory mounted at it.
The chart change that does so is separate.

## 2. OperationTracker silently discards timings when given no factory

`OperationTracker` accepts `metrics_factory=None` and then throws away every
timing it takes — no error, no warning, no series. A component that is fully
instrumented but never handed a factory is indistinguishable from one that was
never instrumented at all, and the only way to notice is to go looking for a
metric that should exist and find nothing.

That is not hypothetical: the Neo4j store, the episode store and the session
store each shipped instrumented and unwired, which is why database latency
appeared to be unmeasurable. Every call was being timed and discarded. This wires
the factory through the resource manager, the database manager and the event
backend's params.

### On the tests

`test_metrics_factory_wiring.py` asserts at the **call sites**, not on the
components. An earlier version tested that each store honours a factory it is
given — which passes whether or not anything passes one, and still passed with
the fix reverted. The tests here fail when the wiring is removed.

## Verification

- All 9 changed Python files parse; the new test module was written against the
  running server.
- Deployed to a test cluster as `memmachine/memmachine:instrumented-mp` and
  exercised with 1, 4 and 8 workers under load.

The two changes are kept as separate commits so either can be reverted alone.

### DIFF
--- a/packages/server/src/memmachine_server/common/episode_store/episode_sqlalchemy_store.py
+++ b/packages/server/src/memmachine_server/common/episode_store/episode_sqlalchemy_store.py
@@ -56,6 +56,11 @@ from memmachine_server.common.filter.sql_filter_util import (
     FieldEncoding,
     compile_sql_filter,
 )
+from memmachine_server.common.metrics_factory import (
+    MetricsFactory,
+    OperationTracker,
+    timed,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -140,12 +145,20 @@ class Episode(BaseEpisodeStore):
 class SqlAlchemyEpisodeStore(EpisodeStorage):
     """SQLAlchemy episode store implementation."""
 
-    def __init__(self, engine: AsyncEngine) -> None:
+    def __init__(
+        self,
+        engine: AsyncEngine,
+        metrics_factory: MetricsFactory | None = None,
+    ) -> None:
         """Initialize the store with an async SQLAlchemy engine."""
         self._engine: AsyncEngine = engine
         self._session_factory = async_sessionmaker(
             self._engine,
             expire_on_commit=False,
+        )
+        self._tracker = OperationTracker(
+            metrics_factory,
+            prefix="episode_store_sqlalchemy",
         )
 
     def _create_session(self) -> AsyncSession:
@@ -179,6 +192,7 @@ class SqlAlchemyEpisodeStore(EpisodeStorage):
             await session.commit()
 
     @validate_call
+    @timed("add_episodes")
     async def add_episodes(
         self,
         session_key: str,
@@ -241,6 +255,7 @@ class SqlAlchemyEpisodeStore(EpisodeStorage):
 
         return episode.to_typed_model() if episode else None
 
+    @timed("get_episodes")
     async def get_episodes(
         self,
         episode_ids: Iterable[EpisodeIdT],
@@ -342,6 +357,7 @@ class SqlAlchemyEpisodeStore(EpisodeStorage):
 
         raise ValueError(f"Unknown filter field: {field!r}")
 
+    @timed("get_episode_messages")
     async def get_episode_messages(
         self,
         *,
@@ -376,6 +392,7 @@ class SqlAlchemyEpisodeStore(EpisodeStorage):
 
         return [h.to_typed_model() for h in episode_messages]
 
+    @timed("get_episode_messages_count")
     async def get_episode_messages_count(
         self,
         *,
@@ -398,6 +415,7 @@ class SqlAlchemyEpisodeStore(EpisodeStorage):
 
         return int(n_messages)
 
+    @timed("get_episode_ids")
     async def get_episode_ids(
         self,
         *,
@@ -420,6 +438,7 @@ class SqlAlchemyEpisodeStore(EpisodeStorage):
         return [EpisodeIdT(row) for row in rows]
 
     @validate_call
+    @timed("delete_episodes")
     async def delete_episodes(self, episode_ids: list[EpisodeIdT]) -> None:
         try:
             int_episode_ids = TypeAdapter(list[int]).validate_python(episode_ids)
--- a/packages/server/src/memmachine_server/server/api_v2/router.py
+++ b/packages/server/src/memmachine_server/server/api_v2/router.py
@@ -1,6 +1,7 @@
 """API v2 router for MemMachine project and memory management endpoints."""
 
 import logging
+import os
 from typing import Annotated
 
 from fastapi import APIRouter, Depends, FastAPI, Response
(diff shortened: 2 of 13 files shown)


edwinyyyu:
Please pass ruff at least. ty is fixed here: https://github.com/MemMachine/MemMachine/pull/1519

wanghy73:
Thanks — ruff passes now. Three things were wrong: `typing.Any` in the `_timed` wrappers I added (ANN401), import sorting, and I'd missed that the job also runs `ruff format --check`.

Fixed with `Concatenate` plus a `Protocol`-bound TypeVar rather than a `noqa`, so the annotations stay meaningful and `ty` stays at its 17 pre-existing diagnostics — the naive `ParamSpec` version pushed it to 19, since an unbound type variable can't carry `_tracker`.

Confirmed those 17 are all the `nebulagraph_python.client has no member NebulaAsyncClient` import, reproduced identically on `upstream/speedkick`, so #1519 should clear them. `speedkick` and `main` are the same commit right now, so it applies to this PR's base directly once merged.

Both commits pass `ruff check` and `ruff format --check` independently, so either can still be reverted on its own.

packages/server/src/memmachine_server/server/api_v2/router.py
```diff
@@ -1074,7 +1080,17 @@ async def configure_long_term_memory(
 )
 async def metrics() -> Response:
     """Expose Prometheus metrics endpoint."""
-    return Response(generate_latest(), media_type=CONTENT_TYPE_LATEST)
+    # With MEMMACHINE_WORKERS > 1 each uvicorn worker is a separate process with
+    # its own registry, so the default generate_latest() returns whichever worker
+    # answered the scrape - about 1/N of the traffic, a different 1/N each time.
+    # Consecutive scrapes then look like counter resets and every rate or
+    # histogram derived from them is wrong.
+    multiproc_dir = os.environ.get("PROMETHEUS_MULTIPROC_DIR")
```
malatewang, line 1088: Who should take care of the directory setup?

packages/server/src/memmachine_server/common/episode_store/episode_sqlalchemy_store.py
```diff
@@ -56,15 +57,57 @@
...
+    OperationTracker,
+)
 
 logger = logging.getLogger(__name__)
 
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+
+class _HasTracker(Protocol):
```
malatewang: The same code duplicates. We can make it common

packages/server/src/memmachine_server/server/api_v2/router.py
wanghy73: Good catch — the answer was "nobody", which is a real gap. Fixed in `725e1de`.

The `/metrics` handler is also the wrong place to look for it: `prometheus_client` mmaps a file per worker when the *first metric is registered*, long before any scrape arrives. Creating the directory here would already be too late.

So `main()` now calls `_prepare_multiproc_dir()` before any metric exists. It creates the directory and clears stale `*.db` files — without that, a previous run's dead workers get summed into the live counters on every scrape. It's best-effort: a read-only or pre-seeded mount logs a warning rather than refusing to start, since failing to boot over metrics would be the wrong trade.

Verified it creates missing nested directories, clears stale files, and warns without raising on an unwritable path.

packages/server/src/memmachine_server/common/episode_store/episode_sqlalchemy_store.py
wanghy73: Agreed, and it was worse than the diff showed — fixed in `00a7e81`.

A later push to satisfy `ruff`'s ANN401 replaced the `Any` annotations with a `Protocol`-bound `TypeVar`, and I copied that into both files too. So by the time you looked it was the decorator *plus* a `Protocol` and three `TypeVar`s duplicated — about 30 identical lines in two places.

Now extracted to `common/metrics_factory/timed.py`, next to `OperationTracker`, and exported as `timed`. Both stores import the one copy; the only thing that differed between them was the docstring.

packages/server/src/memmachine_server/server/app.py
```diff
@@ -144,6 +144,38 @@ def start_http() -> None:
...
+    surfaces as an unrelated error deep in a worker.
+
+    Stale files are cleared for the same reason: they belong to workers from a
+    previous run of this process, and MultiProcessCollector would otherwise add
+    their dead counters to the live ones on every scrape.
+
+    Deliberately best-effort. A deployment may mount the directory read-only or
+    pre-seed it, and refusing to start over metrics would be the wrong trade.
+    """
+    path = os.environ.get("PROMETHEUS_MULTIPROC_DIR")
```
malatewang, line 172: It is better to check the worker number. If the worker number is bigger than 1, setup multiproc dir automatically.

wanghy73: Agreed — done in `b89ee3e`.

The worker count is the thing that decides whether aggregation is needed, so it now drives the setup: above 1 with nothing configured, the directory defaults to `<tempdir>/memmachine-prometheus-multiproc`. An explicit `PROMETHEUS_MULTIPROC_DIR` still wins, which is how two servers on one host keep their metrics apart.

Two details worth flagging:

- If the chosen directory can't be created, the variable is removed again. `prometheus_client` raises in every worker if it can't open the directory it's pointed at, so leaving behind a path only this code picked would turn an unwritable temp dir into a server that refuses to start — the opposite of the best-effort behaviour the rest of the function promises. An operator's own unwritable path still just warns.
- Worker-count parsing moved into `_worker_count()`, shared with `start_http()`, so the two can't disagree about the number.

Setting the env var in `main()` reaches the workers because uvicorn spawns them via `multiprocessing.get_context("spawn")` (`uvicorn/_subprocess.py`) — each child is a fresh interpreter that inherits `os.environ` and imports `prometheus_client` after we've set it.

13 tests added in `test_multiproc_dir.py`: the default, the explicit override at both 1 and 4 workers, nested creation, stale `*.db` clearing, and both failure paths.

malatewang (approved):

PR #1526, opened 2026-08-26
Feat: Add vector store collection registry with SQLAlchemy implementation (collection registry stack 1/5)

### LINKED ISSUE #1524: [Feat]: Cross-process collection registry for vector store metadata

Horizontal scalability: multiple MemMachine server processes should be able to work with the **same** collections and partitions on one backend (e.g. a Qdrant cluster). Strict sharding — each collection managed by exactly one process — is what the `VectorStore` contract requires today ("must be managed by at most one process at a time", `vector_store.py`), and it works; the restriction itself is the limitation. What prevents lifting it is collection *metadata* management, not the data path: each vector store keeps its own ad-hoc catalog of logical collections (Qdrant and Milvus in dummy-vector `__registry` collections inside the backend itself, the SQLite stores in bespoke tables), and for backends with no conditional writes, transactions, or unique constraints there is no way to make create / open-or-create / delete atomic — they are read-check-write sequences guarded only by per-process `asyncio` locks.

A generic, cross-process-safe config registry primitive that collection metadata can be moved onto:

- `ConfigRegistryManager` — provides named registries over one backing store; the isolation guarantee (different names -> disjoint entries; same name -> same entries, across processes) is part of the contract; `open_registry(name, config_adapter)` idempotently prepares storage and returns a scoped handle.
- `ConfigRegistry[ConfigT]` — the handle: `create` (atomic, fails on existing key), `get`, `get_or_create` (atomic; returns the stored config and whether this call created it; never compares configs — equality policy stays with the consumer), idempotent `delete`. No name parameter on any operation, so a consumer holding a handle structurally cannot reach other registries.
- A SQLAlchemy implementation (PostgreSQL + SQLite) using one table per registry with the primary key as the compare-and-set arbiter.

First consumer: QdrantVectorStore's collection registry (tracked separately). Later candidates: MilvusVectorStore's `__registry`, the SQLite vector stores' `_CollectionRow` tables, the segment store's `PartitionRow`, and the session manager's `SessionConfig` (currently a race-prone read-then-insert).

- Keeping the catalog inside each vector database: Qdrant/Milvus offer no CAS primitive, so the races are structurally unfixable there.
- Bespoke per-backend SQL tables: duplicates the subtle insert-conflict handling once per backend instead of once.
- A generic key->config registry (an earlier draft of this proposal): speculative surface with one consumer kind; the collection-specific API says the same things in domain terms and can be generalized later if a non-collection consumer materializes.
- Distributed locks/leases: heavier operationally, and crash-recovery semantics are worse than a unique-constraint insert that is atomic by construction.

### Additional context

An implementation PR follows immediately; further PRs switch QdrantVectorStore and MilvusVectorStore onto it.

---

edwinyyyu:
### Purpose of the change

Horizontal scalability: multiple MemMachine server processes should be able to work with the *same* collections and partitions on one backend (#1524). Strict sharding — each collection managed by exactly one process, as the `VectorStore` contract requires today — has always worked; the restriction itself is the limitation, and what prevents lifting it is collection *metadata* management, not the data path. This PR introduces the enabling primitive — a cross-process-safe **collection registry** (`CollectionRegistry` ABC + a SQLAlchemy implementation, in the `vector_store` package). Today several backends keep collection/partition metadata in ad-hoc per-backend bookkeeping whose create/open-or-create sequences are non-atomic read-check-write, guarded only by per-process asyncio locks. A registry backed by a store with real unique constraints makes entry creation an atomic compare-and-set across processes.

A follow-up PR switches QdrantVectorStore's internal Qdrant-based `__registry` bookkeeping to this registry.

### Description

**Contract** (`common/vector_store/collection_registry.py`): a `CollectionRegistry` is a durable catalog of one vector store's logical collections — a mapping from (namespace, name) to an immutable `CollectionRegistryEntry`. The registry is deliberately **specific to vector store collections** rather than a generic key->config registry: the realistic reuse axis (other vector store backends — Milvus follows in this stack) is within the collection domain, a specific API can be widened later as an additive refactor whereas narrowing a shipped generic ABC is breaking, and speaking the domain removes key/adapter plumbing from the API entirely. Operations:

- `register(namespace, name, entry)` — atomic; raises `CollectionAlreadyRegisteredError` on an existing collection regardless of stored-entry equality
- `get(namespace, name)` — stored entry or `None`
- `get_or_register(namespace, name, entry)` — atomic; returns `(stored_entry, registered)`; never compares entries — config-equality policy belongs to the store, which translates conflicts into its own domain errors
- `deregister(namespace, name)` — idempotent
- `startup()` — idempotently prepares storage

**The entry** (`CollectionRegistryEntry` = `{config, native_collection_name, partition_key}`) stores resolved identity, one format shared by all backends: `config` is the ABC-level `VectorStoreCollectionConfig`, the native name is pinned at registration (so config-serialization changes can never silently repoint collections at new empty native collections), and the partition key carries a per-registration generation (so records written through handles held across a deregistration stay invisible and are never resurrected). Evolution policy: extend by adding optional fields with defaults — no version bump, no migration; backend-specific needs use the same mechanism, not per-backend entry types.

There is deliberately no update operation: entries are immutable once registered (the config hash *is* the native collection identity). Adding one later is purely additive.

**Format evolution without standing version machinery.** There is deliberately no stored format version and no version check. The evolution policy (add-optional-only, defaults reproducing prior behavior — stated on the entry model and guarded by a canned v1-row regression test) is what makes that safe: old rows classify exactly through defaults, so a future breaking change can introduce an entry format version field (default 1) *at the moment it is first needed* — every pre-existing row is correctly version 1 by construction — and ship its migration as a windowed migrate-then-deploy step. The accepted residual: an old binary started against migrated data fails with validation errors at first read rather than being refused at boot. (An earlier draft carried a declarations table with a startup version check and a redeclare primitive; it was removed because it insured against exactly the scenario the evolution policy already insures, at the cost of real API and documentation surface.)

**SQLAlchemy implementation** (`sqlalchemy_collection_registry.py`): each registry owns a dedicated table `collection_registry_<name>` (`key` VARCHAR(255) PK, `entry` JSON, JSONB on PostgreSQL), so registries sharing a database are isolated at the table level and a store can only reach its own registry (idempotent `create_all` in `startup()`, matching the segment store's schema-management convention). Storage keys are `f"{namespace}/{name}"` — an implementation detail; "/" is outside the identifier charset, so distinct pairs can never collide ("__" would be ambiguous). The primary key is the concurrency arbiter:

- `register` = plain INSERT; `IntegrityError` -> `CollectionAlreadyRegisteredError` (the pattern `SQLAlchemySegmentStore.create_partition` already uses)
- `get_or_register` = SELECT fast path, then native `INSERT .. ON CONFLICT DO NOTHING`; on conflict re-SELECT the winner — no exception-driven control flow
- `deregister` = single DELETE

Entries round-trip through pydantic (`dump_python(mode="json")` / `validate_python`); `get_or_register` returns the serialization round trip of the input, so the registering store sees exactly what every later reader gets. Supported dialects: PostgreSQL and SQLite (the two relational providers the server offers). The 32-byte registry-name limit keeps `collection_registry_<name>` within PostgreSQL's 63-byte identifier limit. No new dependencies.

**Adopters in this stack:** Qdrant (next PR) and Milvus (final PR). The SQLite vector stores keep their own `_CollectionRow` bookkeeping — their state is process-local anyway, so the registry buys them nothing today.

**Design doc**: `design/collection_registry.md` — contributor-facing design notes live in the root `design/` directory, outside the deployed docs site (`docs/`) and outside packaged sources.

### Fixes/Closes

Fixes #1524

### Type of change

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test

`server_tests/memmachine_server/common/vector_store/test_sqlalchemy_collection_registry.py`, parametrized over SQLite (file) and PostgreSQL (testcontainer, `integration`-marked): startup idempotence (twice, and a second instance on the same engine), entry round trips, duplicate registration (same and different entry), get_or_register canonicalization and no-comparison semantics, idempotent deregister, registry isolation on a shared engine, key-ambiguity regression (("a__b","c") vs ("a","b__c")), identifier/name/engine validation, the canned v1-row readability guard, and concurrency tests (two instances over one engine, `asyncio.gather`): concurrent register -> exactly one `AlreadyRegistered`; concurrent get_or_register with mismatched entries -> both callers observe the same stored entry, exactly one `registered=True`.

**Test Results:** 45 passed (SQLite + PostgreSQL) locally; `ruff check`, `ruff format --check`, and `ty check packages` clean (no new diagnostics).

### Checklist

### Further comments

Design notes for the reviewer:

- **Collection-specific, not generic**: see the scope note above; `design/collection_registry.md` records the rejected generic draft and the reasoning, so the narrowing reads as deliberate rather than as a missed abstraction.
- **Flat JSON, not typed columns**: every access is get-by-key, pydantic validates at the boundary, and typed columns would turn entry evolution into DDL migrations in a codebase whose schema management is `create_all`-on-startup.
- **Table-per-registry, not one shared table**: simpler keys, database-level isolation, whole-registry cleanup is a `DROP TABLE`.
- **No registry management surface** (list/delete registries): registries are constructed by wiring code from configuration; the database catalog already lists `collection_registry_*` tables for observability, and management operations are exactly the power stores must not have.

### DIFF
--- a/design/README.md
+++ b/design/README.md
@@ -1,0 +1,13 @@
+# Design notes
+
+Contributor-facing notes on the reasoning behind a design: the problem it
+solves, the alternatives weighed, and the constraints that decided it.
+They record why the code is shaped the way it is, which the code itself
+and its docstrings cannot say.
+
+These documents are written for people working on the server. They are
+not part of the deployed documentation site (`docs/`) and are not
+packaged into distributions; user-facing documentation belongs in
+`docs/`.
+
+One document per design, named after what it covers.
--- a/design/collection_registry.md
+++ b/design/collection_registry.md
@@ -1,0 +1,372 @@
+# Vector store collection registry
+
+Contributor-facing design notes for
+`memmachine_server.common.vector_store.collection_registry`.
+
+## Problem
+
+Horizontal scalability: multiple server processes should be able to work with
+the same collections on one backend. Strict sharding (each collection managed
+by exactly one process) has always worked; the blocker to lifting it is
+collection *metadata* management. Backends without conditional writes,
+transactions, or unique constraints (Qdrant, Milvus) cannot make
+create / open / delete atomic, so their ad-hoc catalogs are read-check-write
+sequences guarded by per-process locks.
+
+The collection registry is the enabling primitive: a durable catalog mapping
+(namespace, name) to an immutable `CollectionRegistryEntry`, with
+registration atomic across processes via a real compare-and-set. A vector
+store holds a registry dedicated to it and cannot reach any other registry
+through it.
+
+## Scope: collections, not a generic registry
+
+The registry is deliberately specific to vector store collections. An
+earlier draft generalized it to a key -> config registry with pluggable
+types; that was narrowed because it had exactly one consumer kind, and the
+realistic reuse axis (other vector store backends) is *within* the
+collection domain -- Milvus's needs are the same shape. The compatibility
+asymmetry decides the default: widening a specific API later (extracting a
+generic ABC when a non-collection consumer actually materializes) is an
+additive refactor; narrowing a shipped generic ABC is a breaking change.
+Being collection-specific also lets the API speak the domain --
+`register(namespace, name, entry)` rather than key/adapter plumbing -- and
+turns the key encoding into an implementation detail.
+
+## The entry
+
+`CollectionRegistryEntry` = `{config, native_collection_name,
+partition_key}`. One entry format is shared by all vector store backends:
+`config` is the ABC-level `VectorStoreCollectionConfig` every backend already
+takes, and the two identity fields are the shared mechanism (config-hashed
+shared native collections, per-collection partition keys) that Qdrant and
+Milvus both use.
+
+- `native_collection_name` is pinned at registration instead of being
+  re-derived from `sha256(config)` on every open: any change to config
+  serialization -- even an added optional field -- would otherwise re-hash
+  every stored config and silently repoint existing collections at new,
+  empty native collections.
+- `partition_key` is generation-scoped (minted per registration): records
+  written through handles held across a deregistration land under the dead
+  generation -- invisible, never resurrected by a re-registration, no locks
+  and no per-write round trips.
+
+Evolution policy -- covering the whole stored document: the entry envelope
+and every model nested in it, `VectorStoreCollectionConfig` included, since
+the config is serialized inside the entry's JSON and read back through the
+same validation. Extend by adding optional fields with defaults -- no
+version bump, no migration, old rows validate under the new model. The rule
+has a semantic half: a new field's default must reproduce the behavior that
+existed before the field, because stored rows predating it are read through
+that default. A change whose default cannot mean the old behavior is not
+additive -- it is a breaking change and takes the version bump path.
+Backend-specific needs are expressed the same way, not with per-backend
+entry types.
+
+Terminology: "generation" is reserved for collection incarnations (the
+partition-key uuid); evolution of the entry or config models is a format
+"version". Content-hash native naming has the side effect of never sharing
+native collections across config-model versions (any model change drifts
+the hash) -- conservative quarantine at the cost of gradual fragmentation.
+
+## Portability across vector store providers
+
+Two identity fields locate a collection, and that arity is structural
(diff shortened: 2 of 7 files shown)


edwinyyyu:

PR #1527, opened 2026-08-26
Feat: Back QdrantVectorStore collection metadata with SQL collection registry (collection registry stack 2/5)

### LINKED ISSUE #1525: [Feat]: Redesign QdrantVectorStore collection registry for multi-process collection management

Horizontal scalability: multiple MemMachine server processes should be able to work with the **same** logical collections on one Qdrant cluster. Strict sharding — each collection managed by exactly one process, names partitioned across processes — is what the `VectorStore` contract requires today ("must be managed by at most one process at a time"), and it works. The restriction itself is the limitation, and it cannot be lifted with the current registry design.

Collection metadata lives in per-namespace `<ns>__registry` Qdrant collections (dummy-vector points keyed `uuid5(name)`). Qdrant has no conditional writes, transactions, or unique constraints, so `create_collection` / `open_or_create_collection` / `delete_collection` are non-atomic read-check-write sequences whose only guard is a per-process `asyncio` lock. The moment two processes manage the same name, the contract's guarantees dissolve:

1. `VectorStoreCollectionAlreadyExistsError` stops firing — two concurrent `create_collection` calls both pass the absence check and last-writer-wins the registry point.
2. Two `open_or_create_collection` calls with **different** configs each create a *different* native collection (native names are `sha256(config)` while registry points are keyed by name), one silently wins the registry point, and records written through the losing handle become permanently unreachable — `VectorStoreCollectionConfigMismatchError` never fires.

So this is not a bug under the current contract; it is the reason the contract has to forbid sharing, and lifting it requires redesigning where the registry lives.

Move QdrantVectorStore's collection registry out of Qdrant onto the cross-process config registry primitive (#1524), whose unique-constraint insert is a real compare-and-set:

- The registry insert becomes the atomic commit point of collection creation (native-collection creation first, registration last, so a crash leaves only an empty config-shared native collection that the next same-config creation adopts).
- `AlreadyExists` / `ConfigMismatch` guarantees then hold across processes sharing the same registry database; the per-process lock remains only as in-process serialization.
- Native collection naming is unchanged.

- Keeping the registry inside Qdrant: structurally unfixable — there is no CAS primitive to build atomic create on.
- Distributed locks/leases around lifecycle operations: operationally heavier, and crash-recovery semantics are worse than an insert that is atomic by construction.

### Additional context

Scope: collection *lifecycle/metadata* management. Registry entries store resolved identity (native collection name, plus a generation-scoped partition key), which also makes `delete_collection` safe against handles held in other processes: writes through a held handle land under the dead generation — invisible, never resurrected by a re-creation. Read-after-write visibility tuning on the data path remains a separate concern.

Companion redesign pieces: strict create/open lifecycle (open-or-create removal) and a declared per-instance concurrency scope (PROCESS/MACHINE/CLUSTER) replacing the ABC's blanket one-process-per-collection sentence, so the capability is expressible through the contract.

---

edwinyyyu:
### Purpose of the change

Horizontal scalability: let multiple MemMachine server processes work with the *same* logical collections on one Qdrant cluster (#1525). Strict sharding of collection names across processes — what the `VectorStore` contract requires today — has always worked; this PR is the redesign that lifts the restriction for collection lifecycle management, by moving QdrantVectorStore's collection metadata out of Qdrant onto the SQL-backed config registry from #1526.

**Stacked on #1526**; only the last commit is new to this PR.

Qdrant offers no conditional writes, transactions, or unique constraints, so the current `<ns>__registry` bookkeeping (dummy-vector points keyed `uuid5(name)`) makes `create_collection` / `open_or_create_collection` / `delete_collection` non-atomic read-check-write sequences, guarded only by a per-process asyncio lock. The moment two processes manage the same name:

- `VectorStoreCollectionAlreadyExistsError` stops firing — two concurrent creates both pass the absence check and last-writer-wins the registry point.
- Worse, two `open_or_create_collection` calls with **different** configs each create a *different* native collection (native names are `sha256(config)` while registry points are keyed by name), one silently wins the registry, and records written through the loser's handle become permanently unreachable. `VectorStoreCollectionConfigMismatchError` never fires.

With the registry's unique-constraint CAS as the commit point, both guarantees hold across processes sharing the same registry database, and generationed partition keys make `delete_collection` safe against handles held in other processes. Scope: lifecycle/metadata management — data-path read-after-write visibility tuning remains a separate concern.

### Description

**`qdrant_vector_store.py`**: all `__registry` machinery is deleted (`_REGISTRY_*` constants, uuid5 point ids, `_ensure_namespace_registry_collection`, `_get_registry_entry`, `_parse_entry`, `_register_collection`, `_is_not_found_error`, `registry_replication_factor`). `QdrantVectorStoreParams` gains a required `registry: CollectionRegistry`.

**Registry entries store resolved identity, not just config** (`CollectionRegistryEntry` from #1526): `config`, `native_collection_name`, and a generation-scoped `partition_key` (minted at creation; also the shard key in distributed mode).

- *Native name pinned at creation*: handles are built from the stored name rather than re-deriving `sha256(config)` on every open, so a later change to the config model's serialization (any added field re-hashes every config) cannot silently repoint existing collections at new empty native collections. The naming scheme for new collections is untouched.
- *Generationed partition keys make deletion safe against held handles*: `delete_collection` removes that generation's data and the entry; a handle held across the deletion keeps writing into the dead generation — invisible to every reader, storage-only garbage — and a re-creation mints a fresh generation, so stale records are never resurrected. No locks and no per-write round trips; the residual cost is bounded garbage under dead generations (a future GC can sweep partition keys absent from the registry).

Lifecycle ordering, chosen for crash windows:

- **Create: native-first, register-last.** The registry insert is the atomic commit point. A crash after native creation leaves only an empty native collection, which is shared by config and adopted by the next creation with the same config. (Claim-registry-first would be worse: `open_collection` builds handles straight from the entry, so a crash between claim and native create would hand out handles to a nonexistent native collection.) Invariant: registry entry implies native collection exists.
- **Delete: data-first, unregister-last** (unchanged ordering). A crash in between leaves a registered-but-empty collection (documented on `delete_collection`); retrying the delete completes it.
- **Create-race losers leave an empty native collection** (created before the CAS) and, in distributed mode, an unused shard key. Cleanup is deliberately not attempted — native collections are config-shared, so nothing can safely decide no other logical collection references one — and the residue is bounded by the number of distinct configs attempted.

`_name_locks` stays, with its comment updated: it now only serializes lifecycle operations within a process (preserving today's single-process open-vs-delete interleaving semantics and avoiding redundant idempotent round trips); cross-process correctness comes from the registry CAS.

**Wiring** (`database_manager.py`): `async_get_qdrant_client` builds and starts a `SQLAlchemyCollectionRegistry` named `qdrant_<conf name>` on the engine named by the new `QdrantConf.registry_database`, passing it into the store — QdrantVectorStore holds a registry dedicated to it and cannot reach any other. One registry per Qdrant conf entry, so two Qdrant instances sharing one registry database get separate tables. Misconfiguration (unknown database name, or a conf entry name that cannot form a registry table name) raises `QdrantConfigurationError` with the fix spelled out.

**Configuration** (`database_conf.py`, samples, docs): `QdrantConf.registry_database` (required) names a configured relational database (`postgres` or `sqlite` entry); `registry_replication_factor` is removed — it existed to give the Qdrant-hosted registry read-your-writes, which is moot now. Sample configs and the configuration docs gained the new field; the installation wizard points Qdrant at its existing SQLite entry.

**Design doc**: `design/collection_registry_backends.md`.

### Breaking changes

- `QdrantConf.registry_database` is required: existing Qdrant users add one line of YAML (single-node setups can point at a `sqlite` relational entry). There is deliberately no per-host auto-SQLite fallback — divergent per-host registries would look consistent while reproducing exactly the races this PR removes. Stale `registry_replication_factor` YAML keys are silently ignored (`extra="ignore"`).
- Existing `<ns>__registry` Qdrant collections are no longer read, and there is no data migration. This is mostly self-healing: native collection names are unchanged, so the first `open_or_create_collection` with an unchanged config re-registers it and all data is reachable again. Caveat: paths that re-create with a *currently derived* schema (e.g. the episodic event backend) will orphan old partitions if that schema drifted since original creation — the old registry would have preserved the original config. Stale `__registry` collections can be deleted manually.

### Fixes/Closes

Fixes #1525

### Type of change

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test

- Existing QdrantVectorStore suite unchanged in behavior, now parametrized with a registry over the full client matrix: in-memory (unit) plus HTTP, gRPC, and distributed-cluster containers (`integration`).
- New `TestRegistryIntegration`: no `__registry`-suffixed collections exist in Qdrant after creation; two stores sharing one registry+client see each other's collections (create raises `AlreadyExists` cross-store, delete in one is observed by the other); concurrent `open_or_create_collection` with mismatched configs yields exactly one winner and one `VectorStoreCollectionConfigMismatchError`; `("a__b", "c")` and `("a", "b__c")` are distinct collections (key-ambiguity regression); re-creation mints a fresh generation over the same native collection; records upserted through a handle held across a deletion are unreachable from the re-created collection (resurrection regression).
- `test_database_manager.py`: registry built on the named engine and passed into the store; unknown `registry_database` and un-usable conf entry names raise `QdrantConfigurationError` and close the client.
- `test_database_conf.py`: field swap, required-field validation.

**Test Results:** full server suite 1901 passed locally; targeted integration run (qdrant containers + PostgreSQL registry) 452 passed; `ruff check`, `ruff format --check`, `ty check packages` clean (no new diagnostics).

### Checklist

### Further comments

The same swap for `MilvusVectorStore` lands later in this stack.

### DIFF
--- a/design/README.md
+++ b/design/README.md
@@ -1,0 +1,13 @@
+# Design notes
+
+Contributor-facing notes on the reasoning behind a design: the problem it
+solves, the alternatives weighed, and the constraints that decided it.
+They record why the code is shaped the way it is, which the code itself
+and its docstrings cannot say.
+
+These documents are written for people working on the server. They are
+not part of the deployed documentation site (`docs/`) and are not
+packaged into distributions; user-facing documentation belongs in
+`docs/`.
+
+One document per design, named after what it covers.
--- a/design/collection_registry.md
+++ b/design/collection_registry.md
@@ -1,0 +1,372 @@
+# Vector store collection registry
+
+Contributor-facing design notes for
+`memmachine_server.common.vector_store.collection_registry`.
+
+## Problem
+
+Horizontal scalability: multiple server processes should be able to work with
+the same collections on one backend. Strict sharding (each collection managed
+by exactly one process) has always worked; the blocker to lifting it is
+collection *metadata* management. Backends without conditional writes,
+transactions, or unique constraints (Qdrant, Milvus) cannot make
+create / open / delete atomic, so their ad-hoc catalogs are read-check-write
+sequences guarded by per-process locks.
+
+The collection registry is the enabling primitive: a durable catalog mapping
+(namespace, name) to an immutable `CollectionRegistryEntry`, with
+registration atomic across processes via a real compare-and-set. A vector
+store holds a registry dedicated to it and cannot reach any other registry
+through it.
+
+## Scope: collections, not a generic registry
+
+The registry is deliberately specific to vector store collections. An
+earlier draft generalized it to a key -> config registry with pluggable
+types; that was narrowed because it had exactly one consumer kind, and the
+realistic reuse axis (other vector store backends) is *within* the
+collection domain -- Milvus's needs are the same shape. The compatibility
+asymmetry decides the default: widening a specific API later (extracting a
+generic ABC when a non-collection consumer actually materializes) is an
+additive refactor; narrowing a shipped generic ABC is a breaking change.
+Being collection-specific also lets the API speak the domain --
+`register(namespace, name, entry)` rather than key/adapter plumbing -- and
+turns the key encoding into an implementation detail.
+
+## The entry
+
+`CollectionRegistryEntry` = `{config, native_collection_name,
+partition_key}`. One entry format is shared by all vector store backends:
+`config` is the ABC-level `VectorStoreCollectionConfig` every backend already
+takes, and the two identity fields are the shared mechanism (config-hashed
+shared native collections, per-collection partition keys) that Qdrant and
+Milvus both use.
+
+- `native_collection_name` is pinned at registration instead of being
+  re-derived from `sha256(config)` on every open: any change to config
+  serialization -- even an added optional field -- would otherwise re-hash
+  every stored config and silently repoint existing collections at new,
+  empty native collections.
+- `partition_key` is generation-scoped (minted per registration): records
+  written through handles held across a deregistration land under the dead
+  generation -- invisible, never resurrected by a re-registration, no locks
+  and no per-write round trips.
+
+Evolution policy -- covering the whole stored document: the entry envelope
+and every model nested in it, `VectorStoreCollectionConfig` included, since
+the config is serialized inside the entry's JSON and read back through the
+same validation. Extend by adding optional fields with defaults -- no
+version bump, no migration, old rows validate under the new model. The rule
+has a semantic half: a new field's default must reproduce the behavior that
+existed before the field, because stored rows predating it are read through
+that default. A change whose default cannot mean the old behavior is not
+additive -- it is a breaking change and takes the version bump path.
+Backend-specific needs are expressed the same way, not with per-backend
+entry types.
+
+Terminology: "generation" is reserved for collection incarnations (the
+partition-key uuid); evolution of the entry or config models is a format
+"version". Content-hash native naming has the side effect of never sharing
+native collections across config-model versions (any model change drifts
+the hash) -- conservative quarantine at the cost of gradual fragmentation.
+
+## Portability across vector store providers
+
+Two identity fields locate a collection, and that arity is structural
+rather than a Qdrant/Milvus accident. The pair is the *container* -- the
+unit a provider creates, deletes, caps, and pins fixed physical
+attributes to (dimensionality, metric, index structures) -- and a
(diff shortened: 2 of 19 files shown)


edwinyyyu:

PR #1531, opened 2026-08-26
Feat: Declare concurrency scope and data-plane contracts on storage ABCs (collection registry stack 4/5)

edwinyyyu:
### Purpose of the change

Make the multi-process capability from #1526/#1527 reachable through the API contract. After those PRs, QdrantVectorStore supports multiple processes managing the same collections — but the `VectorStore` ABC still said a collection "must be managed by at most one process at a time", so no consumer was *permitted* to rely on it. This PR replaces that blanket sentence with a declared, per-instance concurrency scope, and states the data-plane contracts that multi-process consumers need to know (#1525).

**Stacked on the open-or-create removal PR**; only the last commit is new to this PR.

### Description

**`ConcurrencyScope`** (`common/data_types.py`): `PROCESS < MACHINE < CLUSTER` — the widest deployment boundary within which concurrent instances of a component may safely manage the same resources. An `IntEnum`, ordered by breadth, so the effective scope of a composed system is the `min` of its parts' scopes.

**Declared on the ABCs** — `VectorStore`, `SegmentStore`, and `CollectionRegistry` gain an abstract `concurrency_scope` property, and the one-process-per-collection/partition sentences are rewritten to defer to it: concurrent management of the same collection/partition is safe within the declared scope; beyond it, consumers shard names, as before.

**Declared by the implementations:**

| Component | Scope | Why |
|---|---|---|
| `QdrantVectorStore` | `min(CLUSTER, registry.concurrency_scope)` | Qdrant is shareable across machines; the collection registry governs |
| `SQLAlchemyCollectionRegistry` | `CLUSTER` on PostgreSQL, `MACHINE` on file-backed SQLite | the CAS is as wide as the database's reach |
| `SQLAlchemySegmentStore` | `CLUSTER` on PostgreSQL, `MACHINE` on SQLite | already safe cross-process via unique-constraint create and row locks |
| `MilvusVectorStore` | `PROCESS` | collection bookkeeping guarded only by in-process locks (widened by the Milvus registry PR later in this stack) |
| `SQLiteVectorStore` | `PROCESS` | search-engine state lives in process memory |
| `SQLiteVecVectorStore` | `PROCESS` | check-then-write bookkeeping within one process |

So a Qdrant store backed by a PostgreSQL registry declares `CLUSTER`, and one backed by a shared SQLite file declares `MACHINE` — the deployment can introspect what it is allowed to do instead of relying on out-of-band knowledge.

**Data-plane contracts** stated on `VectorStoreCollection` (documentation only; both are pre-existing properties that multi-process deployment makes worth stating):

- *Visibility*: a returned write is durably accepted but not guaranteed visible to a subsequent query, from this instance or any other — consumers must not build on read-your-writes.
- *Authority*: stored properties and property filters operate on the collection's own copy of a record, with no freshness guarantee relative to any external source of truth; consumers governed by an external authority (e.g. SQL as the record of access-determining attributes) must re-validate results against it.

**Design doc**: `design/concurrency_scopes.md`.

### Fixes/Closes

Related to #1525 (contract half of the redesign).

### Type of change

(Adding an abstract property is breaking for out-of-tree `VectorStore`/`SegmentStore`/`CollectionRegistry` implementations, which must now declare a scope.)

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test

Scope assertions per implementation: qdrant with a SQLite registry declares `MACHINE` and with a PostgreSQL registry (integration) declares `CLUSTER`; the SQLAlchemy segment store and config registry assert dialect-based scopes across the sqlite/PG fixture matrix; Milvus and both SQLite stores assert `PROCESS`.

**Test Results:** 1897 passed (default run); 585 passed (targeted integration run); `ruff check`, `ruff format --check`, `ty check packages` clean (no new diagnostics).

### Checklist

### Further comments

Milvus's `PROCESS` declaration is honest at this point in the stack; the final PR moves Milvus onto the collection registry and widens it to `min(CLUSTER, registry scope)`.

### DIFF
--- a/design/concurrency_scopes.md
+++ b/design/concurrency_scopes.md
@@ -1,0 +1,104 @@
+# Concurrency scopes and data-plane contracts
+
+Contributor-facing design notes for `ConcurrencyScope` and the visibility and
+authority contracts on `VectorStoreCollection`.
+
+## The problem
+
+The `VectorStore` contract said a logical collection "must be managed by at
+most one process at a time". That is true of some implementations and not
+others, and it is about to stop being true of more of them: the moment one
+backend genuinely supports multi-process management, a blanket sentence is
+either false for it or forbids what it supports -- capability the contract
+cannot express is capability no consumer may use.
+
+## Declared scope
+
+`ConcurrencyScope` (`common/data_types.py`) is an `IntEnum`,
+`PROCESS < MACHINE < CLUSTER`: the widest deployment boundary within which
+concurrent instances of a component may safely manage the same resources.
+Ordered by breadth, so the effective scope of a composed system is the `min`
+of its parts' scopes.
+
+`VectorStore` and `SegmentStore` declare an abstract
+`concurrency_scope` property, and the blanket sentences defer to it:
+concurrent management of the same collection/partition is safe within the
+declared scope; beyond it, at most one instance manages a resource and the
+consumer shards names, as before.
+
+Declarations:
+
+| Component | Scope | Why |
+|---|---|---|
+| `SQLAlchemySegmentStore` | `CLUSTER` on PostgreSQL, `MACHINE` on SQLite | cross-process safe via unique-constraint create and row locks |
+| `QdrantVectorStore` | `PROCESS` | bookkeeping guarded only by in-process locks |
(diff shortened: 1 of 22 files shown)


edwinyyyu:
Reference for #1579 but don't take this implementation exactly. Closing.

PR #1533, opened 2026-08-27
Feat: Back MilvusVectorStore collection metadata with SQL collection registry (collection registry stack 5/5)

edwinyyyu:
### Purpose of the change

Move MilvusVectorStore's collection metadata onto the SQL-backed collection registry from #1526, the same swap #1527 made for Qdrant. Milvus kept the same bookkeeping shape (per-namespace `memmachine_<ns>__registry` Milvus collections with dummy-vector rows) with the same non-atomic read-check-write lifecycle guarded only by per-process locks — and the same cross-process races (#1525). Keeping a bespoke in-Milvus registry was self-containment that was no longer worth its hacks once the shared primitive existed.

**Stacked on the concurrency-scope PR (4/5)**; only the last commit is new to this PR.

### Description

**`milvus_vector_store.py`**: all `__registry` machinery is deleted (`_REGISTRY_*` constants, the dummy-vector registry collections, `_ensure_namespace_registry_collection`, `_get_registry_entry`, the hand-rolled `_parse_entry`, `_register_collection`, `_is_not_found_error`). `MilvusVectorStoreParams` gains a required `registry: CollectionRegistry`. The lifecycle follows the Qdrant design exactly:

- Create: native-first, register-last — the registry insert is the atomic commit point; a crash or lost race leaves only an empty config-shared native collection, adopted by the next same-config creation.
- Delete: data-first, deregister-last; records written through handles held across the deletion land under the dead generation's partition key — invisible, never resurrected by a re-creation.
- `AlreadyExists` now holds across processes sharing the registry database; `_name_locks` remain as in-process serialization only.

Milvus-specific notes: native naming is unchanged (`memmachine_{ns}__{sha256(config)}`); there is no shard-key machinery — the generationed partition key is only the partition key field value; the native schema's partition-key VARCHAR is resized for generationed keys, and native primary ids (`{partition_key}:{uuid}`) now embed the generation, keeping ids unique across generations.

**Concurrency scope**: `MilvusVectorStore.concurrency_scope` widens from `PROCESS` to `min(CLUSTER, registry scope)`. Milvus's default `Session` consistency (process B may read stale data relative to process A's writes) sits within the `VectorStoreCollection` visibility contract — no read-your-writes is promised — but deployments wanting stronger data-plane visibility should set `consistency_level` accordingly.

**Configuration**: `MilvusConf.registry_database` (required) names a configured relational database entry; the wiring builds the registry `milvus_<conf name>` via a `_build_collection_registry` helper now shared with Qdrant. Sample configs, docs, and the installation wizard updated.

**Design doc**: `design/collection_registry_backends.md` (the Qdrant doc, extended and renamed).

### Breaking changes

- `MilvusConf.registry_database` is required: existing Milvus users add one line of YAML. Stale `memmachine_<ns>__registry` Milvus collections are no longer read; as with Qdrant, this is mostly self-healing (native names unchanged, re-creation with an unchanged config re-registers the same native collection), with the same derived-schema drift caveat, and stale registry collections can be dropped manually.

### Fixes/Closes

Related to #1525.

### Type of change

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test

Milvus suite (Milvus Lite) runs with a registry over SQLite; the Milvus-side registry white-box test is removed with the machinery it tested; the scope assertion updates to governed-by-registry (`MACHINE` over the SQLite test registry). DatabaseManager and configuration tests updated for `registry_database` and the shared registry-builder helper.

**Test Results:** full server suite 1899 passed; targeted vector-store + resource-manager integration run 555 passed; `ruff check`, `ruff format --check`, `ty check packages` clean (no new diagnostics).

### Checklist

### Further comments

With Qdrant and Milvus both on the registry, the remaining `PROCESS`-scoped stores are the SQLite-backed ones, whose limitation is in-process search-engine state rather than metadata — the registry deliberately does not chase that.

### DIFF
--- a/design/README.md
+++ b/design/README.md
@@ -1,0 +1,13 @@
+# Design notes
+
+Contributor-facing notes on the reasoning behind a design: the problem it
+solves, the alternatives weighed, and the constraints that decided it.
+They record why the code is shaped the way it is, which the code itself
+and its docstrings cannot say.
+
+These documents are written for people working on the server. They are
+not part of the deployed documentation site (`docs/`) and are not
+packaged into distributions; user-facing documentation belongs in
+`docs/`.
+
+One document per design, named after what it covers.
--- a/design/collection_registry.md
+++ b/design/collection_registry.md
@@ -315,6 +315,13 @@ privileged. `shutdown` is a no-op in the SQLAlchemy implementation
 because the engine is shared and externally owned: the registry does not
 close what it did not open.
 
+**`concurrency_scope` is declared by the registry** rather than derived
+by each store, because the registry is what actually bounds atomicity: a
+store reports `min(its own ceiling, its registry's scope)`, so a
+cluster-capable backend sitting behind a file-backed SQLite registry
+correctly declares itself machine-scoped. One answer per backing store,
+not one per consumer. See `concurrency_scopes.md`.
+
 ## Storage (SQLAlchemy implementation)
 
 - Table per registry (`collection_registry_<name>`: `key` VARCHAR(255) PK,
--- a/design/collection_registry.md
+++ b/design/collection_registry.md
@@ -1,0 +1,372 @@
+# Vector store collection registry
+
+Contributor-facing design notes for
+`memmachine_server.common.vector_store.collection_registry`.
+
+## Problem
+
+Horizontal scalability: multiple server processes should be able to work with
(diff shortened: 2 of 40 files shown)


edwinyyyu:

PR #1537, opened 2026-08-28
Fix: Incarnation-scope SQLite vector store collection resources

### LINKED ISSUE #1536: Handles held across delete and recreate resurrect records in both SQLite vector stores
### What happened

A `VectorStoreCollection` handle stays usable after its collection is deleted. If a collection with the same (namespace, name) is then created, writes through the stale handle reach the **new** collection. Both SQLite-backed vector stores resurrect records this way; Qdrant and Milvus do not.

Everything below reproduces **inside a single process** — no concurrency across processes is involved.

Measured by: create collection -> open handle -> delete collection -> create again with the same name and config -> write through the stale handle -> query a freshly opened handle.

| Store | stale write | record reaches the recreated collection |
| --- | --- | --- |
| `QdrantVectorStore` | succeeds | no |
| `MilvusVectorStore` | succeeds | no |
| `SQLiteVecVectorStore` | succeeds | **yes, immediately** |
| `SQLiteVectorStore` | succeeds | **yes, after the next index save** |

### `SQLiteVecVectorStore`

Records and vector table names derive from (namespace, name) only (`_collection_prefix`). Delete drops those tables; create rebuilds them under the same names. The stale handle holds `records_table` and `vector_table_name`, so its writes land in the new collection's tables and are returned by queries on a freshly opened handle straight away.

### `SQLiteVectorStore` — worse, because it rewrites the live index

`SQLiteVectorStoreCollection` captures `index_path` (name-derived) and `search_engine` at open time, and `_maybe_save_index` writes *that* engine to *that* path. So a handle to a deleted collection can overwrite the index file belonging to its replacement:

```
index after recreate + legitimate write C: (268 bytes, sha 376ae8b8a2a4)
index after write through the STALE handle: (424 bytes, sha 41d3bffdce74)
stale handle rewrote the live index file:  True
after restart, collection contains: B (stale write) = True, C (legitimate) = True
```

The stale write's row also lands in the recreated collection's records table, and its `_PendingOperationRow` is keyed by the same (namespace, name), so the new collection's own pending operations count toward the stale handle's `save_threshold`. Before a save the record is present in SQL but missing from the index — the two disagree; after a save and restart it is fully queryable.

### Why the declared concurrency scope does not settle this

Both stores declare `ConcurrencyScope.PROCESS`, and that scope is the one place where the *application* could be the guarantor: inside a single process it can see every handle it holds and every delete it issues, so a lenient reading -- "a handle outliving its collection is the caller's problem" -- is at least coherent there. Above PROCESS it is not coherent at all. Process A holds a handle, process B deletes and recreates, and A has no way to learn of it; only the store can enforce the invariant. Any implementation declaring MACHINE or CLUSTER must therefore provide it outright.

That asymmetry argues for a uniform guarantee rather than a scope-conditional one:

- A scope-conditional contract is a trap. Consumers write against the lenient reading and break when they move to a wider-scoped backend, or when a backend's scope widens under them -- which is exactly what #1527 and #1533 do to Qdrant and Milvus.
- The discipline it delegates is real work: tracking handle lifetime across tasks, in every consumer, indefinitely, to avoid a failure whose symptom is silent data corruption.
- It is cheap for a store to make structurally impossible, and the registry-backed stores already do.

Note that the reproductions above are *sequential* -- no interleaving is needed to trigger them, so this is a handle-lifetime defect rather than a concurrency defect as such. Interleaving only removes the caller's chance to notice: with one task holding a handle while another deletes and recreates, nothing short of the application serialising its own lifecycle operations avoids it. At that point the declared PROCESS scope is not something the store delivers on its own either.

### Proposed solution

Generation-scope the native resources, in the stores themselves rather than by taking a registry dependency.

**Why not inject `CollectionRegistry`.** The registry exists to give a backend a transactional metadata authority it does not otherwise have: Qdrant and Milvus cannot make create atomic by themselves. The SQLite stores already *are* that authority. `_CollectionRow` lives in the same SQLAlchemy engine as the per-collection tables, and `create_collection` performs the DDL and the metadata write in one transaction, so a crash cannot leave a half-created collection. Injecting a registry would split one transactional authority into two, reintroduce the native-first/register-last crash window that the registry-backed stores accept only because they have no alternative, and add a required `registry_database` setting to what is currently the zero-configuration embedded option. The mechanism should be shared as a design, not as a dependency.

**The change.**

1. `_CollectionRow` gains a `generation` column, minted per creation.
2. `_collection_prefix` incorporates the generation, so a recreated collection gets new table names. Everything else currently keyed by (namespace, name) follows: in `SQLiteVectorStore` the index path, the `_search_engines` cache key, and `_PendingOperationRow`'s key; in `SQLiteVecVectorStore` the records and vector table names and the `{partition_key}:{uuid}` primary ids.
3. A stale handle then addresses dropped tables and fails loudly rather than silently landing in the replacement. That is permitted by the contract and is more useful than the silent invisibility the registry-backed stores offer, which is the most their shared native containers allow.
4. `_maybe_save_index` already opens a session to count pending operations; read the current generation in the same query and skip the save when it does not match the handle's, so a stale handle cannot leave an orphan index file either.

**Migration.** Default `generation` to the empty string, and have the prefix append it only when non-empty. Pre-existing collections keep their table names and index paths and keep working; collections created after the change get a generation, and a legacy collection acquires one when it is next deleted and recreated. No renames, no data movement, no startup migration step.

**Two adjacent defects worth fixing in the same pass.**

- `_PendingOperationRow`s are replayed on startup "applied or not", so the orphaned row left by a stale write is re-applied into the replacement collection's engine. This is why the stale record survived a restart in the transcript above, and generation-keying the rows fixes it.
- `SQLiteVecVectorStore.create_collection` reads, checks, then inserts, even though `_CollectionRow`'s primary key is already (namespace, name) and could arbitrate directly. Mapping `IntegrityError` onto `VectorStoreCollectionAlreadyExistsError` is the pattern `SQLAlchemyCollectionRegistry.register` uses, and would let the store deliver the process-scoped bookkeeping it declares.

Severity note for `SQLiteVectorStore`: `_CollectionRow.index_saved` marks the on-disk index as part of the durable contract once set, so a clobbered index is treated as an error rather than silently rebuilt.

Rough cost: ~6 hours -- 1h for the generation column and prefix in both stores, 1.5h for `SQLiteVectorStore`'s index path, engine cache key, pending-operation key and save guard, 1h for `SQLiteVecVectorStore` plus the create compare-and-set, 1.5h for a shared conformance test across all four stores, 1h for the legacy-generation path and its tests.

### Also needed

State the invariant in the `VectorStoreCollection` contract, unconditionally rather than per concurrency scope, so the fix above is conformance rather than a new behavior and a new backend cannot land without it.

### Notes

Pre-existing on `main`. The open collection-registry PRs (#1526, #1527, #1530, #1531, #1533) leave the mechanism untouched — they only remove `open_or_create_collection` from these two stores and add a `concurrency_scope` property. Same store and a related class of defect as #1468.

---

edwinyyyu:
### Purpose of the change

Both SQLite-backed vector stores let a collection handle reach the collection that replaces it. Native resource names derive from (namespace, name) alone, so deleting a collection and creating one with the same name rebuilds resources under the same names, and a handle opened before the deletion keeps addressing them.

Measured before this change, single-process, no concurrency involved:

| Store | write through a stale handle | reaches the recreated collection |
| --- | --- | --- |
| `SQLiteVecVectorStore` | succeeds | yes, immediately |
| `SQLiteVectorStore` | succeeds | yes, after the next index save |
| `QdrantVectorStore` | succeeds | no |
| `MilvusVectorStore` | succeeds | no |

`SQLiteVectorStore` is the worse case. `SQLiteVectorStoreCollection` captures `index_path` and `search_engine` at open time and `_maybe_save_index` writes that engine to that path, so a handle to a deleted collection rewrites the *live* index file of its replacement:

```
index after recreate + legitimate write C: (268 bytes, sha 376ae8b8a2a4)
index after write through the STALE handle: (424 bytes, sha 41d3bffdce74)
stale handle rewrote the live index file:  True
after restart, collection contains: B (stale write) = True, C (legitimate) = True
```

The stale write's row also lands in the replacement's records table, and its `_PendingOperationRow` carries the same (namespace, name) — and those rows are replayed on startup "applied or not", which is how the record survives a restart. `_save_collection_index` was likewise unscoped, so a stale handle could delete the replacement's applied pending operations and flip its `index_saved`, which marks the on-disk index as part of the durable contract.

### Description

An `incarnation` is minted per creation, stored on `_CollectionRow`, and made part of everything that was previously keyed by (namespace, name):

- `_collection_prefix`, and so the records table and (for sqlite-vec) the vec0 virtual table
- the on-disk index path and the `_search_engines` cache key
- `_PendingOperationRow.incarnation`, the save-threshold count, the "mark applied" updates, and the startup replay, which now skips operations whose incarnation is not the collection's current one
- `_save_collection_index`, so its pending-operation delete and `index_saved` update touch only the incarnation that saved

A stale handle therefore addresses dropped tables and fails loudly rather than writing into the replacement.

This deliberately does not take a `CollectionRegistry` dependency. The registry exists to give a backend a transactional metadata authority it does not have; these stores already are one — `_CollectionRow` lives in the same engine as the per-collection tables and `create_collection` does the DDL and the metadata write in one transaction. Delegating would split one authority into two and reintroduce a crash window that the registry-backed stores accept only because they have no alternative.

### Breaking changes

Existing databases are not migrated. Collections created by an earlier version keep resource names this version does not address, so their data is not reachable after upgrade. That is deliberate: migration is deferred along with the wider question of where the server runs DDL, and older data stays with older server versions for now.

### Fixes/Closes

Fixes #1536.

### Type of change

### How Has This Been Tested?

- [x] Unit Test

Seven new tests across the two stores: a stale handle cannot write to a recreated collection; a stale handle does not clobber the recreated collection's index file, verified by bytes before and after and by reopening the database without a clean shutdown (a clean shutdown rewrites the index from the live engine and hides the damage); a database predating the column is migrated on startup and stays usable; and the unsuffixed name form is pinned as a compatibility contract.

**Test Results:** full server suite 1876 passed, 3 skipped; vector store suite 300 passed; `ruff check`, `ruff format --check` clean; `ty check packages` unchanged at 20 baseline diagnostics, none in the touched files.

### Checklist

### Further comments

Independent of the collection-registry stack (#1526, #1527, #1530, #1531, #1533) and based on `main` so it can merge on its own. #1531 states the invariant this restores as a `VectorStoreCollection` contract and currently names these two stores as known non-conformance; whichever lands second should drop that note.

Terminology follows #1545, which introduces the same concept in the segment store: "incarnation" rather than "generation", since the value is a random uuid and `generation`/`epoch` invite the assumption that incarnations are ordered and comparable. The branch name still says `generation`; renaming it would break this PR's head ref, so it stays.

Two adjacent items are deliberately left out. `SQLiteVecVectorStore.create_collection` still reads, checks, then inserts although its primary key could arbitrate directly, which is a separate defect from this one. And a shared conformance test running this sequence against every `VectorStore` implementation would be worth having, but it needs the container-backed suites and belongs in its own change.

### DIFF
--- a/packages/server/server_tests/memmachine_server/common/vector_store/test_sqlite_vec_vector_store.py
+++ b/packages/server/server_tests/memmachine_server/common/vector_store/test_sqlite_vec_vector_store.py
@@ -7,6 +7,7 @@ from uuid import uuid4
 
 import pytest
 import pytest_asyncio
+from sqlalchemy.exc import OperationalError
 from sqlalchemy.ext.asyncio import create_async_engine
 
 from memmachine_server.common.data_types import SimilarityMetric
@@ -1011,3 +1012,51 @@ class TestFilterEdgeCases:
         assert r1.uuid in uuids
         assert r3.uuid in uuids
         assert r2.uuid not in uuids
+
+
+# ── Collection incarnations ──
+
+INCARNATION_CONFIG = VectorStoreCollectionConfig(
+    vector_dimensions=VECTOR_DIM,
+    similarity_metric=SimilarityMetric.COSINE,
+)
+
+
+def test_table_names_carry_the_incarnation():
+    """Distinct incarnations must name distinct tables."""
+    assert (
+        SQLiteVecVectorStore._collection_prefix("ns", "nm", "abc")
+        == "vector_store_sqlite_vec_2_ns_2_nm_abc"
+    )
+    assert SQLiteVecVectorStore._collection_prefix(
+        "ns", "nm", "abc"
+    ) != SQLiteVecVectorStore._collection_prefix("ns", "nm", "def")
+
+
+class TestCollectionIncarnations:
+    """A handle must never reach a collection recreated under its name."""
+
+    @pytest.mark.asyncio
+    async def test_stale_handle_cannot_write_to_recreated_collection(self, store):
+        await store.create_collection(
+            namespace=NAMESPACE, name=NAME, config=INCARNATION_CONFIG
+        )
+        stale = await store.open_collection(namespace=NAMESPACE, name=NAME)
+        assert stale is not None
+        await stale.upsert(records=[_make_record(vector=_normalize([1.0, 0.0, 0.0]))])
+
+        await store.delete_collection(namespace=NAMESPACE, name=NAME)
+        await store.create_collection(
+            namespace=NAMESPACE, name=NAME, config=INCARNATION_CONFIG
+        )
+
+        resurrected = _make_record(vector=_normalize([0.0, 1.0, 0.0]))
+        with pytest.raises(OperationalError):
+            await stale.upsert(records=[resurrected])
+
+        fresh = await store.open_collection(namespace=NAMESPACE, name=NAME)
+        assert fresh is not None
+        results = await fresh.query(
+            query_vectors=[_normalize([0.0, 1.0, 0.0])], limit=10
+        )
+        assert not results[0].matches
--- a/packages/server/server_tests/memmachine_server/common/vector_store/test_sqlite_vector_store.py
+++ b/packages/server/server_tests/memmachine_server/common/vector_store/test_sqlite_vector_store.py
@@ -7,6 +7,7 @@ from uuid import UUID, uuid4
 import pytest
 import pytest_asyncio
 from sqlalchemy import func, select, update
+from sqlalchemy.exc import OperationalError
 from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine
 
 from memmachine_server.common.data_types import SimilarityMetric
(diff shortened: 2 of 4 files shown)


edwinyyyu:
Rejected by #1579 if #1579 is accepted.

PR #1541, opened 2026-08-28
fix(event-backend): make expand_context return timeline-neighbor episodes

### LINKED ISSUE #1540: expand_context is silently ignored by the event backend's REST search path
## Summary

On the event backend, the `expand_context` parameter of `POST /api/v2/memories/search` has no effect on the response in the default (reranker-less) configuration. The expansion work is performed and then discarded.

## Evidence

Responses are byte-identical for `expand_context: 0` and `expand_context: 5` against a live server (event backend, `reranker: null`, 2k-episode project) — verified by direct comparison of parsed response bodies.

## Mechanism

`expand_context` threads correctly all the way down: spec → `MemMachine.query_search` → `LongTermMemory._search_scored_event` → `EventMemory.query`, which fetches up to N timeline-neighbor segments per seed (a LATERAL query per direction), decodes and materializes all of them into `ScoredSegmentContext.segments`.

The only consumer of those contexts is `LongTermMemory._search_scored_event`, which reads exactly one segment per context (the seed, located by `seed_segment_uuid`) and exactly one property of it (`_episode_uid`), plus the score. The neighbor segments are never referenced. The response schema (`EpisodicSearchResult`) also has no field that could carry context.

So on this path the parameter's entire effect is extra Postgres work and materialization per request.

Where it is *not* inert: with a reranker configured, the windows feed the reranker's candidate texts, so `expand_context` does change scoring there; and library callers of `EventMemory.query` receive the windows directly.

## Expected behavior

The declarative backend treats matched episodes as nuclei, attaches timeline-neighbor episodes as context, and folds them into the returned list with a fill-to-limit algorithm (`DeclarativeMemory._unify_scored_anchored_episode_contexts`): contexts of the best matches are included whole while they fit, then filled by index-proximity to the nucleus until `max_num_episodes` is reached.

The event backend should do the equivalent: derive the neighbor episodes from the expanded segment windows (each segment carries its `_episode_uid`) and unify them into the returned episodes the same way, so `expand_context` returns surrounding context episodes as the parameter implies.

## Environment

Observed at current `main` (231ce171); the consuming code is `packages/server/src/memmachine_server/episodic_memory/long_term_memory/long_term_memory.py` (`_search_scored_event`).

---

edwinyyyu:
Fixes #1540.

## Problem

On the event backend, `expand_context` was silently inert on the search path: `EventMemory.query` fetched and materialized the expanded segment windows (a LATERAL query per direction), but `LongTermMemory._search_scored_event` read only the seed segment's `_episode_uid` and score from each window, and the response schema has no context field. Responses were byte-identical for `expand_context` 0 and 5, while every request paid for the expansion.

## Fix

Bring the event backend to parity with the declarative backend's context handling (`DeclarativeMemory._unify_scored_anchored_episode_contexts`):

- Each scored window contributes the episodes its segments belong to — chronological within the window, the seed's episode as the nucleus. (Every segment carries its `_episode_uid`, so segment-space windows map directly onto episode-space context; with the passthrough segmenter this is 1:1 timeline neighbors.)
- Windows are unified best-score-first with the same fill algorithm: taken whole while they fit within `num_episodes_limit`, then filled by weighted index-proximity to the nucleus (forward recall preferred over backward, same weighting as declarative) until the limit is met. An episode keeps the score of the first window that contributed it.
- The unified context is returned **chronologically**, matching the declarative backend's ordering contract for expanded results.
- `expand_context` is clamped to `num_episodes_limit - 1` (declarative parity).

`expand_context == 0` behavior is unchanged — score-ordered seed episodes exactly as before, so existing callers see no difference unless they pass the parameter. Reranked configurations gain the same folding on top of reranker-scored windows (which already consumed the segments for scoring).

## Follow-up self-review

Two things the first version got wrong:

- **`expand_context` could go negative.** The quota clamp `min(max(0, expand_context), num_episodes_limit - 1)` is `-1` when `num_episodes_limit == 0`, which `SearchMemoriesSpec.top_k` allows (no lower bound). `EventMemory._query` then derives `max_backward_segments = -1` and asks the segment store for a negative window — undefined by the `SegmentStorePartition` contract; the SQLAlchemy store happens to short-circuit on `<= 0`, the in-memory store computes an empty slice and drops the seed. The floor is now applied last.
- **Neither end-to-end test could tell the fix from its absence** — both passed unmodified against the pre-fix `long_term_memory.py`. `FakeEmbedder` maps text to `[len(text), -len(text)]`, so under cosine every document scores exactly 1.0 against every query: all seven timeline episodes are seeds of equal rank, ties keep insertion order (chronological), and `num_episodes_limit=7` returns all seven with or without expansion. The "expansion adds episodes" assertion compared a limit-2 search against a limit-7 one, so the limit alone explained the difference.

The expansion tests now give each episode its own similarity from an explicit search rank, with the match's four timeline neighbours ranked last. That lets them assert the contract rather than an outcome — an exact episode list or exact score values would be pinned to the fixture's ranking and to the `expand_context // 3` split, neither of which the fix claims:

- No correct top-k can return those neighbours, and any nonzero window around the match reaches at least one of them whatever the backward/forward split, so "expansion returned a neighbour the search itself would not" survives a change of ranking or of the split. Chronological order and the episode limit are asserted alongside it.
- The clamp is asserted on the call made to the segment store (`0 <= backward + forward <= limit - 1`, over several limit/`expand_context` pairs) rather than on which episodes come back.
- `expand_context == 0` is asserted as "matches only, best score first", without naming them.

Exact lists and score values stay in the unit tests, which own the fill algorithm and the score-retention rule. The two expansion tests that pass a window fail against the code they cover; the zero-window test pins the path the change leaves alone.

## Tests

- End-to-end through the in-memory event-backend wiring: expansion returns neighbor episodes beyond the plain matches, chronologically ordered, and never exceeding `num_episodes_limit`.
- Unit tests for the window→episode-uid extraction (dedup, nucleus identification) and the unification algorithm (whole-context fit, overflow proximity with forward preference, first-window score retention).

- The two window-passing expansion tests fail against the pre-fix `long_term_memory.py`; the unit tests cover the window→episode-uid extraction and the unification algorithm directly.

`ruff` (pinned 0.16.8 on this base) format + check clean; `ty check packages/server` reports no diagnostics from the touched files. Episodic + server test suites pass (514 passed, 1 skipped).

The same change, based on `speedkick`, is #1547.

### Stack

| # | PR | change |
|---|---|---|
| — | #1541 **(this PR)** | fix(event-backend): make expand_context return timeline-neighbor episodes |

| # | PR | change |
|---|---|---|

This PR is its one commit, `6b080cb4b`, directly on `main`, independent of the others: review and merge it in any order among the independent PRs (a later one touching the same file rebases). The stages' history carries the same change from [vector store scale-out 3/5] up, so their diffs include it until it merges.

### Verification

At its one commit, on 2026-09-21: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`), 1921 tests at its head `6b080cb4b`.


packages/server/server_tests/memmachine_server/episodic_memory/long_term_memory/test_event_backend_wiring.py
```diff
@@ -587,3 +588,278 @@ async def test_score_threshold_none_keeps_all_results_under_euclidean():
...
+    assert not (plain_uids & _NEIGHBOUR_UIDS)
+    assert expanded_uids & _NEIGHBOUR_UIDS
+
+    # Expanded results come back chronologically, within the episode limit.
+    created = [ep.created_at for _, ep in expanded]
+    assert created == sorted(created)
+    assert len(expanded) <= 3
+
+
+async def test_expand_context_zero_returns_matches_in_score_order(
```
edwinyyyu, line 747: `test_expand_context_zero_returns_matches_in_score_order` cannot fail against the pre-fix code: at `expand_context=0` the new clamp is the identity and the `if expand_context > 0:` branch is not taken, and the source change is purely additive, so pre- and post-fix execute the same statements. The other two expansion tests do discriminate; the description's "all three expansion tests fail against the code they cover" is two. Same on #1547.

edwinyyyu (commented):
One test-coverage note (line comment); the same test is on the speedkick merge, #1547.

marvinyu-memverge (commented):
Reviewed at `cde0657`, base `main` @ `4ad28de`. The fold itself checks out; two asks, both on the same seam.

**1. `expand_context` is spent in segments, but every published surface documents it in episodes.**

`EventMemory._query` derives the window straight from it: `max_backward_segments = expand_context // 3`, `max_forward_segments = expand_context - max_backward_segments` (`event_memory.py:450`). The API describes that same integer as episodes in three places - the OpenAPI field description ("The number of additional episodes to include around each matched episode", `doc.py:213`), the Python client (`client/memory.py:385`), and `MemMachine.query_search` (`memmachine.py:997`).

Those two units coincide only under `PassthroughSegmenter`, which the body says. But `TextSegmenterConf` ships and is config-selectable (`episodic_config.py:141`, `type: "text"`, `max_chunk_length` default 500), and under it the new clamp claims parity in episode terms while bounding a segment budget. Concretely, at `max_chunk_length: 500` with episodes of about 2,000 characters (4 segments each), `top_k=10, expand_context=9` clamps to 9, splits 3 backward / 6 forward, and returns roughly 2 neighbour episodes for the 9 that were asked for. No error, and nothing in the response tells the caller.

The conflation predates this PR - it lives in `event_memory`, not in your diff. What changes here is that `expand_context` stops being inert, so this is the first release where it is visible. Is the clamp meant to be in segment space, or is the episode wording meant to be qualified to the passthrough case?

**2. Nothing exercises more than one segment per episode through `search_scored`.**

Every fixture in `test_event_backend_wiring.py` builds with `PassthroughSegmenter()` (lines 192, 449, 510, 707), and `TextSegmenter` shows up only in the `event_memory` tests. That is precisely the configuration in which segments and episodes cannot come apart, so the window-to-episode fold, the dedup in `_episode_uid_context`, and the clamp's parity claim are each pinned only where they are trivially true. One wiring test at `TextSegmenter` with a chunk length small enough to split each episode would cover the dedup path and turn whatever you decide on ask 1 into an assertion. (Separate from the zero-test point you already have open on line 747.)

**Verified.** `_unify_scored_uid_contexts` is a faithful mirror of `DeclarativeMemory._unify_scored_anchored_episode_contexts` (`declarative_memory.py:617`, with `_weighted_index_proximity` at :674): same break condition, same whole-context fit test, same `setdefault` score retention, same `(proximity - 0.5) / 2` forward weighting, same final chronological sort. The one divergence - applying the zero floor after the quota clamp rather than before - is the safer direction, and your comment names it. "Best windows first" holds under both metric directions, since `EventMemory._query` sorts with `reverse=higher_is_better` and that flag accounts for both the reranker and the collection's similarity metric. `num_episodes_limit=0` returns empty rather than one episode: the seed loop's `break` fires after the first append and would return one, but `vector_search_limit` is 0 as well, so there are no seeds - which is also why `assert windows` still holds in the clamp test, as `get_segment_contexts` is called unconditionally with an empty seed list. And the assumption the proximity fill rests on is real: the SQLAlchemy store assembles the window as `[*reversed(backward_rows), seed_row, *forward_rows]` (`sqlalchemy_segment_store.py:374`).

edwinyyyu:
1. Segment space. There is no exact segment-to-episode mapping once chunking is configured, and windowing by true episodes would need new EventMemory/SegmentStore operations, which is not worth it: the passthrough segmenter already gives the Neo4j-backed DeclarativeMemory's behavior, one segment per episode. Expanding by episodes is a matter of configuring it; a splitting segmenter is new behavior. TextSegmenter is neither recommended nor unrecommended -- passthrough stays closest to the declarative backend; TextSegmenter chunks for token efficiency in retrieval (particularly outside the episode-based API) at the cost of not mapping well onto the existing API. The clamp is in segments and bounds episodes a fortiori (a segment belongs to one episode), so the parity claim holds in either unit; the comment at the clamp now says so. The API wording stays. eecee287c.
2. Same commit: the timeline fixture under both segmenters at expand_context=3, limit 4, TextSegmenter(max_chunk_length=9) splitting each episode into three segments. Episodes at the match's score are the match's own window: three neighbours under passthrough, one under the splitting segmenter for any backward/forward split. Fails against the pre-fix source.

marvinyu-memverge (approved):
Approving at `eecee28` (base `main` @ `3809561`). Both asks from my last round are settled.

1. Segments it is. That's a fine call: passthrough is the default (`episodic_config.py:238`), so the API's "episodes" wording holds out of the box, and the clamp comment now says what a splitting segmenter changes.
2. The new splitting-segmenter test does the job. I ran it: 24/24 pass at head, and with `long_term_memory.py` swapped back to its pre-fix version, all three window-passing tests fail, the new one included. Using the match's score to identify the match's own window is sound, since only `tok-3` embeds at cosine 1.0 and every other segment scores strictly lower.

Optional, non-blocking: the only place that says a `type: text` segmenter shrinks what `expand_context` reaches is the code comment. One clause in `TextSegmenterConf`'s description would put that in front of the operator who opts in. Your call.

Commit 1 is a pure rebase of what I reviewed at `cde0657` (identical added/removed lines), so my earlier verified-parity notes still stand.

malatewang (approved):

PR #1545, opened 2026-08-28
Overhaul segment store: shared tables with incarnation-scoped tenant keys (fixes #1544, #1546, #1549)

### LINKED ISSUE #1544: Deleting a segment store partition drops the derivative-link foreign key for all partitions, and can deadlock with a concurrent create
## 1. Deleting one partition disables the foreign key for every other partition

`_drop_pg_child_tables` (`sqlalchemy_segment_store.py:1064-1080`) drops both child tables with `CASCADE`. `segment_store_dv_ln` has a foreign key to `segment_store_sg` declared on the partitioned parents, so the CASCADE does not stop at the partition being deleted -- it drops the parent-level constraint outright.

Measured on PostgreSQL 16, with two partitions `a` and `b`, deleting `b` and then inserting a `dv_ln` row in `a` that points at a nonexistent segment:

```
baseline (nothing dropped):                  orphan accepted = False
after dropping partition b (current order):  orphan accepted = True
after dropping partition b (reversed order): orphan accepted = True
```

After any partition deletion the store no longer enforces the link for the partitions that remain, and `ON DELETE CASCADE` stops working with it -- so `delete_segments`, which relies on it ("CASCADE deletes derivatives via FK", line 682), silently leaves orphaned derivative links behind. Those rows are visible through the public API, because `get_derivative_uuids_by_segment_uuids` reads `segment_store_dv_ln` directly with no join to `segment_store_sg`: deleted segments keep returning their derivative UUIDs.

This is deterministic, not a race, and it happens on the first partition deletion.

Reversing the drop order does not help, as the measurements above show. Dropping without `CASCADE` fails with `DependentObjectsStillExistError`, which is presumably why `CASCADE` is there. Detaching each child before dropping it preserves both the constraint and the cascade:

```
DROP CASCADE, dv_ln then sg (today)      ok      FK still enforced=False  cascade works=False
DROP CASCADE, sg then dv_ln (reversed)   ok      FK still enforced=False  cascade works=False
DROP without CASCADE, dv_ln then sg      error   FK still enforced=True   cascade works=True
DETACH both, then DROP                   ok      FK still enforced=True   cascade works=True
```

## 2. Concurrent create and delete can deadlock

The two paths reach the parent tables in opposite order:

- `_create_pg_child_tables` (1044-1062): `segment_store_sg`, then `segment_store_dv_ln`
- `_drop_pg_child_tables` (1064-1080): `segment_store_dv_ln`, then `segment_store_sg`

Both statement kinds take `ACCESS EXCLUSIVE` on the parent, so the order matters. Nothing serializes the two paths beforehand: the create paths take `LOCK TABLE segment_store_pt IN SHARE ROW EXCLUSIVE MODE` (line 801), while `delete_partition` takes only `SELECT ... FOR UPDATE` on the partition row (936-940), which is a `ROW SHARE` table lock -- and `ROW SHARE` does not conflict with `SHARE ROW EXCLUSIVE`.

Creating one partition while deleting a different one, across seven sampled interleavings against PostgreSQL 16:

```
delete: FOR UPDATE only (today)         -> deadlocks in 4/7 interleavings
delete: LOCK TABLE + FOR UPDATE (fix)   -> deadlocks in 0/7 interleavings
```

The row lock has to stay: it is what makes `delete_partition` wait for in-flight writers, which hold `FOR SHARE` on the same row via `_lock_partition_for_write` (222-234). `SHARE ROW EXCLUSIVE` does not conflict with the `ROW SHARE` those writers take, so the table lock alone would let the drops start before writers finish. The fix is to take the table lock *in addition to* the row lock, not instead of it.

Dropping the constraint explicitly is possible but expensive. `ALTER TABLE segment_store_dv_ln DROP CONSTRAINT ...`, then the two `DROP TABLE`s, then re-adding the foreign key all succeed, but the re-add validates every existing row across all remaining partitions -- an O(N) scan in place of an O(1) partition drop, holding `SHARE ROW EXCLUSIVE` on both tables while it runs. The usual escape hatch is unavailable: PostgreSQL rejects `ADD CONSTRAINT ... NOT VALID` for a foreign key on a partitioned table (`cannot add NOT VALID foreign key on partitioned table`). Removing only the per-partition constraint entry is also refused (`cannot drop inherited constraint`), so it is the whole constraint or nothing.

Detaching is the constant-time alternative, and only the `segment_store_sg` child strictly needs it -- the link child has nothing depending on it. This change detaches both, so the rule reads uniformly and does not depend on which way the foreign key points.

## Scope

The event-memory backend is opt-in: a missing `backend` deserializes as `"declarative"` (episodic_config.py:13-30), and the change that would make `event` the default is not merged. So this only affects a database where someone explicitly configured `backend: event` on PostgreSQL and has deleted a partition -- development and benchmark databases in practice.

Worth knowing for those: the constraint does not come back by itself, because `metadata.create_all(checkfirst=True)` skips tables that already exist. Recreating the schema is the simplest remedy; no migration is warranted at this stage.

## Fix

Both are addressed in the linked PR: detach each child partition before dropping it, and have `delete_partition` acquire the same `LOCK TABLE ... IN SHARE ROW EXCLUSIVE MODE` the create paths take, before the existing row lock. A regression test covers the first (it fails against PostgreSQL before the change and passes after); the second is a lock-ordering change with no black-box assertion, since a deadlock test would be timing-dependent.

---

### LINKED ISSUE #1546: Episodic search reads lock every segment-store partition once prepared statements go generic (Postgres lock-table exhaustion, 500s under load)
### Summary

On PostgreSQL, every segment-store read on the episodic search path (`SQLAlchemySegmentStorePartition.get_segment_contexts` and friends) queries the partitioned parent tables (`segment_store_sg`, `segment_store_dv_ln`) with the partition key as a bind parameter. asyncpg prepares every statement, and PostgreSQL switches a prepared statement to a cached generic plan after 5 executions. A generic plan must be valid for any parameter value, so it references **all** child partitions, and PostgreSQL acquires an `AccessShareLock` on every relation in the plan **before** runtime pruning discards the non-matching ones.

Result: after warmup, every single-partition read locks `#partitions + parents + indexes` relations. With a few hundred partitions and realistic concurrency this saturates the database's CPU with lock churn and generic-plan executor startup, and overflows the fixed-size lock table, at which point searches fail with HTTP 500.

### Measurements

Reproduced on a store with **316 partitions** (pgvector/pgvector:pg16 container, 2 CPUs, default settings). Relation locks held by one backend per execution of `SELECT uuid FROM segment_store_sg WHERE partition_key = $1 LIMIT 1` as a prepared statement:

```
unprepared (custom plan):      9 relation locks   (parent + matched child + indexes)
prepared, executions 1-5:      9 locks each       (plan-time pruning)
prepared, execution 6:      1269 locks            (generic plan is built: plans all 316 children)
prepared, execution 7+:      319 locks            (cached generic plan: locks every child, then prunes)
```

Under a closed-loop REST search benchmark (long-term episodic only, 4 uvicorn workers, SQLAlchemy pool 10+5 per worker, ~128 concurrent requests):

- PostgreSQL pinned at **153-158% of its 2-CPU allowance** while the Python workers idled at 16-23% CPU each — the store, not the application, was the bottleneck.
- **239 requests failed with HTTP 500**: `asyncpg.exceptions.OutOfMemoryError: out of shared memory` (lock-table exhaustion). The arithmetic: ~40 sessions x 319 locks ≈ 12,600 needed, vs `max_locks_per_transaction x (max_connections + max_prepared_transactions)` = 64 x 104 ≈ 6,656 available.
- Throughput anti-scales: deepening the request queue *reduces* throughput (measured 104 → 63 req/s going from 64 to 128 concurrent requests with one worker) as per-query lock/plan overhead and lock-manager contention grow.

The failure threshold is roughly `#partitions x concurrent_sessions > 64 x max_connections`. One partition is created per project, so a deployment with a few hundred projects hits this under modest concurrency with default PostgreSQL settings.

### Why the locks happen

The `AccessShareLock` per referenced relation is PostgreSQL's protection against concurrent DDL (e.g. `DROP`/`DETACH` of a partition mid-query) and cannot be skipped. What can be avoided is referencing 316 relations in the first place: the store knows the partition key at call time and even creates the child tables itself (`segment_store_sg_p_<key>`), so the plan only needs to name one child.

### Fix options

1. **Query the child tables directly** (preferred): the partition handle already owns its key; issuing DML against `segment_store_sg_p_<key>` / `segment_store_dv_ln_p_<key>` makes every plan reference exactly one relation — no 316-way planning, 1-2 locks, immune to the generic-plan flip. Doable while keeping SQLAlchemy by mapping the ORM entities onto per-partition `Table` objects with `orm.aliased(SegmentRow, child_table)`. SQLite (no child tables) keeps the parent-table queries.
2. `plan_cache_mode = force_custom_plan` on the store's sessions: keeps plan-time pruning (9 locks) at the cost of replanning every call — still pays planning against the partitioned parent.
3. Raising `max_locks_per_transaction`: delays the 500s but does not address the CPU burn; not a fix.

A PR implementing option 1 is being folded into #1545 (single PR for the outstanding segment-store defects).

---

### LINKED ISSUE #1549: Segment store partition handles are not fenced to a partition incarnation: stale handles silently operate on a recreated partition
### Summary

A `SQLAlchemySegmentStorePartition` handle is not fenced to the partition incarnation it was opened against. If a partition is deleted and re-created under the same key while application code still holds the old handle, the stale handle silently operates on the new incarnation — while carrying the old incarnation's `config` and `payload_codec`. If the codec/config differ between incarnations, reads mis-decode and writes encode wrong bytes into the successor partition; even when they match, a handle the application believes dead keeps reading and writing the new partition's data.

This is the segment-store analog of #1536 (vector-store handles held across delete and recreate). Deletion *alone* now fails loudly on the PostgreSQL event backend (after #1545, a stale handle's statements reference dropped child tables and error) — the silent case is specifically delete + recreate, because the recreated child tables reuse the same names. On SQLite there is no per-incarnation object at all, so even plain deletion is only detected by the partition row's absence.

### What "ensuring the handle does not outlive the partition" can mean

The handle is an object the application holds; it cannot be destroyed remotely, and in-process invalidation cannot work in general (with `MEMMACHINE_WORKERS` pre-forking and multi-instance deployments, the delete can happen in a process that does not hold the handle — so a registry can improve same-process error messages but can never provide the guarantee). The achievable property is **fencing**: any use of a handle after its incarnation ends fails deterministically with a stale-handle error.

### Options

1. **No schema change — OID fencing (PostgreSQL only).** At open, capture `to_regclass('segment_store_sg_p_<key>')::oid`; verify it inside the same transaction as the existing `FOR SHARE` partition-row pin on every write (raise on mismatch). A recreated table necessarily has a new OID, and the pin excludes concurrent deletion, so the check is race-free. Does not cover SQLite; per-operation convention in code.
2. **Small schema change — an `incarnation` column on `segment_store_pt`** (opaque id assigned at create). The handle captures it at open; the existing write pin becomes `WHERE partition_key = :key AND incarnation = :incarnation` with a returned-row assertion, and reads can make the same cheap check. Same guarantee as (1) but explicit in the data model and dialect-uniform (covers SQLite). Requires a migration to backfill existing rows.
3. **Schema change in the namespace — incarnation suffix in child table names** (`segment_store_sg_p_<key>_<incarnation>`). The fence becomes structural: a stale handle references names that no longer exist, so every read and write fails loudly with zero per-operation checks, and a new incarnation can never collide with a dying one. Costs: needs the column from (2) anyway, a rename migration for existing children, breaks the pinned physical naming contract, and `segment_store_dv_ln_p_` + a 32-char key + an 8-char suffix sits exactly at PostgreSQL's 63-byte identifier limit.

Option (2) is the recommended shape: smallest change with the full guarantee on both dialects, and the check rides the locking structure #1545 established (the `FOR SHARE` pin), so it introduces no new race to reason about.

### Status

Filed for tracking; intentionally not being implemented right now because options (2)/(3) are schema changes (maintainer decision to defer).

---

### LINKED ISSUE #1557: [Bug]: Segment store corrupts non-UTC timezone-aware timestamps on SQLite (write path + filter bounds)
### What happened?

`SQLAlchemySegmentStore` corrupts non-UTC timezone-aware segment timestamps when running on SQLite, in two ways:

**1. Write path: the stored instant is wrong.**

SQLite's `DateTime(timezone=True)` does not store timezone information: SQLAlchemy's SQLite dialect discards `tzinfo` and persists the datetime's wall-clock fields verbatim, reading them back naive. The store's read path already assumes the stored value is UTC (it reapplies the separately stored `timestamp_timezone_offset` on read), but the write path persisted `ensure_tz_aware(segment.timestamp)` without normalizing to UTC. So a non-UTC timestamp was stored as its original zone's wall clock, then reinterpreted as UTC on read — shifting the instant by the original UTC offset.

Concretely, a segment timestamp of `2024-01-01 13:30:45-08:00` comes back as `2024-01-01 05:30:45-08:00`.

**2. Filter bounds: datetime comparisons decide by wall clock, not instant.**

Property filters on `timestamp` compile through `_compile_column_leaf` in `common/filter/sql_filter_util.py`, which bound datetime values as-is. On SQLite the bind processor renders an aware datetime as wall-clock digits with the offset dropped, and the comparison against the stored text is lexical — so `timestamp <= 2024-01-01T08:00+08:00` excludes a row stored at `2024-01-01T00:00Z`, even though that is the very same instant.

PostgreSQL is unaffected in both cases (`timestamptz` stores and compares true instants); the bug is that the two backends disagree.

### Expected behavior

A timezone-aware timestamp roundtrips with its instant intact (the original offset is reconstructed from `timestamp_timezone_offset`), and a datetime filter bound means an instant regardless of the zone it is written in — identically on SQLite and PostgreSQL.

### Fix

Normalize to UTC before persisting (`ensure_tz_aware(...).astimezone(UTC)` in `_insert_segments`) and before binding datetime comparison values in `sql_filter_util._compile_column_leaf`. Both halves, with regression tests parametrized over UTC/-08:00/+05:30 (verified fail-before/pass-after on SQLite), are folded into #1545, which rewrites this store. The same fix for the write path originated in #1462.

This issue is scoped to the segment store only; the episode and cluster stores have the same write-path defect, tracked separately so each can be fixed in its own PR: #1558 (episode store), #1559 (cluster store).

---

### LINKED ISSUE #1558: [Bug]: Episode store corrupts non-UTC timezone-aware created_at on SQLite (write path + time-range bounds)
### What happened?

`SqlAlchemyEpisodeStore` corrupts non-UTC timezone-aware `created_at` values when running on SQLite.

SQLite's `DateTime(timezone=True)` does not store timezone information: SQLAlchemy's SQLite dialect discards `tzinfo` and persists the datetime's wall-clock fields verbatim, reading them back naive. `add_episodes` writes a caller-supplied `entry.created_at` without normalizing it to UTC first, so a non-UTC value is stored as its original zone's wall clock and reinterpreted as UTC on read — shifting the instant by the original UTC offset (e.g. `13:30:45-08:00` written, `13:30:45Z` effectively read back: 8 hours late).

The `start_time`/`end_time` range bounds in `_apply_episode_filter` are bound as-is for the same reason, so a bound named in a non-UTC zone is compared against the stored wall clock lexically rather than by instant.

Server-generated columns (`server_default=func.now()`) are always UTC and roundtrip fine. PostgreSQL is unaffected (`timestamptz` stores and compares true instants); the bug is that the two backends disagree.

### Expected behavior

A timezone-aware `created_at` roundtrips as the same instant, and `start_time`/`end_time` bounds select by instant regardless of the zone they are written in — identically on SQLite and PostgreSQL.

### Fix

Normalize to UTC before persisting and before binding the range bounds (`ensure_tz_aware(...).astimezone(UTC)`); #1462 contains this fix with regression tests. Datetime bounds in *property filters* compile through the shared `sql_filter_util._compile_column_leaf`, whose normalization is covered by #1557 / #1545; the write path and the explicit `start_time`/`end_time` bounds here are episode-store code and remain to be fixed in this store.

Scoped to the episode store; segment-store counterpart: #1557.

---

### LINKED ISSUE #1559: [Bug]: Cluster store corrupts non-UTC timezone-aware timestamps on SQLite (last_ts, pending created_at)
### What happened?

`ClusterStateStorageSqlAlchemy` corrupts non-UTC timezone-aware timestamps when running on SQLite.

SQLite's `DateTime(timezone=True)` does not store timezone information: SQLAlchemy's SQLite dialect discards `tzinfo` and persists the datetime's wall-clock fields verbatim, reading them back naive. `save_state` writes two caller-supplied timestamps without normalizing them to UTC first:

- `last_ts` on cluster rows (`ClusterInfo.last_ts`), and
- `created_at` on pending-event rows.

A non-UTC value is stored as its original zone's wall clock and reinterpreted as UTC on read, shifting the instant by the original UTC offset. PostgreSQL is unaffected (`timestamptz` stores true instants); the bug is that the two backends disagree.

### Expected behavior

Timezone-aware `last_ts` and pending `created_at` values roundtrip as the same instant on every dialect.

### Fix

Normalize to UTC before persisting (`ensure_tz_aware(...).astimezone(UTC)` at both write sites); #1462 contains this fix with regression tests.

Scoped to the cluster store; counterparts for the same root cause: #1557 (segment store), #1558 (episode store).

---

edwinyyyu:
### Purpose of the change

Fixes #1544, #1546, and #1549 by construction: the segment store's per-tenant partitioned layout is replaced with shared tables and an incarnation-carrying tenant registry, on every dialect. Design record: `design/segment_store_shared_tables.md`.

Also fixes #1557, #1558, and #1559: #1462's UTC-normalization fixes are folded in for every SQL store (segment, episode, cluster), so no store ships with the SQLite timestamp corruption.

### Why an overhaul instead of the earlier targeted fixes

This PR began as three targeted repairs to the partitioned layout (detach-before-drop for #1544, a management lock for the create/delete deadlock, child-table DML for #1546), each verified on its own. Further investigation showed the layout itself was the root cause and could not be fully repaired:

- PostgreSQL's FK integrity triggers still addressed the partitioned parents internally, so writes and deletes kept a per-backend lock spike proportional to partition count even after every client statement was fixed (10 -> 166 -> 10 relation locks across the trigger's generic-plan attempt).
- Writer-vs-lifecycle-DDL deadlocks through the shared parents survived every lock discipline: 41-83 deadlocks per 20-second churn on upstream and on every intermediate fix, in two cycle shapes captured from the server log, one of them a lock-upgrade cycle inherent to dropping a child that carries a cloned FK.
- Scaling requirements ruled out every per-tenant-table variant: tenant creation must stay cheap at 1e5-1e6 tenants (a per-tenant table costs ~5 catalog relations plus disk files), with 1e4-1e7 rows per tenant.

### Description

- **Shared tables everywhere.** The ORM models are the physical schema on every dialect. PostgreSQL partitioning, per-tenant DDL, detach machinery, the store-wide management lock, and the per-partition entity cache are removed. With no partitions there is no generic-plan lock fan-out, and with no lifecycle DDL there is nothing for the deadlock class to live in: the churn smoke that measured 41-83 deadlocks/20s on every partitioned build measures zero, with ~60x more write throughput and ~200x more lifecycle throughput.
- **Incarnation-scoped tenant identity.** `segment_store_pt` is the tenant registry: logical key (primary key) plus an `incarnation`, a random unique-constrained UUID. Data rows carry the incarnation alone, so a data query cannot be built without resolving the registry, and referencing the wrong tenant is structurally impossible. A deleted-and-recreated tenant never sees its predecessor's rows, even mid-purge. A colliding mint is rejected rather than left to probability: the unique constraint rejects a collision with a live incarnation, and the mint transaction re-checks the purge queue after its insert so an incarnation whose garbage is still awaiting purge is re-minted (race-free with the existing tables: the check runs after the registry insert, and no new queue entry for the minted value can appear before commit, since the only registry row carrying it is uncommitted). Any integrity rejection with no committed row under the key is likewise retried with a fresh incarnation, up to `_MAX_MINT_ATTEMPTS`; a persistent cause surfaces as `SegmentStoreAttemptsExhaustedError` with the database error chained. Rejections are warning-logged at the detection site, since a genuine collision is astronomically unlikely and the log is the signal for broken randomness or a persistent database error even when the re-mint heals it.
- **O(1) deletion plus a purge queue.** `create_partition` is a row insert (microseconds, no DDL). `delete_partition` takes `FOR UPDATE` on the registry row (waiting out writers' `FOR SHARE` pins), enqueues the incarnation on `segment_store_gc` (logical key kept for forensics), and deletes the row: O(1) at any tenant size, the pool-model contract of "unreachable immediately, erased asynchronously".
- **Purge as an ABC capability.** `purge_deleted_partitions() -> bool` is part of the `SegmentStore` interface. The caller's whole protocol is "call until False"; sizing is not a call argument, because callers cannot know engine-appropriate bounds. Deployments set them once at construction: `purge_max_segments` (default 10000) and `purge_max_partitions` (default 100, its own bound because retiring an empty entry costs ~0.95 ms, ~200x a segment row, and empty partitions are cheap to mass-create-and-delete). Each call is one transaction (measured ~147k-312k rows/s), committing its progress or nothing. Links follow by `ON DELETE CASCADE`, 50-68% faster than manual link deletion at one and four links per segment; cascade deletion saturates ~3M link rows/s (380k segments/s at 1 link/segment, 46k at 64), so heavily linked partitions purge in sub-second calls. Link fan-out is set by the deriver, which the store cannot reject after derivation. Before retiring an entry the call also reclaims any link rows that escaped referential integrity, in batches drawn from the same budget and logged as a warning; a link row deletes ~3x cheaper than a segment row (1.0 vs 3.3 us/row, batched), so the shared budget bounds the call without assuming a ratio. Entries are claimed one at a time, oldest first (enqueue stamped by the database clock, indexed), with `FOR UPDATE SKIP LOCKED`, so a call neither materializes nor locks the rest of the backlog and concurrent purgers from any process share the queue; only the claiming call touches a dead incarnation's rows, so reclamation is deadlock-free by construction. On SQLite, which drops locking clauses, purgers serialize on the write lock at the DELETE: two may claim the same entry, and the second finds no rows and re-retires it. The store never schedules purging; implementations whose deletes reclaim physically return `False`. The delete path does not purge: an inline drain was tried, first of the global queue and then scoped to the deleted key, and removed, since prompt physical erasure is not a promise the store can keep on every dialect (on SQLite any writer past the busy timeout fails) and the sweeper reclaims within its interval; a deployment must run the sweeper, and the ABC says so.
- **Fencing (#1549).** Writes pin the registry row under `FOR SHARE` with an incarnation predicate. Reads add the same predicate to their data statement as an `EXISTS`, so one statement checks liveness and reads; a read that finds no rows issues the registry check on its own, to tell an empty partition from a stale handle, and a windowed read ends with one registry read because its statements take separate snapshots. On SQLite the driver defers BEGIN until the first write, so a SELECT-only fence checks nothing; BEGIN IMMEDIATE would mean taking over transaction management of the caller's shared engine, so the fence is a self-checking registry-row UPDATE (same write lock, scoped to the transaction, match count as the staleness check), and deletion opens its transaction the same way. A handle that outlives its partition or its incarnation raises `SegmentStorePartitionHandleStaleError` on every dialect; SQLite previously had no stale-handle detection at all. Measured cost: none on reads that find rows, one registry round trip on reads that find none.
- **Foreign keys.** The segment table's FK to the registry is removed on purpose (registry and data rows decouple so deletion is O(1)); the link table's FK to the segment table and its cascade remain, enforced by a single-row indexed check with no partition fan-out. On SQLite, enforcement is per-connection state registered at ENGINE creation (`enable_sqlite_foreign_keys`, applied by the DatabaseManager to its engines): a listener added later misses connections the shared engine already pooled, a pre-existing bug surfaced in review round 4. The store states the requirement on its `engine` parameter and does not verify it, which is the reference practice for SQLite foreign keys; an engine without the pragma leaves link rows that outlive their segments until the partition is dropped, when the purge reclaims them with a warning. The SQLite vector stores carry the same latent pattern; filed as #1568.
- **Folded in: #1462's SQL-store fixes (fixes #1557, #1558, #1559).** App-supplied timezone-aware datetimes are normalized to UTC before persisting in every SQL store: segment timestamps, episode `created_at` and its `start_time`/`end_time` bounds, cluster `last_ts` and pending `created_at`. Without this, SQLite (whose `DateTime(timezone=True)` discards tzinfo) read non-UTC values back shifted by their offset. Datetime filter values are normalized to UTC-aware instants at `Comparison`/`In` construction, the filter language's contract stated on `FilterExpr` (a value denotes an instant; naive means UTC), so every consumer receives instants and compilers only choose a representation; the SQL column leaf binds values as-is, and remaining per-backend normalizations are idempotent defenses, removable separately. Two consumers change by design: the Neo4j compiler's `datetime.timestamp()` read a naive value in the server's local zone and now receives instants, with its ISO-string branch parsing under the same rule (pinned under a non-UTC process zone); the in-memory short-term evaluator tags a naive stored metadata datetime as UTC at comparison time, since stored user data is not rewritten. The episode store's bounds spell the convention inline as two explicit steps, `ensure_tz_aware(...).astimezone(UTC)`, kept unwrapped so the naive-means-UTC decision stays visible at each site. PostgreSQL is unaffected. On SQLite, rows the episode and cluster stores previously wrote from non-UTC-offset datetimes hold the local wall clock as if it were UTC and read back shifted before and after this change alike; the offset was never stored, so only new writes are correct, and a range query spanning the upgrade mixes the two conventions.

### Measurements

Layout comparison and scaled runs are in the design doc. Headlines (pgvector:pg16, one harness for all arms): tenant creation 0.006 ms vs 6-12 ms of per-tenant DDL; ingest and read parity or better at 40x2k and 1M-pair scales including a 500k-row tenant; O(1) delete at any size; churn 0 deadlocks vs 41-83.

Store-API ABAB against upstream main (3 interleaved rounds, one PG instance, medians, at the revision before the read fence was folded into the data statement): ingest +42% (11.1k vs 7.8k pairs/s), `delete_segments` +20%, event/derivative lookups 10-17% faster, windowed context expansion 8% faster (5.9 vs 6.4 ms), lifecycle create+open+delete 4.4x faster (2.9 vs 12.8 ms/cycle); seed context reads +0.19 ms (1.30 vs 1.11 ms) from the fence's separate registry round trip, since removed. Read-path ABAB of that fold (5 interleaved rounds, medians): seed context reads 1.17 vs 1.42 ms, event lookups 1.04 vs 1.61 ms, derivative lookups 1.05 vs 1.33 ms, windowed context expansion 8.74 vs 9.67 ms (8 rounds x 600 reps, paired median -0.91 ms); reads that find nothing unchanged. Server-side (`EXPLAIN ANALYZE`, best of 30) the `EXISTS` conjunct plans as a one-time InitPlan costing ~3 us per statement, so the client-side gain is the round trip. Purge drains ~216k rows/s; the batched-delete pattern was benchmarked on a 100k-row incarnation with the cascade in place: `uuid IN (SELECT ... LIMIT n)` 252k rows/s vs a PostgreSQL-only ctid batch at 238k and an unbatched single-DELETE ceiling at 295k, so the portable pattern is also the fastest batched one. Claim and mint-check statements measure 215/156 us.

### Testing

- The behavioral suite runs unchanged against the new layout on both dialects: segment store file 82 in the `-m "not integration"` lane, 86 in the `-m integration` lane; full server suite 1926 passed, 3 skipped. Concurrency coverage runs on both dialects wherever the property exists on both: lifecycle churn, racing deletions (single-enqueue assertion), overlapping segment deletes, the mint-vs-deletion collision race, and O(1) deletion via recorded SQL.
- Purge: `test_purge_reclaims_oldest_garbage_first` (FIFO, explicit stamps), `test_purge_queue_stamps_enqueue_time_from_database_clock`, `test_purge_bounds_entries_processed_per_call`, `test_purge_batches_links_that_escaped_integrity` (orphans staged through a second SQLite engine without the pragma; fails against the unbounded form), `test_purge_skips_entries_claimed_by_concurrent_purger` (fails with `SKIP LOCKED` ablated), `test_concurrent_purges_reclaim_everything` (both dialects), `test_purge_claims_queue_entries_incrementally` (fails against the claim-all form), `test_purge_reclaims_only_dead_incarnations`, `test_purge_bound_comes_from_params`, `test_default_purge_bound_comes_from_params`.
- Incarnations: `test_incarnation_with_garbage_left_is_never_reused` (queue re-check) and `test_incarnation_colliding_with_live_partition_is_never_reused` (unique constraint) force a collision on both creation paths and both dialects, each failing with its guard ablated; `test_mint_detects_collision_with_concurrent_deletion` pins the insert-then-check ordering with a staged concurrent deletion (swapping the statements fails only this test); `test_persistent_integrity_error_surfaces_with_cause` pins the bounded re-mint with the driver error chained, both dialects; `test_persistent_mint_failure_raises_instead_of_looping` pins the attempt cap on both creation paths.
- Fencing: `test_stale_handle_raises_after_delete` / `..._after_recreate` (both dialects), `test_reads_check_liveness_inside_the_data_statement`, `test_windowed_read_raises_when_partition_dies_between_statements` (both dialects; fails without the closing registry read), `test_recreated_partition_is_isolated_from_old_rows`, `test_delete_partition_touches_only_registry_and_queue` (recorded SQL), `test_write_pin_blocks_partition_delete`, the SQLite fence test (a write racing delete-plus-purge can no longer orphan rows; the two SQLite race tests carry started-events so a loaded box cannot pass them vacuously).
- Lock-necessity tests (integration lane, under a second): each locking property is pinned by staging the interleaving it serializes (blocked-ness observed via `pg_stat_activity`, no grace sleeps). Verified by per-lock ablation on both generations: removing a lock from the new store fails exactly its targeted tests, and the pre-overhaul store fails the churn and concurrent-delete tests with all its locks intact (reproducible `DeadlockDetectedError` in the two cycle shapes on #1546). The churn test also caught the `_open_or_create_partition` race, now a retry loop bounded by `_MAX_MINT_ATTEMPTS`.
- Timestamps: #1462's regression tests for the episode and cluster stores are ported; `test_timestamp_roundtrips_with_timezone` and `test_timestamp_filter_compares_instants_not_wall_clocks` (#1557) are parametrized over non-UTC zones on both dialects; `test_older_than_compares_instants_not_wall_clocks` covers the semantic history bound; the filter-parser tests pin construction-time normalization; `test_iso_string_coerces_as_utc_instant` pins the Neo4j string branch under a non-UTC process zone (skipped where `time.tzset` is unavailable); `test_datetime_metadata_filters_tag_naive_stored_values` pins the in-memory evaluator. Each fails without its fix.
- Erasure and scheduling: `LongTermMemory` deletes the collection and the partition and returns; the partition is unreachable at once and its rows are reclaimed by the sweeper within its interval. Its handles are nulled so later use raises (`test_event_backend_unusable_after_drop_session_partition`), and the wiring test asserts the delete path never purges. The resource manager runs one background purge task per store (bounded calls a short pause apart while the store reports a backlog, one idle call per interval otherwise; failures logged and retried; cancelled on close); the loop is a module-level coroutine so a pending task does not pin the manager (`test_purge_task_does_not_pin_the_manager`), and `get_segment_store` after `close()` raises `ResourceManagerClosedError`. Both purge bounds are validated at construction; the server uses the defaults, and config plumbing is future work.
- Other review-round fixes, each with a test that fails pre-fix: unloadable codec config commits no registry row (both creation paths); trailing-newline partition keys rejected (`re.fullmatch`), and the session-id-to-partition-key mapping validates with the store's own validator instead of a drifted copy; the StaticPool guard raises `ValueError` (`test_static_pool_engine_is_rejected`); pre-3.35 SQLite runtimes are rejected at construction (`test_old_sqlite_runtime_is_rejected`); empty-input `add_segments` short-circuits like `delete_segments`. Stale-handle handling above the store (API status mapping, cross-replica cache eviction) is filed as #1571.
- Retained: `test_delete_partition_keeps_other_partitions_cascading` and `test_delete_partition_keeps_foreign_key_enforced` now hold trivially, since no DDL can damage the constraint.
- `ruff format --check`, `ruff check`, and `ty check` clean (zero added diagnostics).

### Compatibility

No migration from the partitioned layout is provided: the event backend is opt-in and pre-GA, and existing databases recreate their schema (see the design doc). The partition-key contract (`[a-z0-9_]`, max 32 bytes) is unchanged.

### Type of change

- [x] Breaking change (schema layout of the opt-in event backend)

---

https://claude.ai/code/session_01MpFwBnZ6SxdMe3pSuMTCHD

https://claude.ai/code/session_01MbYdqGZsuws6Z2WHYfCCR5


edwinyyyu:
Copied to `speedkick` as #1548 — all five commits cherry-picked from `fad1f22c`, no conflicts, net diff byte-identical. If this PR gains further commits they will need porting across too.

edwinyyyu (commented):
Self-reviewed at `53ece734c0c39a9c64101764f5d8adcee5324a69`.

edwinyyyu:
Self-reviewed at `eb52ee1f4fcb2248c23fe12b07dab3fcd47ab308`.

edwinyyyu:
Self-reviewed at `82e6d3c7e36b0f3e749bd93288384030cfd52340`.

edwinyyyu:
Self-reviewed at `3f31e179ff757120d2f9ab756eefe107c55657f4` (skimmed). Segment store is essentially finalized. Wiring is maybe not finalized.

edwinyyyu:
Self-reviewed at `705487c8cab5010d207597d5bf4370fc89779bca` (skimmed).

edwinyyyu:
Self-reviewed at `dcd284a5b078947bc605d86937678adb5102907b` (skimmed).
Focused review on segment store. Other fixes are folded in for other components if they are in the same class to avoid fragmentation, or wiring to support the new changes.

edwinyyyu:
Self-reviewed at `2e58e1fc828ed83db9945446e2dd12b69f8c6b40` (skimmed). More changes incoming.

edwinyyyu:
Self-reviewed at `10d1f6acd845de4ea8eb5da1401753a0de49c7ee` (skimmed).

edwinyyyu:
I think we need more proper tenant records. We can make the lifecycle of each resource appear atomic but there is nothing to make tenant lifecycle appear atomic.

edwinyyyu:
Self-reviewed at `c38d54335532b1d715c28e4d4266f69f85e2acc7` (skimmed).
Essentially converged, very minor changes since `2e58e1fc828ed83db9945446e2dd12b69f8c6b40`.

edwinyyyu:
Merged into speedkick -- reopen if needed for independent merging into main.

PR #1550, opened 2026-08-29
feat: add OrcaRouter as a first-class language model provider

zsanig22-dotcom:
Adding [OrcaRouter](https://www.orcarouter.ai) as a first-class `orcarouter` provider in MemMachine's language model registry lets anyone running MemMachine's memory layer point their summarization and generation models at an OpenAI-compatible gateway that routes across many models from a single endpoint — with no custom base-URL plumbing.

MemMachine positions itself as *"the open-source memory layer for AI agents"* and advertises itself as **LLM agnostic** — working with OpenAI, Anthropic, Bedrock, Ollama, and any LLM provider. Today that flexibility is wired through the `resources.language_models` registry, where each entry names a provider (`openai-responses`, `openai-chat-completions`, `amazon-bedrock`, `litellm`). This PR mirrors the existing `openai-chat-completions` adapter: a new `orcarouter` provider that reuses MemMachine's `OpenAIChatCompletionsLanguageModel` and defaults `base_url` to `https://api.orcarouter.ai/v1`, so users can configure it exactly like the Ollama/OpenAI-compatible entries already in the sample configs.

OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding `orcarouter` as a first-class provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

What changed:
- `OrcaRouterLanguageModelConf` with a default model of `orcarouter/auto` and default base URL `https://api.orcarouter.ai/v1`
- Registration in `LanguageModelsConf` (parse, serialize, getters, `contains_language_model`)
- Build path in `LanguageModelManager._build_orcarouter_language_model` and runtime add/remove/status in the config service and API spec (`config_spec`, `doc`)
- Docs: install guide, `docs/open_source/configuration.mdx`, and the three `sample_configs/*.sample` files
- Regenerated `docs/openapi.json` provider enum
- Unit tests for config parse/round-trip/getters and manager build

Verification:
- `ruff check` and `ruff format --check` clean on all changed files
- `pytest packages/server/server_tests/...` — 293 passed (including new OrcaRouter tests)
- L3 live test against `https://api.orcarouter.ai/v1` with model `orcarouter/auto` returned a valid completion (200)

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/docs/install_guide/install_guide.mdx
+++ b/docs/install_guide/install_guide.mdx
@@ -189,6 +189,11 @@ resources:
         model: "llama3"
         api_key: "EMPTY"
         base_url: "http://host.docker.internal:11434/v1"
+    orcarouter_model:
+      provider: orcarouter
+      config:
+        model: "orcarouter/auto"
+        api_key: <ORCAROUTER_API_KEY>
   rerankers:
     my_reranker_id:
       provider: "rrf-hybrid"
@@ -306,6 +311,11 @@ resources:
         model: "llama3"
         api_key: "EMPTY"
         base_url: "http://host.docker.internal:11434/v1"
+    orcarouter_model:
+      provider: orcarouter
+      config:
+        model: "orcarouter/auto"
+        api_key: <ORCAROUTER_API_KEY>
   rerankers:
     my_reranker_id:
       provider: "rrf-hybrid"
--- a/docs/open_source/configuration.mdx
+++ b/docs/open_source/configuration.mdx
@@ -500,6 +500,11 @@ To see a complete example of a potential `cfg.yml` file, check out [GPU-based Sa
             model: "llama3"
             api_key: "EMPTY"
             base_url: "http://host.docker.internal:11434/v1"
+        orcarouter_model:
+          provider: orcarouter
+          config:
+            model: "orcarouter/auto"
+            api_key: <ORCAROUTER_API_KEY>
     ```
 </Tab>
 <Tab title="With Comments">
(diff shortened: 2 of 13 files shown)


malatewang:
Is it really necessary to make the code change to support OrcaRouter? I think the existing openaicompatible configuration should be able to work.

zsanig22-dotcom:
Fair point — you're right that OrcaRouter exposes an OpenAI-compatible endpoint, so it already works today through the existing `openai-chat-completions` provider with a custom `base_url`, exactly like the Ollama sample:

```yaml
my_model:
  provider: openai-chat-completions
  config:
    model: "orcarouter/auto"
    api_key: <ORCAROUTER_API_KEY>
    base_url: "https://api.orcarouter.ai/v1"
```

The build path in this PR is functionally identical to the chat-completions builder; the new `orcarouter` provider doesn't change how the model is constructed or called. What it adds is only ergonomics and surface:

1. Defaults (`model: orcarouter/auto`, `base_url: https://api.orcarouter.ai/v1`) so users don't have to copy an endpoint or model id.
2. An explicit entry in the provider enum, so the runtime add-model API schema (`AddLanguageModelSpec`) and OpenAPI doc accept and validate `orcarouter` instead of treating it as an anonymous custom base URL.
3. Self-documenting sample configs.

So the code change isn't strictly necessary for OrcaRouter to function. If you'd prefer not to grow the provider registry for that benefit, I can reduce the PR to a docs + sample-config-only change (showing the `openai-chat-completions` + `base_url` snippet above) and drop the code paths, or close it entirely. Happy to go either way — let me know your preference.

PR #1578, opened 2026-09-02
fix(vector-store): create payload indexes even when the collection exists

wanghy73:
## The defect

`_create_native_collection` wrapped `create_collection` **and every
`create_payload_index`** in one `try`, then swallowed "already exists" for the whole
block. A collection that already existed therefore raised on the first call, took the
already-exists path, and was left with **no payload indexes at all** — despite the
docstring promising both were created idempotently.

Two creators are easy to arrive at. The guarding lock is keyed on the
`AsyncQdrantClient` object:

```python
_name_locks: ClassVar[WeakKeyDictionary[AsyncQdrantClient, defaultdict[tuple[str, str], asyncio.Lock]]]
```

so it serialises callers inside one process and nothing across them. With
`MEMMACHINE_WORKERS` above 1 each worker has its own client and its own lock. A crash
between the two calls leaves the same state.

## The change

The collection and the indexes now sit under separate guards, and each index is created
individually and tolerant of already-exists.

## Evidence

Against Qdrant 1.19 in testcontainers.

**Before** — `test_indexes_are_created_when_the_collection_already_exists` fails, and not
with one index missing:

```
AssertionError: the tenant partition index is missing: a collection that already
existed never had its payload indexes created ... present: []
assert 'sys-partition_key' in set()
```

`payload_schema` is empty — none of the twelve.

**After** — both new tests pass, along with the rest of the suite: **293 unit, 142
integration**. ruff and ruff format clean.

## Scope, checked rather than assumed

A filtered query on an unindexed collection returns only the matching tenant's points, so
what a missing `sys-partition_key` index costs is the **multitenant storage layout and
query speed, not isolation**. I verified this directly rather than reasoning from
`is_tenant=True`.

Only the already-exists path reproduces. Two clients creating *simultaneously* both
succeed, which the second test pins.

## What a reviewer needs to know

These tests are marked `integration` and need a real server — local-mode Qdrant ignores
payload indexes, so the defect is invisible there. **CI will not exercise them**, since
`addopts = ["-m", "not integration"]`. To run them:

```
pytest packages/server/server_tests/memmachine_server/common/vector_store/test_qdrant_vector_store.py \
  -m integration -k TestCollectionLifecycleAcrossWorkers
```

Docker required; the fixture starts its own Qdrant.

## Not affected

The mm-tb6 benchmark collection was checked and holds all twelve payload indexes
including `sys-partition_key`, covering every point — no measurement ran against an
unindexed collection.

### DIFF
--- a/packages/server/server_tests/memmachine_server/common/vector_store/test_qdrant_vector_store.py
+++ b/packages/server/server_tests/memmachine_server/common/vector_store/test_qdrant_vector_store.py
@@ -1,5 +1,6 @@
 """Tests for QdrantVectorStore."""
 
+import asyncio
 import math
 from datetime import UTC, datetime, timedelta, timezone
 from unittest.mock import MagicMock
@@ -7,7 +8,7 @@ from uuid import UUID, uuid4
 
 import pytest
 import pytest_asyncio
-from qdrant_client import AsyncQdrantClient
+from qdrant_client import AsyncQdrantClient, models
 
 from memmachine_server.common.data_types import PropertyValue, SimilarityMetric
 from memmachine_server.common.filter.filter_parser import (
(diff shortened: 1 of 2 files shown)


edwinyyyu (approved):
Heads up: The implementation needs to change completely to properly support horizontal scalability without sharding. Also having dynamic schema/index creation is not good, so that should likely change as well.

Issue #1577, opened 2026-09-02
Deleting a project that is in use returns 204 and then never completes: the deletion worker gives up on SessionInUseError, leaves a partial deletion, and only a restart retries


edwinyyyu:
### What happened

Deleting a project while clients are still using it returns 204 and then never completes. The background deletion worker refuses the episodic half because the cached memory instance is in use, logs the failure, and drops the job; nothing retries it. The project stays in `delete` status: it keeps serving reads and writes with 200 from the process-local instance cache, its segment-store and vector-store data stays on disk, and a later `POST /projects/delete` is a no-op because the status is already `Deleted`. Only a server restart, which re-queues deleted sessions at boot, completes the deletion.

The deletion is also partial by then. The worker gathers the episode-store delete and the episodic-memory delete concurrently, so the episode-store rows are gone before the episodic branch raises. Searches on the stuck project then return fewer results and log "search_scored dropped N episode(s) found in the event index but missing from EpisodeStorage (likely index/storage drift)".

Reproduced on the event backend with SQLite and again with PostgreSQL: four concurrent clients alternating `POST /memories` and `POST /memories/search` on one project, and one `POST /projects/delete` issued 150 ms in. Every one of the 120 client requests returned 200, the delete returned 204, and the server logged:

```
[ERROR] memmachine_server.main.memmachine - Failed to delete session manual_org/race_proj
  ...
memmachine_server.common.errors.SessionInUseError: Session 'manual_org/race_proj' is in use and can't be deleted. Reference count: 1.
```

Afterwards the `sessions` row read `('manual_org/race_proj', 'delete')`, the project's 65 segment rows and vectors were still present, a search on it returned 200 with results, and the drift warnings above appeared on every search. After a restart the boot-time re-queue deleted it ("Deleted session manual_org/race_proj").

Line numbers as of `231ce171` (main):

- `MemMachine.delete_session` (`packages/server/src/memmachine_server/main/memmachine.py:582`) marks the session `Deleted` and enqueues it; it returns early on a session already in that status (line 601), so the API cannot re-trigger a failed deletion.
- `_delete_session_worker` (line 342) catches any exception from `_delete_queued_session`, logs it, and moves on: no retry, no backoff.
- `_delete_queued_session` (line 357) gathers `_delete_session_episode_store` (line 369) and `_delete_session_episodic_memory` (line 386) concurrently, so the episode rows are deleted whether or not the episodic branch succeeds.
[...]

Issue #1576, opened 2026-09-02
A search or add on a nonexistent project creates it (session row, partition and collection), so deletion is not final and the search 404 path is unreachable


edwinyyyu:
### What happened

A search or add on a project that does not exist creates it. `POST /memories/search` for a never-created `(org_id, project_id)` returns 200 with empty results, and afterwards the project exists: a `sessions` row with status `active`, a segment-store partition, and a vector-store collection have all been created. `POST /projects/get` then returns 200 for it.

Observed sequence (event backend, SQLite; PostgreSQL behaves the same):

```
POST /memories/search {"org_id": "manual_org", "project_id": "never_made", "query": "x", "top_k": 1, "types": ["episodic"]}
-> 200 {"status":0,"content":{"episodic_memory":{"long_term_memory":{"episodes":[]}, ...
sessions table before: [('manual_org/keep_me', 'active')]
sessions table after:  [('manual_org/keep_me', 'active'), ('manual_org/never_made', 'active')]
segment_store_pt rows: 1 -> 2
```

The same applies after a deletion: once the deletion worker has removed a project, the next stray search or add recreates it under the same ids.

Cause, line numbers as of `231ce171` (main): both `MemMachine.add_episodes` (`packages/server/src/memmachine_server/main/memmachine.py:714`) and the episodic search path (line 773) enter `episodic_memory_manager.open_or_create_episodic_memory` (`episodic_memory/episodic_memory_manager.py:225`), whose missing-session branch (line 270) calls `create_new_session` and builds the memory, which in turn opens or creates the partition and collection through the event backend's service locator (`long_term_memory/service_locator.py:125` and `:140`).

The search handler was written expecting the opposite: `search_memories` (`server/api_v2/router.py:318`) maps a `RuntimeError` containing "No session info found for session" to a 404 "Project does not exist", but that branch is unreachable for a missing project because the session is created before any such error could occur.

### Expected

Reads and writes on a nonexistent project should fail with 404 (the mapping the search handler already tries to apply), and only `POST /projects` should create one. If implicit creation on add is intended, it should not extend to search, and it should not resurrect a project that was explicitly deleted.

### Notes

Found by a manual end-to-end run against the server. Beyond the surprise for API users, implicit creation makes project deletion non-final: any client still issuing searches after a delete brings the project back, with fresh empty storage under the same ids.

*Investigated and written by Claude (Claude Code), filed from the account of the user who commissioned the investigation.*

Issue #1575, opened 2026-09-02
Project deletion never completes when semantic memory is disabled: the deletion worker requests the semantic service unconditionally and gives up


edwinyyyu:
### What happened

With semantic memory disabled in the server config, deleting a project never completes: the session row stays in `delete` status forever, the project cannot be recreated (`POST /projects` returns 409), and a `POST /memories` on it raises `SessionDeletedError`, which is unhandled on that route and reaches the client as a dropped connection.

Config that reproduces it (event backend on SQLite, everything else default):

```yaml
semantic_memory:
  enabled: false
  config_database: profile_storage
```

Flow: create a project, add a few episodic memories (`types: ["episodic"]`), `POST /projects/delete` (returns 204), then `POST /projects` with the same ids. Server log:

```
[ERROR] memmachine_server.main.memmachine - Failed to delete session manual_org/proj_a
  ...
  File ".../main/memmachine.py", line 383, in _delete_session_episode_store
    await self._cleanup_semantic_history(episode_ids)
  File ".../main/memmachine.py", line 1178, in _cleanup_semantic_history
    semantic_service = await self._resources.get_semantic_service()
  ...
memmachine_server.common.errors.ResourceNotReadyError: No database configured for semantic storage.
```

Line numbers as of `231ce171` (main): `_delete_queued_session` at `packages/server/src/memmachine_server/main/memmachine.py:357` gates the semantic delete on `self._conf.semantic_memory.enabled` (line 361), but `_delete_session_episode_store` (line 369) calls `_cleanup_semantic_history` unconditionally (line 383), and that method (line 1168) requests the semantic service unconditionally (line 1177). With semantic memory disabled there is no semantic database, so `SemanticManager.get_semantic_storage` (`common/resource_manager/semantic_manager.py:89`) raises. The worker (`_delete_session_worker`, line 342) logs the exception and drops the job; nothing retries it. Because `delete_session` (line 582) returns early once the status is `Deleted`, a second delete request is a no-op, so the only way out is a server restart, which re-queues deleted sessions at boot (`start`, line 411) and fails the same way.

Note the partial state left behind: the deletion gathers the episode-store delete, the episodic-memory delete and (when enabled) the semantic delete concurrently, so by the time the episode-store branch raises, the episodic branch may already have dropped the vector collection and segment partition. The project is then neither deleted nor intact.

### Expected

[...]

Issue #1574, opened 2026-09-02
Tracking: move multitenancy from sharded (one process per collection, per-tenant tables and shard keys) to non-sharded (shared structures served by any process)


edwinyyyu:
### Why this issue exists

The issues linked below were filed one at a time, most of them as bugs. Read individually they look like an accumulation of unrelated defects in the storage layer. Read together they are one change of model: the codebase was built for **sharded** horizontal scaling (each logical collection or partition owned by exactly one process, and separated from other tenants by a physical structure of its own), and the requirement is now **non-sharded** scaling (any process serves any tenant, and tenants are rows or values inside structures they share). This issue states that change once, records why the previous model was there, classifies each linked issue against it, and tracks what remains.

### The model the codebase was built on, and why

Two decisions, both explicit and documented when they were made.

**1. Ownership by process sharding.** #1201 (merged 2026-04-02) set the `VectorStore` contract that is still in `common/vector_store/vector_store.py` today:

> A given logical collection identified by a (namespace, name) pair must be managed by at most one process at a time. The consumer is responsible for sharding names across processes.

Its description states the guarantees plainly: "safe concurrent usage on a single process" and "safe concurrent sharded usage (one node per logical collection) on multiple processes". With ownership fixed per process, lifecycle operations (create, open-or-create, delete) could be read-check-write sequences guarded by per-process `asyncio` locks, and backends without transactions or unique constraints (Qdrant, Milvus) could keep their catalog of logical collections inside the backend itself. The multi-process guarantee was correct under its contract and was never broken: strict sharding always worked. The restriction itself is what became the limitation (#1524, #1525).

[...]

PR #1579, opened 2026-09-02
Design: MemMachine server redesign (tenants, event memory API, configuration, lifecycle, schema management)

edwinyyyu:
## Summary

A design proposal, no code: `design/server_redesign.md` and one specification per component under `design/components/`. It started as a tenant lifecycle design and now covers the server, because the lifecycle cannot be fixed inside the current API, configuration and wiring. Against `speedkick`, which carries the segment store overhaul it refers to (#1548, the merged copy of #1545). Tracking: #1574. Line references are to `speedkick` at 7752e4cb. Settings are named, never given numeric defaults. The premise: this is a clean slate; the first deployment carries no data forward.

## Decisions the document records

- **Tenant and components have separate lifecycles.** A tenant is a name, an id and a state (`active`, `deleting`, `deleted`); it exists the moment its row is inserted and holds nothing by itself. Each component's per-tenant resources (the event store's partition, a memory subsystem's stores) are enabled, configured and disabled on their own through `PUT`/`DELETE /v1/tenants/{id}/components/{name}`, each with its own state, jobs and tombstone; a template is the set of sections a create enables at once. Deleting a tenant disables every component and retires the id. The id is a UUID minted per lifetime and is the physical key in every store (no incarnation anywhere); a deleted tenant's row keeps only the id and `deleted_at` (no former name, so no personal data outlives the tenant in the registry).
- **Tombstones.** A component tombstone is swept in rounds: a round that purges something is followed by another; a round that finds nothing on its first batch stamps `clean_at`; after a retention of at least a day on the database clock a verification round runs, and only if it finds nothing again is the tombstone removed (finding something clears `clean_at` and the rounds start over). The tenant row goes when its last component tombstone does. Every remote client must have a request timeout, which is what makes the retention safe.
- **Reconciler without held locks.** A claim is an atomic state transition on the job row (`pending` to `running`), one job per claim, on its own connection; the hook runs with no database lock held on any row. A `running` job whose claim is older than `reclaim_after` is taken over, which is the one thing time decides in the reconciler and is liveness only: a slow step may run twice, and idempotent hooks plus store fences make the second run a wasted write. Ordering comes from the state machine on the component row, idempotent hooks, strict creates, and the sweep step calling `delete` before it purges, so a resource a late `provision` created is still unlinked and removed. Eligibility is computed at claim from `last_run_at`, `last_outcome` and `attempts` (backoff capped, reset on success; a replay with more log continues at once); `reset_replay` updates existing rows only.
- **Events are the caller's ingestion type**, with `session_id` and `source_id` beside `timestamp` (system fields with server semantics: the one total order is within a session; the source is the filterable identity, its readable name a context part) and a `Context` that is a mapping from part kind to one registered Pydantic part. Blocks are a registered family of kinds; a segment is one block, so a block's kind is a system field of the segment. `Event` and `Context` are reshaped for one reason: to prohibit dynamic index creation. The event store keeps events and a per-tenant log of additions and deletions with commit-ordered positions; every field round-trips, the timestamp with its ingested offset. Ingest responds 202 once durable and the replay job is pending; processing is the `replay` job per memory subsystem, one consumer each, which advances its watermark only after both derived stores hold the batch, and `?wait=` polls the watermarks. The job carries no content and no ids: the log holds them. No rebuild of derived data: a reprocessing is a new ingestion into a new tenant. Log compaction is an operator command.
- **Keys and fencing.** Every store fences on its own registry: the SQL-backed stores on a row beside their data plus a purge queue written in the delete's transaction (a row is live, a queue entry is dropping), so their fence is in-statement and they depend on no second component; the others on a three-state row in the shared key registry, injected as a view scoped to one store. Stores whose data is elsewhere read the row, perform the remote write acknowledged as applied, and read the row again; no lock spans remote I/O. Creates are strict; idempotency is the component's `provision`, which knows the key's provenance. Data consumers hold stateless handles (`EventPartition`, `SegmentPartition`, `VectorCollection`), the store's operations bound to one key with no method taking a key, so a wrong key is unrepresentable past construction.
- **Vector store**: containers are one per embedder the composition built, named by the embedder id, provisioned by the schema command; a native per-tenant object is named by the key's hex; the store rejects undeclared property keys; datetimes are stored as integer microseconds where a backend has no datetime type. A table surveys Qdrant, Milvus, pgvector, Pinecone, S3 Vectors, Weaviate, Chroma and the two SQLite stores; the SQLite stores move to shared tables.
- **Concurrency scope** (`process` < `host` < `cluster`; #1531 is the reference): every resource computes its scope from its constructor arguments, the deployment declares the scope it runs at, and startup refuses anything narrower. The **reconciler is a role**.
- **Composition and settings, five rules.** Fixed topology with pluggable slots chosen by `kind`; composition is one Python function; settings are data, `ServerSettings` the root of the file; scoped views wherever a shared resource serves several holders; three scopes told apart by identity (resources, configured objects, call arguments). `EpisodicMemoryManager` is the resource; it builds one `EpisodicMemory` per request from the tenant's handles, embedder, format and cached segmenter and deriver, and runs the search stages.
- **Search is two stages.** `EpisodicMemory.query` is the vector stage with one limit and one threshold (`min_similarity`, cosine) and returns hits in descending score, each a context window with the matched segment marked; the manager runs the optional reranking stage over the rendered windows with its own candidate count and threshold, so over-fetching is one limit set above another. `SearchOptions` is one model: the tenant's `search` section with every field set, the request's overrides with every field optional. Events are not returned by search. Expansion walks the anchor's session in the one total order (session, timestamp, event position, index, offset), counted in segments, the one unit, and returns the two sides without the anchor, which the caller holds, so a neighborhood survives an anchor that fails the filter (#1498); MCP tools expose no expansion counts to the model.
- **Eviction**, from `agentic_expansion` where it runs in production: with it on, `encode` treats near-duplicate derivatives as a cluster (stored neighbors above a cosine threshold, earlier derivatives of the same batch, itself) and trims a cluster over the target size from the temporal middle, deleting stored derivatives and skipping new ones; segments and events are untouched, so the loss is to search only. A mutable tenant option whose threshold is calibrated per embedder.
- **Properties and filtering.** System fields are typed parameters stored under a reserved `memmachine_` key namespace. User properties are scalar, bounded, immutable, copied verbatim to derived data; filter indexes are declared once per vector store, never dynamically; undeclared keys are filtered in the segment store, whose expression indexes a deployment names in settings and the schema command creates. A filter is a constructed closed-union tree, a JSON object at the API and in MCP. Routing by backend capability: the declared part is evaluated during the search, the undeclared part afterward by the segment store with bounded widening; the store scores nothing by id. Every count is a maximum.
- **API** `/v1/tenants`, `/v1/tenants/{id}/components/{name}`, `/v1/tenants/{id}/events`, `/v1/tenants/{id}/episodic-memory/*`, one error handler, closed error-code set including `component_not_enabled` and `component_not_active`, no tracebacks in responses.
- **Schema management**: Alembic per component with its own version table and an `include_object` limited to its tables; `memmachine schema upgrade` is the only thing that runs DDL that is not tenant-specific, creates the declared property indexes outside its transaction, and provisions containers; `serve` fails only when a database is behind its code, so a previous release may restart after a migration under expand/contract.
- Not wired in: short-term memory, semantic memory, declarative memory and the graph stores, the retrieval agent.

## Component specifications

`design/components/`: one file per component (tenant service with registry, component enablement, jobs, reconciler and tombstone pass; key registry; event store; segment store; vector store; `EpisodicMemory`; `EpisodicMemoryManager`; context; blocks; ingest service; filters and properties; server and settings), each with its API, storage, fencing, settings and the changes required of an existing component; identifiers are typed `UUID` throughout. Race matrices tabulate every concurrent pair on a tenant or component row and on the data path. Every SQL-backed component's schema is given with SQLAlchemy types mapped to PostgreSQL and SQLite, constraints and indexes. Every contract is an ABC; every store is two, the store (lifecycle, and constructing handles) and the handle (data). Types are few and each answers one need (`StoredEvent`, `IngestResult`, `LogEntry`, `SearchHit`, `Tenant`, `SearchOptions`); listings take the last position or name as the cursor. Job kinds are four (`provision`; `delete`, the unlink; `sweep`; `replay`), all defined by the tenant service; `MemorySubsystem` extends `TenantComponent` with `replay` and `watermark`.

One naming scheme for values: settings, tenant configuration (per component row, with options), templates, overrides, defaults, partition configuration, request parameters, job arguments; in identifiers `Settings` and `Config`.

## What must be built first

A section names what the first deployment freezes and so must precede the first tenant (the key scheme, including the physical key inside each store, and the tenant tables with tombstones; the event store with positions; stored-record conventions; a registry row per vector key; the public surface; Alembic from the first migration), what must land before churn, what can follow at any time, and the narrowest first deployment.

## Relation to issues

Listed in the document's "Relation to open issues": #1548, #1530, #1531, #1571, #1575, #1576, #1577, #1572, #1573, #1564, #1565, #1537, #1563, #1535, #1570, #1542, and #1574 as the tracking issue.

## Open questions

Hierarchy (flat proposed; an org-level share if semantic memory returns); event size limits; readable metadata as a `json` block kind or a designed field; retention by age, source or session as a job kind; which vector backends the first release implements; selecting the ungrouped stream in a search; log compaction as a scheduled duty.

https://claude.ai/code/session_01MbYdqGZsuws6Z2WHYfCCR5


edwinyyyu:
Chroma facts for the backend table, measured against **chromadb 1.5.9** (local `EphemeralClient` and `PersistentClient`, plus a `chroma run` server over HTTP). Posting these because the doc marks Chroma's row as partly unverified.

**Rejects a write to a dead tenant: yes, but not classifiably.** Operations route by the collection's UUID, not its name, so a handle obtained before a delete faults afterwards rather than writing into a replacement:

```
[3a] recreated, old_id=UUID('55d92cdf-...') new_id=UUID('3f06c8f0-...') same=False
[3c] stale handle .add()  -> NotFoundError: Collection [55d92cdf-...] does not exist.
[3d] NEW collection count (1 => stale write LANDED in new) -> 0
[4]  after recreating the name, count=0   (old data genuinely gone)
```

Same result on `PersistentClient`, so it is not an ephemeral-client artifact. This is what makes recording the UUID in the ledger row at creation, as the doc already specifies, the right call — the UUID *is* the fence on this backend.

**Duplicate create is rejected, but the error is not typed.** The same condition surfaces differently by transport, and never as the `UniqueConstraintError` the module defines and exports:

```
local binding:  chromadb.errors.InternalError, code 500, "Collection [probe] already exists"
over HTTP:      chromadb.errors.ChromaError,   code 400, "Collection [errtest] already exists"
isinstance UniqueConstraintError = False in both
```

So a store distinguishing already-exists from any other failure has to match the message. Worth noting in the row, since `KeyExistsError` on `create_collection` depends on classifying this.

**Creation is genuinely atomic, not merely rejected.** Eight concurrent `HttpClient`s racing `create_collection` against a server, barrier-synchronised, 25 rounds:

```
create          successes-per-round histogram: {1: 25}
                rounds with >1 collection of that name: 0
                loser errors: {'ChromaError': 175}
get_or_create   successes-per-round histogram: {8: 25}
                rounds where winners saw DIFFERENT ids: 0
```

The uniqueness constraint lives in the persisted sysdb rather than client memory — a second independent `PersistentClient` over the same directory sees the collection and its duplicate create raises.

**Metadata values must be scalars.** A nested dict raises `TypeError: argument 'metadata': Cannot convert Python object to MetadataValue`; a JSON string round-trips unchanged. Relevant if anything beyond the container reference is ever stored on the collection.

**Not verified, so the doc's caveat should stay:** per-collection cost on a single node, and the collection counts Chroma Cloud actually supports. I only exercised correctness and concurrency semantics, not scale.

edwinyyyu:
`5f7264acb8b12b36c3e7bf6078f4afb5729b6927`

Four systematic problems account for most of the list. The design describes two registry models, a three-state key registry and the SQL stores' row-plus-purge-queue, in the vocabulary of the first, so several general statements are unimplementable for the event and segment stores. The reconciler's locking is stated three incompatible ways, and the design's own "no lock across remote I/O" rule is broken by the reconciler itself. The garbage story has open ends the design's own requirement forbids. And the main document was not re-swept after the handle and replay reshapes, so its store sections, wiring and API tables lag the component files.

Design-level: decisions needed
- A1 Two registry models. The SQL stores' registry rows have no state column and their logical delete removes the row. So the fence predicate with AND live (main 1012), the "row present but not live" state for SQL stores (1134), and provision's live/creating/dropping branch with KeyReusedError (798, 1123; manager 67; tenant_service 39) cannot exist for the event and segment stores, where create raises KeyExistsError on a purge-queue entry. The create table's "no row: creates" (1141) is false while that entry stands. purge on a live key raises in the vector store and is undefined in the SQL stores.
- A2 Tombstone prune is not computable. "Purge found nothing on its first batch" (465; tenant_service 255) is recorded in no column: success writes nothing, resets clear nothing. The pass also resets the sweeps and evaluates the predicate in one pass with no stated order. Read literally, no tombstone is ever pruned. ORDER BY swept_at puts never-swept tombstones last on PostgreSQL.
- A3 Reconciler locking. "Every job step holds the tenant row's lock" (535) versus "replay holds only its job row" (701; tenant_service 218), with sweep unlisted, and matrix row 296 calls replay/delete "serialized" by a lock replay never takes. A LIMIT n claim in one transaction cannot coexist with "lock held for the step" and "one connection per executing job" (507, 559). Locking tenant rows in claim order across tenants can deadlock two reconcilers. The requirement that no lock spans remote I/O (48, 284) is violated by the hook running under the job and tenant locks (513, 539). On SQLite, BEGIN IMMEDIATE for the whole step (517) holds the file lock past every store's busy timeout, and the shipped code documents BEGIN IMMEDIATE as impossible on a shared engine while the sketch shares one engine; the claim recurs in segment_store 138 and tenant_service 221.
- A4 Watermark before vectors. The manager advances the watermark in the segment-write transaction (94), but vectors are written after segments (episodic_memory 132). A crash between the two loses that batch's derivatives permanently, since there is no rebuild.
- A5 Replay job gaps. Nothing in the design inserts the replay row (tenant_service 128 says provision, but hooks cannot write the job table). The event store registers as a component with an abstract replay hook and no semantics, and every ingest resets its job. purge_interval after any more outcome delays a replay that has more log. Reset clears neither the last outcome nor attempts, and backoff to the power of attempts is uncapped and never reset on success, so a recovered job waits out its whole failure history. Reset as update or upsert decides whether a deleted tenant gets a resurrected row whose foreign key blocks the prune. Replay after delete removed the per-tenant row is undefined; provision's "insert or update" of that row can reset the watermark to zero. "Each batch is processed by its own process" (manager 140) contradicts the single consumer.
- A6 Provision's version. "The section at the job's version" (520; tenant_service 232) cannot be retrieved: only the current configuration and a counter exist. A provision step that finds deleting marks itself done and then trips the unguarded "last provision done, set active" transition (526).
- A7 Compaction and bootstrap. Nothing calls compact_log, so the log grows forever. A subsystem enabled on an existing tenant gets no provision jobs and its routes answer 409 forever. The switch between bootstrapping from events and replaying the log is unstated. "Contiguous" positions are false after compaction and deletion. read_log resolves an added entry by uuid, so a deleted-and-re-ingested id makes an old entry carry new content.
- A8 Removed component. Delete and sweep jobs are per registered component, so rows under a component dropped from the deployment are never purged; only provider ids are startup-guarded.
- A9 Retention premise. "A stale write is always issued before the delete commits" (396) is wrong: the read precedes, the issue can follow. The bound is read-to-issue delay plus timeout.
- A10 include_events. Manager 42 says SearchDefaults matches query's signature including it, checked at import; query has no such parameter, and no object holds an event store to serve it.
- A11 encode(format_options). A per-call value that decides what is embedded, which rule 5 makes structural; it is in no tenant section and no API body.
- A12 Containers and types. The vector store requires a containers setting with dimensions; the settings example declares none and puts dimensions on the embedder. A datetime-typed indexed property has no sqlite-vec representation; sqlite-vec's epoch seconds break the half-open range boundary. vector_store_pt is never given.
- A13 Expansion and the order. session_ids on expand contradicts "stays in the anchor's session" and is absent from the API body. unit=events has no store support. Cursors are promised with no parameter. A null session id under tuple comparison gives the ungrouped stream empty neighbourhoods. Ties break on a random uuid rather than the commit-ordered position. "The index is per session so a window never leaves it" needs an equality predicate the signature lacks. The segment's timezone offset has no source column in the event row, so it cannot be replayed.
- A14 Search semantics. Allowlist path mechanics, over-fetch size with a reranker, min_score before or after reranking, hit ordering, seed marking, cross-hit duplicates, the extent of include_events, and whether the user filter bounds context windows are all unstated.
- A15 Schema management. The "at head" startup check defeats the expand-and-contract rollout it sits beside. A deployment adding an index online is non-tenant DDL outside the setup step, and CREATE INDEX CONCURRENTLY cannot run under the advisory transaction lock. Autogenerate against partial metadata drops unknown tables unless filtered; chain order within one database is unstated.
- A16 Ingest identity and bounds. Optional id with per-id idempotency means a retry without an id duplicates. No way to filter by event id or select the ungrouped stream. Session and source values "bounded by the property naming contract" means 32 bytes and the key charset. The naming bound, the tenant name bound and blocks.max_bytes have unstated or duplicated enforcement points.
- A17 Chroma. The per-tenant collection name is unspecified and the address is written only at live, so a crash between create and live loses the UUID and the resume and no-row purge paths have nothing to resolve by.
- A18 Undefined types and hooks. MemorySubsystem, StoredEvent, AddResult, DeleteResult, Cursor, Page, Expansion, the request and response models, Tenant, TenantState; two different QueryResult types; IngestResult.position when every id is skipped; list_events order and cursor; no hook reports the applied version or the watermark that Tenant, replay_position and wait_processed need; validate(section) is not in the ABC.
- A19 Settings and states. ServerSettings is both the root model and s.server; its field list omits ingest, filter and context bounds; request_timeout exists at two levels. PATCH and rename on a provisioning tenant are undefined. Which vector backends are implemented is never said.
- A20 Migration paragraph. It contradicts the clean-slate premise and misstates the current code: rows are keyed by incarnation today, vector names are content-addressed, and the session key is hashed only when it fails validation.

Mechanical: contradictions and stale text

- B1 Principle 3 lists the job kinds as provision, delete, purge, replay (269); elsewhere the four are provision, delete, sweep, replay.
- B2 Main 594 still has the subsystem exposing encode(tenant_id, events) to the ingest service, and 600 still has a cache keyed by configuration with "no per-tenant object".
- B3 The ingest service is built before the tenant service (313), holds "the list of subsystems" (333), and the sketch passes them (1426); ingest_service.md takes the tenant service and no subsystems.
- B4 The sketch omits the manager's engine and metrics factory, the tenant service's settings, and the engine for SQL-backed vector stores; two spellings of the segment store class; components "register with" a service they are passed to.
- B5 Inline create runs only with ?wait= in the main document and unconditionally in tenant_service.md.
- B6 Chroma "has no $ne" (970) versus the table; "every backend evaluates membership" versus sqlite-vec; Pinecone and Chroma missing from the filters-during-search list.
- B7 The main document's vector store section is still key-taking with allowed_ids; "key-taking operations behind a stateless handle" (1740); create_partition(key) without its config; the event-store operation list omits read_events_after and returns a tuple.
- B8 What the delete request does to replay rows is stated three ways, twice with the stale term "component-defined job".
- B9 Delete's ?wait= is defined "as for create", which waits for active; five requests take ?wait=, not three; timeout status for ingest is unstated.
- B10 The search body lacks min_score; the expand body lacks session_ids; hit segments lack the uuid expand needs; four spellings of the event id; the tenant section omits include_events.
- B11 KeyRegistry is a Protocol in its code block and an ABC in its prose; two column tables disagree; the main document's method list is incomplete.
- B12 The README error list lacks five raised errors; the handler maps none of InvalidContextError, InvalidBlockError, UnsupportedFilterError, so a caller's bad block is a 500; no code for KeyReusedError or AttemptsExhaustedError; matrix 298 gives 409 while the tombstone row exists.
- B13 Progress "returned by every bounded step" versus purge_deleted_partitions returning a bool.
- B14 Smaller items, listed in the file: "Four rules" for five; "rule 4" and "rule 5" naming two different lists; context.get(Author); session id missing from the Event vocabulary; "both derivers" where one exists; "tombstone sweep" beside "tombstone pass"; the filter example's kind property; a dangling "see Serialization"; two typos; two code references off by a few lines.
- B15 The PR body's open questions still name catch_up and in-request processing versus a queue.

edwinyyyu:
Reducing scope.

edwinyyyu:
Will continue to develop design here.

PR #1582, opened 2026-09-04
Add an addressable timeline: server, MCP tools, and a Claude Code integration

edwinyyyu:
Everything the `claude_memory` prototype needs from MemMachine, so it can
become a thin client with no local process of its own. `memory_demote` and
`memory_annotate` are deliberately out of scope.

The prototype works by importing MemMachine as a library and holding a local
daemon, because the daemon is what keeps the embedder loaded and the stores
open. A server does that already. What was missing was not a place to run —
it was an addressable timeline: a way to name a point inside a stored
conversation and read outward from it. That is what this adds.

## Store and engine

- **Neighbour fetch.** `get_segment_contexts` answers "give me these segments
  with context", which is the wrong shape when you already hold the seed: the
  filter decides which segments are valid *results*, so a seed failing it comes
  back with no context, and the seed is spliced into every window.
  `get_neighbor_segments` / `get_neighbor_events` take the seed as an
  *address* instead — located whether or not it passes, excluded from the
  result, filter constraining only the surroundings. They differ in what the
  budget counts: segments is a flat cost and the only way to read inside one
  long event; events is "so many turns either side", so the length of what is
  in the way does not decide how far the window reaches.
- **Addressing.** `find_segment_uuids_by_prefix` resolves an abbreviation,
  `get_adjacent_segment_uuids` says how short one can be (two index lookups
  instead of a partition scan), and `list_event_headers` reads the timeline's
  shape — position, size, opening segment — without decoding any content.
  Verified on SQLite and PostgreSQL.
- **EventMemory.** Batches are ordered before segmenting; `serialize_encode`
  guards the read-then-write that eviction introduces; eviction bounds how many
  near-duplicates a region of vector space accumulates, dropping from the
  temporal middle so the earliest and latest survive. `query_vector` lets a
  caller supply the search point. `gap_marker` marks an elided stretch inside
  one event, so a filtered or sampled window cannot read as one continuous
  statement nobody made.
- **Three bugs found along the way**, each independently useful: SQLite row_id
  reuse could let a suspended `delete()` erase a live record's vector, leaving
  it fetchable but unsearchable with nothing reporting it; datetime filters
  were compared on wall clock rather than instant, so a local-time range
  excluded rows inside it; an interrupted index save left an unparseable file
  where the store treats that as fatal.
- **turbovec** joins usearch and hnswlib behind an extra — quantized, roughly
  flat instead of a linear float32 scan, and the one engine here whose
  checkpoint survives a power failure.

## Server

`/memories/timeline/{search,expand,outline,resolve}`. `search` ranks points
rather than episodes and hands back an address; `expand` reads around one, with
the seed shown in place because a window whose centre is missing cannot be read
as a timeline; `outline` reports shape without content — it answers *where*,
which neither of the others does. Addresses are the shortest unambiguous
prefix, accepted in that form everywhere, and one that names nothing or several
things is a 404 listing candidates rather than an empty result that reads like
a miss.

`MemMachine.open_timeline` hands the session's long-term memory to a request
for its duration. Search and add each shape one call; the timeline is read by
navigating from one address to the next, and wrapping each step separately
would reopen the session every time while still not letting a caller thread an
address through.

Event backend only, and it says so.

## MCP

`search_timeline`, `expand_timeline`, `outline_timeline`, registered beside the
existing `add_memory` / `search_memory` / `delete_memory` and calling the same
service functions the routes call. Both transports serve them with no second
definition and no loopback request.

## Claude Code integration

A `Stop` hook and a skill, under `integrations/claude-code/`. No daemon and no
per-session process: the tools are already served over HTTP, so Claude Code
connects to a URL. The hook is stdlib-only, because its import cost is paid
every turn and its work is one file read and one POST.

The whole transcript goes on one timeline — messages, tool calls, results,
reasoning — since replaying what happened means reading a contiguous stretch of
it. Only messages are indexed; the rest is reached by expanding, so a pasted
file cannot outrank the sentence that asked about it. Text that arrives with a
user turn without being typed by the user is classified on its shape and kept
out of the index. The compaction summary is the case that earns that rule:
compaction continues the same session, so the summary sits among the very turns
it paraphrases.

## Three questions this answers

**Is the daemon only for running locally?** Yes. It exists to keep the embedder
loaded and the stores open across calls; the prototype's own clients already
hold no state. A server does exactly that, so against a server there is nothing
left for it to do.

**Is any per-session state needed?** Almost none. The capture mark — how much
of the transcript has been sent — is genuinely per-session, and a file holds it
fine; it is written after the post succeeds, so a crash repeats a turn rather
than losing one. The prototype's other state is a "what have I already
surfaced" set, which only ever fed a suppression channel that ships disabled.
Deliberate recall does not need it: you asked for something, so a result that
does not apply is visible as a mismatch. Nothing here carries it.

**Is that set reset at a compaction boundary?** In the prototype, no — and it
is the sharpest reason not to carry it. `DESIGN.md` §5 records the inversion:
after a compaction the turns marked "already shown" are precisely the ones no
longer in context, so the record means the opposite of what it claims. (The
*in-context exclusion filter* is compaction-aware, via `last_compaction_time`;
the seen set is the piece that never was.) If a channel ever needs suppression,
the truncation belongs in a `PreCompact` hook — noted in the README, not built.
The capture mark is unaffected either way: the transcript is append-only and
compaction continues the same file, so a line count stays valid.

**And where do MCP tools belong?** On the server, sharing the service layer —
which is where MemMachine already puts them. A translating shim would need a
second definition of every tool and a loopback request per call, and would
still have to be spawned per session. Serving them over HTTP means the client
spawns nothing at all.

## Verification

1990 passed, 16 skipped. New coverage: neighbour fetch and addressing against
the real store on both SQLite and PostgreSQL, timeline access against a real
event-backed `LongTermMemory`, the HTTP surface including address failures, and
the capture hook's transcript classification. `ruff` and `ruff format` clean;
`ty` reports the same 15 pre-existing optional-extra import errors as
`speedkick` does. `test_version` fails on `speedkick` too and is untouched.

## Not included

The `langchain-text-splitters` → stdlib splitter port from the prototype. It
cuts a heavy transitive dependency but supports no feature here, and vendoring
174 lines is a separate decision.

https://claude.ai/code/session_016En3jxWURvpGvzk1nHPDss


edwinyyyu:
Doing this in one shot was not good.

PR #1593, opened 2026-09-08
Make cosine the only similarity, and add get_cosine_similarity (speedkick)

edwinyyyu:
Stacked on #1591 (which is itself stacked on #1588) — only the last commit, `Make cosine the only similarity, and read it back per record`, belongs to this PR.

Follows the `default` branch, where this removal has already been made.

### Purpose of the change

Every embedder MemMachine ships produces vectors meant to be compared by cosine similarity. `OpenAIEmbedder` hard-coded `COSINE`; `AmazonBedrockEmbedder` defaulted to it; `SentenceTransformerEmbedder` only reported whatever the model declared. `SimilarityMetric` let the rest of the stack ask which of four metrics was in play, and the answer was cosine everywhere — while every layer that touched a score paid for the other three: direction flags threaded through scoring, thresholds applied in whichever direction the metric implied, and a per-backend table in each store mapping the enum onto native metric names.

### Description

**The enum is gone.** Scores are cosine similarities in `[-1, 1]`, higher is better, and the names say so:

| Before | After |
|---|---|
| `QueryMatch.score` | `QueryMatch.cosine_similarity` |
| `SearchMatch.score` | `SearchMatch.cosine_similarity` |
| `query(score_threshold=…)` | `query(min_cosine_similarity=…)` |
| `compute_similarity(a, b, metric)` | `compute_cosine_similarity(a, b)` |
| `VectorStoreCollectionConfig.similarity_metric` | removed |
| `Embedder.similarity_metric` | removed |
| `similarity_gt(metric)` | removed (no callers) |
| `VectorSearchEngineFactory = (int, SimilarityMetric) -> …` | `(int) -> …` |

`min_cosine_similarity` no longer needs a direction to be meaningful — that is the point of the rename, not just tidier spelling.

**And `get_cosine_similarity` arrives.** With one metric there is finally something worth reading back per record, so this adds `VectorStoreCollection.get_cosine_similarity` and `VectorSearchEngine.get_cosine_similarities`. Every backend can serve it, by one of two routes:

- **Keyed vector access** — hnswlib and usearch gather stored vectors and score locally; sqlite-vec pushes it into SQL (`vec_distance_cosine` per rowid, one point query each because vec0 plans `rowid IN (…)` as a full scan); Qdrant and Milvus retrieve by id, which is *exact* where an id-filtered ANN query is not guaranteed to return every matching point.
- **Comprehensive filtered search** — turbovec cannot hand a vector back at all (it stores only TurboQuant codes), so it leans on its filtered search, which stops early only once it already holds `limit` matches; asking for `limit = len(keys)` therefore returns every key present.

Similarities may be computed from quantized stored vectors, so the contract says plainly that they may not match a similarity computed from a fresh embedding — which covers both turbovec's estimate and hnswlib's normalize-on-insert.

### Two consequences beyond the vector store

`default`'s `packages/core` has no vector graph store, so these two have no counterpart there and are worth reviewing on their own terms.

1. **The vector graph stores carried a metric per stored embedding**, as a companion `similarity_metric_for_<name>` property written beside every vector, in both Neo4j and NebulaGraph. That property is gone, and `Node.embeddings` / `Edge.embeddings` are `dict[str, list[float]]` rather than `dict[str, tuple[list[float], SimilarityMetric]]`. Existing graph data keeps the orphaned companion properties; nothing reads them.

2. **NebulaGraph's similarity search is now always exact KNN.** Its vector indexes offer only L2 and IP, and its `cosine()` does not support `APPROXIMATE` — so with cosine as the only similarity, no index it can build serves a query. `use_ann` could never be true and `_create_vector_index_if_not_exists` could only ever take its skip path, so both are removed rather than left as unreachable code. This is a real behavioral narrowing for NebulaGraph deployments that were relying on an L2 or IP index. The now-unused ANN tuning parameters (`ann_index_type`, `ivf_nprobe`, `hnsw_ef_search`, `force_exact_similarity_search`) are deliberately left in place for a separate change, since they are user-visible configuration.

Tests that existed only to pin multi-metric behavior are removed: euclidean/dot search sections in the engine suites, euclidean ordering and threshold-direction tests in the store suites, the `_similarity_metric_to_nebula` mapping table test, and the euclidean threshold-inversion regressions in the event-backend wiring suite (that inversion is no longer reachable).

### Verification

- `pytest packages/server/server_tests`: **1916 passed, 3 skipped**. The one failure, `test_get_version`, is a git-describe version-string artifact of the local checkout and fails identically on untouched `speedkick` (`0.3.9.post2.dev5+g8dd1c0178`); see #1592's sibling discussion.
- `ty check --project packages/server`: 17 diagnostics, the same 17 as the base branch.
- `ruff check` / `ruff format --check`: clean.

---

### DIFF
--- a/packages/server/pyproject.toml
+++ b/packages/server/pyproject.toml
@@ -80,7 +80,7 @@ litellm = [
     "litellm>=1.63.0,<1.86",
 ]
 turbovec = [
-    "turbovec>=0.7.0",
+    "turbovec>=1.0.0",
 ]
 
 [project.scripts]
--- a/packages/server/pyproject.toml
+++ b/packages/server/pyproject.toml
@@ -79,6 +79,9 @@ milvus = [
 litellm = [
     "litellm>=1.63.0,<1.86",
 ]
+turbovec = [
+    "turbovec>=0.7.0",
+]
 
 [project.scripts]
 memmachine-server = "memmachine_server.server.app:main"
--- a/packages/server/server_tests/memmachine_server/common/configuration/test_embedder_conf.py
+++ b/packages/server/server_tests/memmachine_server/common/configuration/test_embedder_conf.py
@@ -9,7 +9,6 @@ from memmachine_server.common.configuration.embedder_conf import (
     AmazonBedrockEmbedderConf,
     OpenAIEmbedderConf,
 )
-from memmachine_server.common.data_types import SimilarityMetric
 
 
 @pytest.fixture
@@ -32,7 +31,6 @@ def aws_embedder_conf() -> dict[str, Any]:
             "aws_access_key_id": "key-id",
             "aws_secret_access_key": "secret-key",
             "model_id": "amazon.titan-embed-text-v2:0",
-            "similarity_metric": "cosine",
         },
     }
 
@@ -77,7 +75,6 @@ def test_valid_aws_bedrock_embedder_config(aws_embedder_conf):
     assert conf.aws_access_key_id == SecretStr("key-id")
     assert conf.aws_secret_access_key == SecretStr("secret-key")
     assert conf.model_id == "amazon.titan-embed-text-v2:0"
-    assert conf.similarity_metric == SimilarityMetric.COSINE
 
 
 def test_valid_ollama_embedder_config(ollama_embedder_conf):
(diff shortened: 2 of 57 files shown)


edwinyyyu:
Superseded by #1598, which combines the cosine-only change with the read-back replacement so the implementation moves rather than being deleted and re-added.

PR #1611, opened 2026-09-10
(Depends on #1617) Context parts, block kinds and kind-keyed segmenter and deriver tables (speedkick)

edwinyyyu:
The second half of `design/event_memory_handoff.md` (branch `design/tenant-lifecycle`, PR #1579), split out of #1597 on 2026-09-10. Stacked on #1597 and, on it, #1617 (eviction): their commits come first, and this PR's are the last four: the rebased original, the block-typing fix, the context class, and the table wording on the memory's parameters. Design: `design/components/blocks.md` ("Processing") and `design/components/context.md` ("Rendering").

## What changes

**Blocks** (`event_memory/data_types.py`)
- `Block` is an ABC with `kind` as the discriminator every registered family uses and `render(options) -> str | None`; the union stays closed over `TextBlock`. The segment row's `block_kind` column reads `block.kind`.
- `Event.blocks` and `Segment.block` are typed `Block`, the ABC: a before-validator decodes encoded data through the registered union, and `SerializeAsAny` serializes the concrete kind, so a handler's typed block assigns without narrowing. (`RegisteredBlock` as the field type left the branch with three type errors in the store and the segmenter, now gone.)

**Context parts** (`event_memory/data_types.py`)
- `Context` is a class built from parts and read by kind (`Context(Author(...))`, `context.get("author")`, `context.with_part(part)`), never `None`; a caller never writes a part's key, so cannot write a wrong one, and two parts of one kind are rejected at construction. `Author` is the first registered kind; an unregistered kind decodes to `UnknownPart` and round-trips unchanged. The class owns its wire form, `{kind: fields}`, through a Pydantic core schema, so a model field of the type accepts an instance or the encoded form and serializes to the encoded form. `ProducerContext`, `NullContext` and the discriminated union are gone.
- **Server** (`long_term_memory.py`): the producer id is an `Author` part beside `source_id`, so rendered segments keep their `producer: text` shape.

**Segmenter and deriver** (`segmenter/`, `deriver/`)
- `Segmenter` and `Deriver` are tables from block kind to handler, built from handlers in order with a later handler replacing an earlier one for its kind. A kind with no handler passes through as one segment, unchanged, and derives nothing; a step never raises on a kind.
- `BlockSegmenter[B]` and `BlockDeriver[B]` are the handler contracts, one kind each (`kind: ClassVar[str]`), typed by the block class: `split(event, block) -> list[Piece]` and `derive(segment, block) -> list[str]`. The table builds every envelope, so a handler decides only the pieces (`Piece(offset, block)`) or the texts.
- `TextSegmenter`, `WholeTextDeriver` and `SentenceTextDeriver` keep their names as `text` handlers. `PassthroughSegmenter` is gone, and so is `type: passthrough`: one segment per block is the identity segmentation, the complete answer for any kind, and a default gets no name. `segmenter` omitted in configuration means it; `segmenter: {type: text, max_chunk_length: N}` selects the splitter.

**Composition** (`event_memory.py`, `formatting.py`)
- A derivative is always text: `Derivative.block` becomes `text`, plus `block_kind` (the segment's) for the vector record. Every derivative gets the same context processing.
- `format_header` (`formatting.py`) is the one composition point for the embedded text and the rendered header: the timestamp, written per a `DateTimeFormat`, then the context parts `parts` names in that order (default `("author",)`), then the content. Parts carry no order of their own; a part not listed contributes nothing; a new kind is placed by listing it. `parts` is a parameter of the composers, not a field of `DateTimeFormat`, which writes a timestamp and nothing else. A `BlockDeriver` handler owns its `datetime_format` and `parts`, so a kind or a handler can embed under its own composition, and the `Deriver` table composes each derivative's text with them; `Derivative.text` is the text to embed. `render_segments` takes the caller's `datetime_format` and `parts` per call. `EventMemoryParams.format_options` is gone.

## Decisions

1. **One kind per handler.** The table has already dispatched, so a multi-kind handler would only re-narrow and grow back the `NotImplementedError` branch this removes; kind and block class are one-to-one, so the handler's `B` gives it a typed block.
2. **Handlers return pieces and texts, not segments and derivatives.** Every envelope field is the event's or the segment's by contract; a handler that built them could only get them wrong.
3. **Derivatives are text; composition lives in the deriver table, with the handler's options.** The text deriver used to rebuild the header itself, which was the one place a block kind and a part kind met. The table composing through one function removes that matrix: a part's contribution is per part kind, a block's content is per block kind, and there is one composition point. The format is the handler's because the handler decides what is embedded and different kinds or handlers may want different formats.
4. **Order is explicit and owned by the format options**, not by the parts; timestamp first and content last are fixed. A handler's `parts` order what it embeds, a caller's order the display.
5. **Unhandled kinds: identity for the segmenter, nothing for the deriver, never an error at ingest.** The deriver's hole is one a kind's registration should close by declaring its deriver or none (`blocks.md`, proposed); raising at ingest would fail a batch on replayed history whose kind the server no longer registers.
6. **Filter translation lives in `event_memory/utils.py`** (moved from `system_filters.py` in #1597's history).
7. **Not in this change:** the kind table and `memmachine.block_kinds` entry points, `UnknownBlock`, base-from-defaults and per-kind tenant options (`blocks.md`, proposed).

## Verification

- `uv run --frozen pytest packages/server/server_tests/memmachine_server/episodic_memory packages/server/server_tests/memmachine_server/common/filter packages/server/server_tests/memmachine_server/common/test_property_keys.py packages/server/server_tests/memmachine_server/common/configuration`: 620 passed (SQLite).
- `... pytest packages/server/server_tests/memmachine_server/episodic_memory/event_memory/segment_store -m integration`: 111 passed (PostgreSQL via testcontainers).
- Rebased 2026-09-11 onto #1597's review round, then again onto its lookup-and-walk split and `QueryHit` (`seed_index`, `cosine_similarity_threshold`, id lists without `None`, the unconfined walk from a seed with no session, `rerank` without `limit` and `min_score`); the same suites rerun at the counts above.
- Known, pre-existing: `TextSegmenter`'s splitter keeps `strip_whitespace=True`, so pieces rebuild a block only up to boundary whitespace; the test asserts that and the one-line fix changes embedded chunk text, so it is a separate decision.


edwinyyyu:
Closed in favor of its main port, #1687, which carries this PR's content at its final tip; the review history stays here. See the "Source and review" section of #1687 for what was reviewed and what was adapted.

edwinyyyu (commented):
`packages/server/src/memmachine_server/episodic_memory/event_memory/segmenter/passthrough_segmenter.py` survives this change with no consumer: the diff removes `PassthroughSegmenterConf`, the `type: passthrough` branch in service_locator.py and test_passthrough_segmenter.py, but not the module, so `git grep PassthroughSegmenter` at this head hits only its own definition. The description says "`PassthroughSegmenter` is gone"; the module should go with it. long_term_memory.py (`_episode_to_event` neighborhood) also still says "under non-passthrough segmenters". Same finding on the port, #1687.

edwinyyyu:
The passthrough module is removed on the port, #1687, in 528f25011, together with the long-term memory comment that named it. This PR stays closed as reviewed.

PR #1613, opened 2026-09-11
(Depends on #1606) Make a vector store one collection, its handles partitions, and its lookup get_partition on both stores (speedkick)

edwinyyyu:
### Purpose of the change

Lands the store-as-collection decisions on both storage components without pulling in the tenant lifecycle. Stacked on #1606 (its commit shows in this diff until it merges); stacked on #1597 through it.

- **One store is one collection.** A `VectorStore` is built for one collection, named at construction with its vector dimensions and its declared `indexed_properties`. The composition root builds one store per cell of the purpose-by-embedder matrix (`long_term_memory__<embedder>`, `semantic_memory__<embedder>`), and the collection name is what keeps two stores over one engine or client apart: it names the native Qdrant and Milvus collections, the SQLite table prefixes, and each store's registry. `namespace`, `VectorStoreCollectionConfig`, the content-addressed native names and the config-mismatch error go.
- **Handles are partitions, and the verb is `get_partition` on both stores.**

  ```
  provision()                                        # idempotent DDL
  create_partition(partition_key)                    # strict
  get_partition(partition_key) -> Partition | None   # None when absent
  delete_partition(partition_key)                    # idempotent
  ```

  `get_partition` reads the registry row and binds; a lookup that does I/O is fine as long as nothing has to be closed, and staleness stays a property of a handle already held, raised by its operations. The segment store's `open_partition` becomes `get_partition`; `open_or_create_partition` and `close_partition` go from both stores, and create-if-absent is the composition root's (`get_or_create_partition` beside the vector store ABC, and its counterpart in the service locator).
- **Each partition records what it was created under** (dimensions and declared schema, in the SQLite registry rows and the Qdrant and Milvus registry points), and `get_partition` raises `VectorStorePartitionSchemaMismatchError` when a store built with others opens it. The SQLite registries are shared tables keyed by collection and partition, so stores of different collections share one engine.
- **`DatabaseManager.get_vector_store(backend, collection=, vector_dimensions=, indexed_properties=)`** builds and caches one store per (backend, collection), provisions and starts it, and refuses the same collection asked for with other dimensions or keys. The "second service refused" guard from #1606 is gone, since a cell is a store by construction.

### What stays with the lifecycle work

- Keys stay `str` under today's identifier contract. The decided `partition(key: UUID)` needs keys minted once and never reused, which the tenant service's tombstone provides; a UUID derived from the session id would be reused across delete and recreate.
- Logical delete, purge and the per-operation registry read. `delete_partition` deletes synchronously as `delete_collection` did.
- The schema command. The composition root calls `provision()` when it builds a store until then.
- Shared tables for the SQLite stores' records; records tables and index files stay per partition, named by collection and key.

### Tests

The four store modules, the segment store module, the resource manager and wiring tests are rewritten to the shape: fixtures build a store per collection and provision it; `create_collection(namespace, name, config)` calls become `create_partition(name)`, and so on. Tests of removed paths go: open-or-create idempotency and config mismatch on both stores, the segment store's lost-race retry arm (that loop lived in the removed method; the composition root's create-if-absent makes one attempt), and the namespace tests. Added: two stores of different collections over one engine are isolated; provisioning creates the payload indexes on a Qdrant collection that already exists and survives two concurrent provisioners (integration); the same collection asked for with other dimensions or keys is refused; collections on one backend share its engine.

`uv run pytest packages/server/server_tests`: 2016 passed, 2 skipped. `ruff check` and `ruff format --check` clean. `ty check packages` reports only the diagnostics already on the base.

https://claude.ai/code/session_01DtbsPQafKpBUn7ksnDU2UY

### DIFF
--- a/docs/open_source/configuration.mdx
+++ b/docs/open_source/configuration.mdx
@@ -345,6 +345,9 @@ To see a complete example of a potential `cfg.yml` file, check out [GPU-based Sa
               # token: $MILVUS_TOKEN
               # db_name: default
               consistency_level: Session
+              request_timeout: 30
+              # indexed_properties:
+              #   category: str
       ```
     </Tab>
     <Tab title="With Comments">
@@ -376,6 +379,9 @@ To see a complete example of a potential `cfg.yml` file, check out [GPU-based Sa
               # token: $MILVUS_TOKEN # Required for Zilliz Cloud or auth-enabled Milvus
               # db_name: default # Optional Milvus database name
               consistency_level: Session # Strong, Session, Bounded, or Eventually
+              request_timeout: 30 # Seconds a request may take; required
+              # indexed_properties: # Property keys every collection indexes, by type name
+              #   category: str
       ```
     </Tab>
     </Tabs>
--- a/packages/server/server_tests/memmachine_server/common/configuration/test_database_conf.py
+++ b/packages/server/server_tests/memmachine_server/common/configuration/test_database_conf.py
@@ -122,6 +122,13 @@ def db_conf_dict() -> dict:
                     "api_key": "test-key",
                     "is_distributed": True,
                     "registry_replication_factor": 3,
+                    "request_timeout": 12.5,
+                    "indexed_properties": {"category": "str"},
+                    "hnsw_config": {"ef_construct": 256, "payload_m": 32},
+                    "optimizers_config": {"default_segment_number": 4},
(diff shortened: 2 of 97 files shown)


edwinyyyu:
Folded into #1606, which now sits directly on speedkick. The old tip is kept as tag pre-reorg-1613 locally.

PR #1615, opened 2026-09-11
(Depends on #1606) Make filter negation the complement on every backend, and name absence IsNull (speedkick)

edwinyyyu:
Stacked on #1606 (`feat/vector-store-handoff-speedkick`, head `c6af2c21`), whose commits this branch includes; review only the top commit. #1606 in turn carries a stale copy of #1597 (commit `8ed2bfca`; #1597's head is now `2f1ae2d1`) which its owner will rebase, so the five pre-existing `ty` diagnostics about `Block` and `EventMemoryParams` come from that stale copy and are untouched here.

### What changes

**`IsMissing` is renamed `IsNull`.** That is the name the textual filter language already uses (`IS NULL`, `IS NOT NULL`) and the name the design settles on. Its docstring now says what it means: the field holds no value, true for a record that does not carry the field, and property values are never null, so a missing key is the only no-value state. No occurrence of `IsMissing` or `is_missing` is left in the tree.

**`NotEquals` is removed.** The parser maps `!=` and `<>` to `Not(Equals(...))`, `NOT IN` to `Not(In(...))` and `IS NOT NULL` to `Not(IsNull(...))`, so all three negations mean one thing. Every compiler arm and every `_SUPPORTED_FILTER_NODES` entry for the node goes with it; the filter union is closed, so `ty` found the stragglers. The module docstring of `filter_expression.py` now states the two-valued rule instead of the sentences distinguishing `NotEquals` from `Not(Equals)`. Short-term memory's value comparison loses its `!=` branch, which only the removed arm reached.

**`sql_filter_util.py`'s `Not` is fixed.** It compiled to `~inner`, which is SQL three-valued logic: `NOT (k = v)` on a row whose `k` is NULL is unknown and the row is dropped, so a negated filter silently excluded rows holding no value. It now compiles to `~func.coalesce(inner, false())`, the same shape `common/vector_store/sql_columns.py` already used. Totalizing at the `Not` node is enough; leaves are left alone. The segment store, the episode store, both SQLAlchemy semantic storages and short-term memory read through this compiler.

What each compiler's `Not` arm does, after this change:

| Compiler | `Not` compiles to | Complement over the record set? |
| --- | --- | --- |
| `common/filter/sql_filter_util.py` (segment store, episode store, pgvector and vector-store semantic storages) | `NOT COALESCE(<operand>, FALSE)` | yes — fixed here |
| `common/vector_store/sql_columns.py` (SQLite, sqlite-vec vector stores) | `NOT COALESCE(<operand>, FALSE)` | yes — already |
| `common/vector_store/milvus_vector_store.py` | De Morgan down to the leaves; each negated leaf is `(<inverse leaf>) \|\| (<field> is null)` | yes — already; Milvus's own `not` and `!=` are both false on a missing field |
| `common/vector_store/qdrant_vector_store.py` | `Filter(must_not=[<operand>])`, which admits points lacking the field | yes — already |
| `episodic_memory/short_term_memory/short_term_memory.py` | Python `not` over the episode predicate, whose leaves are false on a missing field | yes — already |
| `server_tests/.../in_memory_vector_store_collection.py` (test double) | Python `not` | yes — already |
| `common/vector_graph_store/neo4j_vector_graph_store.py` | `NOT (<cypher comparison>)` | no — see "Not done" |
| `common/vector_graph_store/nebula_graph_vector_graph_store.py` | `NOT (<nGQL comparison>)` | no — see "Not done" |
| `semantic_memory/storage/neo4j_semantic_storage.py` | `NOT (<cypher comparison>)` | no — see "Not done" |

**Conformance suite.** `declared_schema_contract.py` replaces its two absence tests with the complement law over a fixture of one record holding every declared key, one record lacking each declared key in turn, and one record holding none. For each of twelve filters — `Equals` on a string, a bool and a datetime, `Ordering` on an int, a float and a datetime, `In`, `IsNull` on a key some records hold and on a key of a record that holds none, an `And`, an `Or`, and a nested `Not` — it asserts that `Not(F)` admits every record minus the ones `F` admits, and that `Not(Not(F))` admits exactly what `F` does. Two direct assertions stay: `IsNull` matches exactly the records without the key, and `Not(Equals(...))` keeps a record without the key. The same law is asserted for `compile_sql_filter` over all three of its encodings (`column`, `json`, `properties_json`), and the segment store gains a context query whose negated filter keeps a segment carrying no property at all — the `->>`-on-PostgreSQL path the `COALESCE` guards. Parser tests assert that `k != 'v'` and `k <> 'v'` yield `Not(Equals(...))`.

### Decisions

The one semantic change relative to #1606 is that **`!=` no longer excludes absence**. On #1606, `k != v` kept only records holding a comparable value other than `v`, while `k NOT IN (v)` and `NOT (k = v)` also kept records holding nothing — the same negation written three ways meant two different things. The rule this follows is the settled one: the grammar is two-valued, a leaf on a field with no value is false, `IsNull` on it is true, and `Not` is the ordinary boolean complement, so `!=` and `NOT IN` are literally `NOT =` and `NOT IN`. A caller who means "present and other than x" now writes `k IS NOT NULL AND k != x`. If you would rather keep the old `!=`, reverting is a self-contained change: restore the `NotEquals` node and its compiler arms and point the parser's `NE` branch back at it; the `sql_filter_util` `Not` fix and the `IsNull` rename stand on their own.

Two smaller calls, both noted because they were judgment rather than instruction: the dead `!=` branch of `ShortTermMemory._compare` is deleted, since the removed `NotEquals` arm was its only caller; and `common/neo4j_utils.render_comparison` keeps its `!=` handling, because three call sites outside the filter compiler still pass it operator strings.

### Verification

Every command below was run from the worktree root as `uv run --frozen --all-extras ...`.

| Command | Result |
| --- | --- |
| `ruff check packages/server` | All checks passed |
| `ruff format --check packages/server` | 358 files already formatted |
| `ty check --project packages/server` | Found 5 diagnostics — the same 5 `invalid-argument-type` errors a detached worktree at `c6af2c21` reports, all from #1606's stale copy of #1597; no new ones |
| `pytest packages/server/server_tests/memmachine_server/common packages/server/server_tests/memmachine_server/episodic_memory` | 1399 passed, 574 deselected |
| `pytest .../episodic_memory/event_memory/segment_store -m integration` | 112 passed, 112 deselected (PostgreSQL via testcontainers) |
| `pytest .../common/filter/test_sql_filter_util.py .../common/vector_store/test_qdrant_vector_store.py -m integration` | 209 passed, 161 deselected (PostgreSQL, and Qdrant over HTTP and gRPC via testcontainers) |
| `pytest packages/server/server_tests` (whole suite, default markers) | 2097 passed, 2 skipped, 1638 deselected |

**Against the unfixed compiler.** The `sql_filter_util` complement assertions were written and run before the `COALESCE` change, with only the `IsMissing` → `IsNull` rename applied: `pytest .../test_sql_filter_util.py -k complement` gave **15 failed, 7 passed**. Every failure was a negation dropping the row that holds no value, for example `assert {'alpha'} == {'alpha', 'epsilon'}` for `Not(Equals(field="tag", value="a"))` over the properties-JSON encoding. The seven that passed are the cases with no absent value to lose (`IsNull` leaves, and leaves on a `NOT NULL` column). After the fix all 22 pass, on SQLite and on PostgreSQL.

**Which backends ran the conformance suite.** The contract mixin is mixed into four vector store test modules, and each ran its 14 negation assertions: SQLite, sqlite-vec, Milvus Lite (behind `importorskip`, installed here), and Qdrant against an in-memory client by default plus a `qdrant/qdrant:v1.17.0` testcontainer over both HTTP and gRPC under `-m integration`. The testcontainer is started by the suite's own session fixture; no container already running on this machine was touched.

**Which it did not.** The in-memory vector store collection is a test double and does not mix in `DeclaredSchemaContract` — contrary to what one might expect from the file's placement — so the complement law does not run against it; its `Not` is Python `not` and its leaves are false on a missing key, so it already satisfies the law, but nothing asserts it. Neo4j, NebulaGraph and the Neo4j semantic storage have no backend in this environment and their filter tests are `-m integration`; they were not executed.

### Not done

The two graph stores and the Neo4j semantic storage get the rename and the `NotEquals` removal only. Their `NOT` compilation is left exactly as it was, because there is no backend here to test a change against.

- `common/vector_graph_store/neo4j_vector_graph_store.py` emits `NOT (<condition>)` around a Cypher comparison such as `n.city = $p`. On a node with no `city`, `null = $p` is null and `NOT null` is null, so the node is dropped: **not the complement**. `NOT (<ref> IS NULL)` is fine, since `IS NULL` is already two-valued.
- `common/vector_graph_store/nebula_graph_vector_graph_store.py` emits `NOT (<condition>)` around an nGQL comparison such as `v.prop = 'x'`, with the same three-valued outcome on a missing property: **not the complement**.
- `semantic_memory/storage/neo4j_semantic_storage.py` emits `NOT ({condition})` around the same kind of Cypher comparison: **not the complement**.

Fixing those means either making each leaf total (`coalesce(<comparison>, false)` in Cypher, or `<condition> OR <ref> IS NULL` under negation) or pushing negation to the leaves the way the Milvus compiler does, and it should land with a run against a real Neo4j and NebulaGraph.


edwinyyyu:
Folded into #1606 (the IsNull rename, Not(Equals) for !=, and the COALESCE fix in sql_filter_util are all in it). The old tip is kept as tag pre-reorg-1615 locally.

PR #1616, opened 2026-09-14
[vector store contract 6/6] Close the filter union, and make negation the complement on every backend

edwinyyyu:
### Purpose of the change

The filter language becomes a closed union: `Equals`, `Ordering`, `In`, `IsNull`, n-ary `And`/`Or`, `Not` (`common/filter/filter_expression.py`); every compiler is an exhaustive `match`. `!=` and `<>` parse as `Not(Equals)`, `NOT IN` as `Not(In)`, `IS NOT NULL` as `Not(IsNull)`, so a negation means one thing. An ordering comparison takes a number or a date: `field > 'string'` and `field > true`, which parsed before, are now rejected by the parser.

A leaf on a field holding no value is false, `IsNull` on it is true, and `Not` is the complement over the record set on every vector store and the segment store. The SQL compilers render it as `NOT COALESCE(x, FALSE)`: `sql_filter_util.py` compiled `~inner`, three-valued, and a negated filter silently dropped rows holding no value on every SQL-backed store. Milvus pushes `Not` to the leaves by De Morgan, since its own `not` excludes entities lacking the field; Qdrant's `must_not` already admits them. The Neo4j and NebulaGraph graph stores and the Neo4j semantic storage keep their `NOT` over a comparison three-valued. A predicate matches only a value of the compared type on the stores that declare their schema; on Qdrant, whose collections are in strict mode from [vector store contract 5/6], the compiler answers a leaf of another type with a filter no point satisfies, since the server would refuse the condition for the field's index. An `In` over no values admits nothing: every compiler renders it as false, so EventMemory's empty id lists keep meaning "nothing", and its complement admits everything.

Last in the stack, because the store work is the priority and this change touches many files mechanically; the slices under it are written against the nodes `main` has today. The complement law is checked over thirteen filters in `declared_schema_contract.py`, across the column, JSON and properties encodings of the SQL stores.

Adaptation for `main`: the routing of [vector store scale-out 4/5] is rewritten on the closed union (`And.operands`; `Equals` for `Comparison` with `=`); the speedkick-era commit 8ca5f0627 was written on #1597's tree, and its rewrites of #1597's `_system_predicates`, `_lift_typed_filters` and their tests have nothing to apply to on `main`. The event memory test spells the field `context.producer`, `main`'s name for it. Same 60 files as the origin.

### Stack

| # | PR | change |
|---|---|---|

| # | PR | change |
|---|---|---|
| [vector store contract 6/6] | #1616 **(this PR)** | Close the filter union, and make negation the complement on every backend |

This PR's own change is its one commit, `545ec401e`; the rest of its diff is the PRs under it. Stacked on #1628.

### Verification

At every commit from [sqlite store fixes 1/7] (`d7a6042b2`) to the tip `545ec401e`, on 2026-09-21: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`); 2085 tests at this PR's head `545ec401e`. Those commits were verified before the re-order that moved #1460 into the stage, which left their trees unchanged (`git diff` against the verified commits is empty). The 22 commits below `d7a6042b2` (the scale-out stage and the independent PRs' copies in its history), whose trees the re-order changed only by the absence of #1460's files, are being re-verified as this is written; this paragraph is replaced with the result.


packages/server/src/memmachine_server/common/filter/filter_parser.py
```diff
@@ -296,6 +224,16 @@ def _parse_in_value(self) -> int | str:
             )
         return value
 
+    def _parse_ordered(self) -> OrderedValue:
+        """Parse the value of an ordering comparison: a number or a date."""
+        value = self._parse_value()
+        if isinstance(value, bool) or not isinstance(value, int | float | datetime):
```
edwinyyyu, line 230: New rejection: `field > 'string'` and `field > true` parsed before this change and now raise `FilterParseError`. Fine either way, but it is an API change that needs a sentence in the body and a parser test.

edwinyyyu (commented):
One API change the body does not mention and no test covers (line comment).

PR #1617, opened 2026-09-14
(Depends on #1597) Evict near-duplicate derivatives at ingest (speedkick)

edwinyyyu:
The eviction part of `design/event_memory_handoff.md` (branch `design/tenant-lifecycle`, PR #1579), split out of #1597 for review. Stacked on #1597: its commits come first, and this PR is the last commit. #1611 is stacked on this one.

## What changes

**Options** (`event_memory/data_types.py`, `EventMemoryParams`)
- `EvictionOptions(cosine_similarity_threshold, search_limit, target_size)`: the cosine similarity at or above which eviction is considered, how many stored derivatives at or above it are fetched per new derivative, and how many of a new derivative and those stored ones are kept when there are more. `EventMemoryParams.eviction: EvictionOptions | None`; `None` keeps every derivative and issues no query.

**Encode** (`event_memory.py`), after embedding and before any write:
- The batch is sorted by `(timestamp, uuid)` first, so what a batch evicts among its own derivatives is what serial ingestion would.
- `_compute_batch_predecessors`: for each derivative, the earlier batch indices whose cosine similarity is at or above the threshold, from the batch's own embeddings.
- One batched neighbor query with `min_cosine_similarity=cosine_similarity_threshold` and `limit=search_limit`; the stored neighbors' timestamps are read from their segments (`get_segment_uuids_by_derivative_uuids`, then `get_segments`), since on #1598's contract the vector store answers uuids and scores only. A neighbor whose segment is gone is not a member.
- `_select_eviction_targets`: per derivative, the members are its stored neighbors not already displaced in this batch, its batch predecessors not already skipped, and itself; over `target_size`, they are sorted by timestamp and the temporal middle is trimmed, keeping the earliest `target_size // 2` and the latest remainder. Stored members in the middle are displaced, batch members skipped.
- Displaced records are deleted from the vector store and unlinked with `delete_derivatives`; skipped derivatives are never written to either store.

**Store** (`segment_store.py`, `sqlalchemy_segment_store.py`, the in-memory fake)
- `delete_derivatives(derivative_uuids)`: removes link rows by derivative uuid and leaves the segments. Locks the partition for write and, on PostgreSQL, the link rows in uuid order, the same order as `delete_segments`.

## Decisions

1. **Two indexed reads per batch for timestamps.** The vector store returns no properties, so a stored neighbor's timestamp comes from its segment. A timestamp-only read can follow if ingest profiling asks for it.
2. **The `serialize_encode` lock** from the branch is not ported, as #1597 says.

## Verification

- `uv run --frozen pytest packages/server/server_tests/memmachine_server/episodic_memory packages/server/server_tests/memmachine_server/common/filter packages/server/server_tests/memmachine_server/common/test_property_keys.py`: 480 passed (SQLite; 470 on #1597 alone).
- `... pytest packages/server/server_tests/memmachine_server/episodic_memory/event_memory/segment_store -m integration`: 113 passed (PostgreSQL via testcontainers; 110 on #1597 alone).
- This commit is the exact reverse of #1597's "Move eviction to its own change", so the tree at its head is the tree #1597 had before the split.


edwinyyyu:
Closed in favor of its main port, #1685, which carries this PR's content at its final tip; the review history stays here. See the "Source and review" section of #1685 for what was reviewed and what was adapted.

PR #1618, opened 2026-09-14
[vector store contract 4/6] Let a deployment tune a Qdrant collection's HNSW, optimizers and quantization

edwinyyyu:
### Purpose of the change

`QdrantConf` gains `hnsw_config`, `optimizers_config` and `quantization_config`, plain mappings mirroring qdrant-client's `HnswConfigDiff`, `OptimizersConfigDiff` and `QuantizationConfig`, so qdrant-client stays optional for configuration parsing; the store's params validate them against qdrant's own models. They apply to the store's one collection (its registry is SQL tables since #1631). The sample configurations and the configuration docs' parameter table carry the three keys.

`m` must be 0 or unset: the collection is multi-tenant and disables the global graph in favor of per-partition payload indexing, so a deployment tunes `payload_m`, which defaults to 16 as before.

Two commits: the options, and the docs that describe them, the parameter table rows and the sample configurations' comments. Adaptation for `main`: the options land on the one-collection store of [vector store scale-out 5/5], whose tests build their store through the module's `_params` helper, and the collection they apply to is the store's one native collection rather than a registry-minted one.

### Stack

23 PRs on `main`: 5 independent ones, and three consecutive stages numbered on their own: the vector store scale-out, the SQLite store fixes and the vector store contract changes. A stacked PR's diff on GitHub is cumulative until the PRs under it merge.

**Independent PRs**, each directly on `main`; review and merge in any order:

| # | PR | change |
|---|---|---|
| — | #1541 | fix(event-backend): make expand_context return timeline-neighbor episodes |
| — | #1661 | Overhaul segment store: shared tables with incarnation-scoped tenant keys (port of #1548) |
| — | #1630 | Bound every request to a remote vector store by a configured timeout |
| — | #1682 | Hand the Qdrant vector store a metrics factory, so its tracker emits (port of #1532) |
| — | #1624 | Make no memory request create a project |

**The vector store stages**, consecutive: each stacked on the one below. [vector store scale-out 1/5] is directly on `main` and 2/5 on it alone; they do not depend on the independent PRs. From 3/5 up, the chain's history also carries the independent PRs' commits beneath it, since the later stages were written on top of them ([vector store contract 2/6] on #1624's session policy, for one); so a merge of an independent PR rebases the chain without conflict, and until they merge those PRs' changes show in a stacked PR's diff.

| # | PR | change |
|---|---|---|
| | | *Stage 1, the vector store scale-out: what horizontal scalability without sharding requires.* |
[...]

PR #1621, opened 2026-09-14
[vector store 2/12] Route system filters to the vector store and user-defined ones to the segment store (speedkick)

edwinyyyu:
### Purpose of the change

EventMemory took one filter tree and split it by the store's declared keys, so what the vector store evaluated depended on the store's schema: a user-defined key a deployment declared was evaluated in the vector store, against the decision that user-defined properties are never indexed there.

EventMemory.query now takes two filters. `system_filter` names system fields by their bare names (`timestamp`, and the fields its caller writes under the `_`-prefixed layout); it is evaluated during the vector search and again by the segment store. `property_filter` names user-defined properties (`m.<key>`) only; the vector store never holds a user-defined property, so it is applied by the segment store to the seeds the one fetch returned, conjoined with the system filter. A bare name in `property_filter` or a `m.<key>` in `system_filter` is refused, and a vector record carries the system properties and never a user-defined one.

The API still carries one filter, so the long-term-memory adapter takes it apart by field namespace with `split_conjuncts` and refuses a clause that mixes the two under one disjunction or negation. EventMemory itself splits nothing.

### Stack

Slice 2 of 12, every PR targeting `speedkick`; merge bottom-up.

| # | PR | change |
|---|---|---|
| 2 | #1621 **(this PR)** | Route system filters to the vector store and user-defined ones to the segment store |
| 3 | #1622 | Create a session's storage with the session, never on a request |
| 4 | #1623 | Make the examples that write to a project create it first |
| 5 | #1624 | Make no memory request create a project |
| 6 | #1625 | Remove open-or-create and close from both stores |
| 7 | #1626 | Rename logical collection to partition, and open to get, on both stores |
| 8 | #1627 | Make a vector store one collection, with string-keyed partitions |
| 9 | #1628 | Make a vector store filter only on the properties it declares |
| 10 | #1630 | Bound every request to a remote vector store by a configured timeout |
| 11 | #1618 | Let a deployment tune a Qdrant collection's HNSW, optimizers and quantization |
| 12 | #1616 | Close the filter union, and make negation the complement on every backend |

This PR's own change is its four commits: `eaf0133f` (the routing, first cut), then on review `34fa10de` (folds the one-field `FilterOptions` into a plain field), `27724f6d` (removes the widening loop and the field: one fetch of `vector_search_limit`) and `07b0b218` (6 files changed, 240 insertions(+), 89 deletions(-); EventMemory takes `system_filter` and `property_filter` apart, evaluates only the former in the vector store, and the adapter splits the API's one filter by namespace); it sits directly on `speedkick`, and #1622 is stacked on it.

### Verification

`ruff check`, `ruff format --check`, `ty check` (two pre-existing spacy diagnostics), `pytest packages/server/server_tests packages/client/client_tests`: 2213 passed, 3 skipped, on the slice-10 tree, 2026-09-14; every slice passed its own run when built.


edwinyyyu:
Superseded by modifying #1597

edwinyyyu:
Closed: #1597 stands in for this slice. Its typed query parameters (since/until/session_ids/source_ids/block_kinds) are the system filters the vector store evaluates, and `property_filter` is the segment store's alone, which is the end state this PR was an intermediate step toward. The stack is renumbered to 13 with #1597 as slice 2.

PR #1624, opened 2026-09-14
Make no memory request create a project

edwinyyyu:
### Purpose of the change

The simple chatbot example, the TypeScript REST demo and the Dify plugin's add-memory tool wrote to a project without creating it, relying on the write to create it. Each now creates its project before its first memory request and accepts 409 as the project already existing. No behavior changes for them; they stop depending on a write creating a project.

Adding memories to, or searching, a project that did not exist created it, with the server's default configuration, without the caller's knowledge. Now only the create-project request creates a project: a write or a search opens the session and answers 404 for an unknown project, as the search endpoint already promised; the manager's open-or-create goes.

Two callers depended on the implicit creation. `org_id` and `project_id` default to `universal`, so the API promises the project `universal/universal`; the server creates it, once, at startup, and leaves one that already exists as it is. The MCP add tool names its own project and has no create-project counterpart, so it creates the project it writes to, once, and says so. The API doc strings and the OpenAPI document say which requests create a project.

A breaking API change on `main`.

Adaptation for `main`: `add_memories` keeps main's default of all memory types (#1555 is not ported); on `main` a session's storage is still created lazily by the first open, which #1622 in the chain moves to creation. The deleted-session test keeps only its open assertion: on `speedkick` it also pinned that re-creating a session being deleted is refused, and #1677's `create_or_validate_session` accepts one whose data matches (#1539, which it ports, refused it); reported on #1677 rather than fixed here.

### Stack

| # | PR | change |
|---|---|---|
| — | #1624 **(this PR)** | Make no memory request create a project |

| # | PR | change |
|---|---|---|

This PR is its 3 commits, `555ef9cde`, `17f12e541`, `5efe2a3a2`, directly on `main`, independent of the others: review and merge it in any order among the independent PRs (a later one touching the same file rebases). The stages' history carries the same change from [vector store scale-out 3/5] up, so their diffs include it until it merges.

### Verification

At every one of its 3 commits, on 2026-09-21: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`), 1921 tests at its head `5efe2a3a2`.


edwinyyyu (commented):
1. The no-implicit-project guarantee has a gap for non-episodic requests (line comment).
2. `memmachine-server --stdio` (app.py) calls `initialize_resource()` directly and never `init_global_memory`, so `ensure_default_project` does not run for that entrypoint; "at startup" covers the HTTP lifespan, mcp_http and mcp_stdio only. Same on the closed speedkick counterpart, #1639.

**packages/server/src/memmachine_server/main/memmachine.py:710** -- The 404 for an unknown project is raised only inside `if MemoryType.Episodic in target_memories`. A write with `types: ["semantic"]` (any request without episodic) to an unknown project is answered 200: `episode_storage.add_episodes` a few lines up has already inserted rows under the unknown session key (no FK to the sessions table), and the semantic manager ingests without a session lookup. Even with episodic included, the episode rows are persisted before this check runs, so the 404 arrives after the write. The description's "a write or a search opens the session and answers 404 for an unknown project" holds only for episodic requests. Same on #1639.

PR #1626, opened 2026-09-14
[vector store 5/10] Rename logical collection to partition, and open to get, on both stores

edwinyyyu:
### Purpose of the change

A vector store's logical collection becomes a partition, the segment store's word for the same thing, and both stores' lookup is get_partition, answering None like a Python get. Identifiers only, produced by the script below with one manual exception: Milvus's `_CLIENT_REQUESTS` entry `create_collection` is the native client method and keeps its name; the (namespace, name) identity, the per-partition config and every docstring are as they were, and the next change gives them their meaning. The native clients' create_collection and delete_collection keep their names.

```sh
set -e
cd "$(git rev-parse --show-toplevel)"
git mv packages/server/server_tests/memmachine_server/common/vector_store/in_memory_vector_store_collection.py \
       packages/server/server_tests/memmachine_server/common/vector_store/in_memory_vector_store_partition.py
git ls-files -z 'packages/server/*.py' | xargs -0 perl -0pi -e '
  s/VectorStoreCollection(?!Config)/VectorStorePartition/g;
  s/in_memory_vector_store_collection/in_memory_vector_store_partition/g;
  s/vector_store_collection(?!_schema|_namespace)/vector_store_partition/g;
  s/open_collection/get_partition/g;
  s/def create_collection\(/def create_partition(/g;
  s/def delete_collection\(/def delete_partition(/g;
  s/\.create_collection\((\s*namespace=)/.create_partition($1/g;
  s/\.delete_collection\((\s*namespace=)/.delete_partition($1/g;
  s/\.create_collection(?=\s*=\s*AsyncMock|\.assert_)/.create_partition/g;
  s/\.delete_collection(?=\s*=\s*AsyncMock|\.assert_)/.delete_partition/g;
  s/"create_collection"/"create_partition"/g;
  s/"delete_collection"/"delete_partition"/g;
  s/only delete_collection is invoked/only delete_partition is invoked/g;
  s/test_delete_collection_/test_delete_partition_/g;
  s/open_partition/get_partition/g;
'
uv run ruff check --fix --quiet packages/server
uv run ruff format --quiet packages/server
```

### Stack

23 PRs on `main`: 7 independent ones directly on `main`, then the vector store chain stacked on them, then the SQLite store fixes stacked on the chain, then the chain's last two slices stacked on the fixes. A stacked PR's diff on GitHub is cumulative until the PRs under it merge.

| # | PR | change |
|---|---|---|
| | | *Independent PRs, each directly on `main`; review and merge in any order:* |
[...]

PR #1625, opened 2026-09-14
[vector store contract 3/6] Remove open-or-create and close from both stores

edwinyyyu:
### Purpose of the change

Nothing calls them since a session's storage is created with the session ([vector store contract 2/6]): `open_or_create_partition` and `close_partition` leave the vector store interface and its four backends, `open_or_create_partition` and `close_partition` leave the segment store interface and its implementation, and the two config-mismatch errors that only open-or-create raised go with them. A store creates on `create_partition`, strictly, and looks up on `get_partition`, answering None; create-if-absent is the owner's, where the key's provenance is known.

Source changes are deletions only. The tests that exercised open-or-create as a fixture use a test-side create-if-absent instead (`_get_or_create_partition(store, key)`: a strict create, then the lookup), and the tests of its own semantics go.

Adaptation for `main`: the deletions land on the one-collection store of [vector store scale-out 5/5], so the removed methods take a partition key; the Qdrant two-workers integration test, which raced two open-or-creates, races two strict creates instead, which the registry's primary key arbitrates, and binds both handles with `get_partition`.

### Stack

23 PRs on `main`: 5 independent ones, and three consecutive stages numbered on their own: the vector store scale-out, the SQLite store fixes and the vector store contract changes. A stacked PR's diff on GitHub is cumulative until the PRs under it merge.

**Independent PRs**, each directly on `main`; review and merge in any order:

| # | PR | change |
|---|---|---|
| — | #1541 | fix(event-backend): make expand_context return timeline-neighbor episodes |
| — | #1661 | Overhaul segment store: shared tables with incarnation-scoped tenant keys (port of #1548) |
| — | #1630 | Bound every request to a remote vector store by a configured timeout |
| — | #1682 | Hand the Qdrant vector store a metrics factory, so its tracker emits (port of #1532) |
| — | #1624 | Make no memory request create a project |

**The vector store stages**, consecutive: each stacked on the one below. [vector store scale-out 1/5] is directly on `main` and 2/5 on it alone; they do not depend on the independent PRs. From 3/5 up, the chain's history also carries the independent PRs' commits beneath it, since the later stages were written on top of them ([vector store contract 2/6] on #1624's session policy, for one); so a merge of an independent PR rebases the chain without conflict, and until they merge those PRs' changes show in a stacked PR's diff.

| # | PR | change |
|---|---|---|
| | | *Stage 1, the vector store scale-out: what horizontal scalability without sharding requires.* |
[...]

PR #1627, opened 2026-09-14
[vector store scale-out 5/5] Make a vector store one collection, with string-keyed partitions

edwinyyyu:
### Purpose of the change

A vector store was a factory of logical collections, each identified by a (namespace, name) pair and created with its own dimensions, metric and schema; Qdrant and Milvus shared one native collection among logical collections of equal configuration, under a name derived from a hash of that configuration, and the registry mapped (namespace, name) to it.

A store is now one collection: `VectorStore(collection, vector_dimensions, similarity_metric, indexed_properties)` names its one native collection (or its tables and index files) at construction, every partition of it shares the collection's dimensions, metric and schema, and `provision()` creates the collection's durable resources idempotently, before `startup`. `create_partition(key)`, `open_or_create_partition(key)`, `get_partition(key)` and `delete_partition(key)` take a string key; a partition is a payload value (Qdrant), a partition-key value (Milvus) or a pair of tables (the SQLite stores) inside the collection, and the registry beside it records what each partition was created under, so a store built with other dimensions, another metric or another schema raises VectorStorePartitionSchemaMismatchError instead of reading columns and vectors that are not there. Collection names may be 64 bytes; the hash-derived native names go, and with them `VectorStoreCollectionConfig` and the per-partition config. The SQL registry of #1631 is keyed by (backend, collection, partition key) and stores the partition's schema.

`DatabaseManager.get_vector_store(backend, collection=, vector_dimensions=, similarity_metric=, indexed_properties=)` builds and caches one store per (backend, collection), provisioning it until a schema command exists; asking for a collection again with other dimensions, another metric or other keys is a configuration error. The event backend's collection is `long_term_memory__<embedder>` and the semantic memory's `semantic_memory__<embedder>`, one cell of the purpose-by-embedder matrix each; the two SQLite stores of one backend share its engine. The event backend still opens a session's partition lazily (`open_or_create_partition`), and the registry arbitrates the race between workers.

The SQLite stores change shape only: their registry tables become `vector_store_sqlite_pt` and `vector_store_sqlite_vec_pt`, keyed by collection and partition key, the pending-operation log is keyed the same way, and per-partition table names embed the collection. No migration is provided; existing SQLite vector data is orphaned. Their data path, the pending-operation protocol and the index files are as they were; the SQLite fixes stacked above rewrite the write path.

The data path is as it was: a partition stores every property of a record and filters on any key, with the declared keys indexed.

The rename of #1626 (logical collection to partition, open to get) is folded in as the first commit: the database schema and the store's semantics change here, so the rename is part of the redesign rather than a separate PR. The similarity metric stays a per-store parameter and open-or-create stays: the cosine-only change is [vector store contract 1/6] and the open-or-create removal [vector store contract 3/6], both stacked above. Three follow-up commits: the store tests' parameter helpers are typed so ty accepts their overrides; the sweeper-pinning test keys its task by (backend, collection); the backend configuration describes the registry as the partition registry, as the docs do.

### Stack

| # | PR | change |
|---|---|---|

| # | PR | change |
|---|---|---|
| [vector store scale-out 5/5] | #1627 **(this PR)** | Make a vector store one collection, with string-keyed partitions |

This PR's own change is its 5 commits, `f6293ecdd`, `3e24a9a7e`, `a057790d3`, `442e779b8`, `505ca2361`; the rest of its diff is the PRs under it. Stacked on #1702; #1460 is stacked on it.

### Verification

This PR's commits are being re-verified as this is written: the re-order that moved #1460 into the SQLite stage changed the trees under it only by the absence of #1460's files, and before the re-order every commit had passed `ruff check`, `ruff format --check`, `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`) and the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`). This paragraph is replaced with the result.


edwinyyyu (commented):
Two behavior changes the description does not state.

1. Renames with no migration: the SQLite registry tables (`vector_store_sqlite_cl` -> `vector_store_sqlite_pt`, `vector_store_sqlite_vec_cl` -> `vector_store_sqlite_vec_pt`), per-partition table prefixes that now embed the collection, and Milvus's registry and native names without the `memmachine_` prefix. Existing data is orphaned.
2. `DatabaseManager.build_all` no longer builds SQLite vector stores eagerly, `async_get_sqlite_vector_store`/`async_get_sqlite_vec_vector_store` are removed, and the Qdrant/Milvus client getters no longer build a store, so with no warm path on main nothing builds a vector store before the first session create or request.

Also, "MemMachine warms the event backend's store through the locator" in the Purpose is speedkick-only (#1551 is not on main, as the adaptation sentence itself says), and main's ty at this base is 0.0.81 (#1646), not 0.0.59.

PR #1630, opened 2026-09-14
Bound every request to a remote vector store by a configured timeout

edwinyyyu:
### Purpose of the change

A remote vector store request had no bound on this side: qdrant-client bounds its own at 5 s by default, pymilvus not at all, so a Milvus write could hang a request indefinitely, and a Qdrant deployment could not choose its bound.

`request_timeout_seconds` on `QdrantConf` and `MilvusConf`: an integer number of seconds, default 30, required positive (zero would expire every request on arrival), named for its unit like `max_retry_interval_seconds` on the embedder and language-model configurations. Why the two stores differ in the diff: `AsyncQdrantClient(timeout=)` is a client-wide default, the httpx request timeout on REST and the deadline of every gRPC call, so the one kwarg where the client is built bounds every request `QdrantVectorStore` makes and the store itself does not change (qdrant-client takes an integer and applies `math.ceil` to it). `MilvusClient(timeout=)` bounds connecting and reconnecting only; each data call takes its own `timeout=`, so the Milvus store carries the value and passes it on every request it makes. A `cfg.yml` that predates the option loads unchanged. The sample configurations, the configuration docs' parameter table and their Qdrant and Milvus examples show the option.

### Stack

| # | PR | change |
|---|---|---|
| — | #1630 **(this PR)** | Bound every request to a remote vector store by a configured timeout |

| # | PR | change |
|---|---|---|

This PR is its 6 commits, `f4b5aea9c`, `95dab3075`, `b9f19fe3d`, `6a7be31de`, `874ec4707`, `ce97ac053`, directly on `main`, independent of the others: review and merge it in any order among the independent PRs (a later one touching the same file rebases). The stages' history carries the same change from [vector store scale-out 3/5] up, so their diffs include it until it merges.

### Verification

At every one of its 6 commits, on 2026-09-21: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`), 1916 tests at its head `ce97ac053`.


edwinyyyu (commented):
Two gaps against the description.

1. One Milvus request is still unbounded (line comment below); `test_every_request_carries_the_timeout` never calls `get` on a collection handle, so its `get` spy sees only the registry lookup.
2. Docs: only the Milvus YAML example in docs/open_source/configuration.mdx gains `request_timeout_seconds`. The parameter table in the same file (which has a `config.consistency_level` row), the Milvus example in docs/open_source/databases.mdx, and Qdrant (no example anywhere) do not.

**packages/server/src/memmachine_server/common/vector_store/milvus_vector_store.py:375** -- No `timeout=` on this request; every other client call in the file passes `self._request_timeout_seconds`. pymilvus defaults the per-request timeout to None, so `get` stays unbounded, and the title and body ("every request") say otherwise.

malatewang (approved):

edwinyyyu:
@malatewang These changes (or similar) will get folded into #1631 since a timeout is required for #1631. This was mistakenly taken out of the chain.

edwinyyyu:
Folded into #1631 as its own six commits (`c5c0efc1d`, `e443dac65`, `aedcad290`, `914dcd78b`, `92a38429e`, `7de506272`), since the tombstone retention there only has to exceed the longest a request to the backend can be in flight, which this bound sets. Review the timeout in #1631.

Issue #1563, opened 2026-09-01
Handles held across delete and recreate resurrect records in QdrantVectorStore


edwinyyyu:
### What happened

A `VectorStoreCollection` handle stays usable after its collection is deleted, and its writes reach the collection that replaces it.

`QdrantVectorStore` uses the logical name directly as the tenant discriminator, with nothing distinguishing one incarnation of a name from the next:

```python
partition_key=name,
shard_key=name if self._is_distributed else None,
```

So: open a collection, hold the handle, delete the collection, create one with the same `(namespace, name)`, then write through the old handle. The record carries the same partition key the new collection reads, and is returned by its queries.

This needs no concurrency to reproduce — a purely sequential create / open / delete / create / write sequence is enough.

### Expected

A handle is invalid once its collection is deleted, and a write through one must never become visible in a collection later created with the same namespace and name.

### Fix

Give each creation a fresh incarnation and make the partition key carry it, so records written through a stale handle land under a value nothing resolves to. #1527 does this.

### Notes

Same defect class as #1536, which covers both SQLite-backed vector stores. This is the Qdrant instance of it. #1531 states the invariant as a `VectorStoreCollection` contract.

PR #1631, opened 2026-09-15
[vector store scale-out 2/5] Mint an incarnation per collection life in a SQL-arbitrated registry, so any process may serve any Qdrant or Milvus collection

edwinyyyu:
### Purpose of the change

Any number of server processes can now share one Qdrant or Milvus backend: the `VectorStore` contract's "a collection must be managed by at most one process; the consumer shards names across processes" sentence goes, and each store states its own bound. The SQLite stores keep theirs (one process per collection for the engine-backed store, one node for sqlite-vec).

What the sentence was standing in for: Qdrant and Milvus have no transactions, unique constraints or conditional writes, so the catalog each store kept inside its backend (the per-namespace `__registry` collections) could not arbitrate two processes creating, deleting or reclaiming the same logical collection. Create was a read then a write under a per-process `asyncio` lock, so two workers creating the same name both succeeded (Qdrant's registry upsert last-wins; Milvus's `insert` does not enforce primary-key uniqueness, leaving two live entries); the logical name was the tenant discriminator on every point, so a handle held in one process kept writing into a collection another process had deleted and re-created (#1563, sequential); and delete was a filter-delete by name that a write in flight during it outlived.

A logical collection is still identified to callers by its `(namespace, name)`; inside the store it is identified by an incarnation minted when it is created, and every point carries the incarnation in the payload field (Qdrant) or partition-key field and primary id (Milvus) that carried the name. A collection deleted and re-created under the same name starts empty, its predecessor's points are never seen by it and never reclaimed out from under it, and a handle is bound to one incarnation: once that life is deleted, every operation of the handle raises `VectorStoreCollectionHandleStaleError`, checked before the remote call, to refuse a handle known to be dead, and, for a write, after it, so a write completed under an incarnation that died meanwhile raises instead of reporting success. A read is not checked afterward: the records of a deleted collection stay until a purge round claims its tombstone, so a read in flight when the deletion commits returns a snapshot from before it, never mid-deletion state, which is the answer it would have given had it run a moment earlier. No lock spans the remote call.

The catalog that mints incarnations is a `VectorStoreCollectionRegistry`, an ABC in the new `vector_store/collection_registry/` package that the stores are given in their params. A registry belongs to one vector store, and so to the one vector deployment the store's client reaches: every store on that deployment, in any process, uses it, and no store on another deployment does, since a store reclaims its registry's tombstones through its own client. Its one implementation, `SQLAlchemyVectorStoreCollectionRegistry`, keeps each registry in its own table pair in a relational database, `collection_registry_{vector_store_name}_ct`/`_gc`, so the registries of several vector stores can share one database or live in different ones; the database manager builds one per configured backend, naming the vector store by the backend's key under `resources.databases`. It supports PostgreSQL and SQLite, as the segment store does, and refuses an engine of another dialect at construction with a `ValueError`. Its operations are `register`, `get`, `is_live`, `unregister` and `claim_due`. A registry holds a row per live collection keyed by `(namespace, name)` carrying the incarnation and the configuration the collection was created with (JSONB on PostgreSQL, as the segment store stores its JSON). Creation is native collection first (idempotent, as #1681 left it), registry insert last, so a crash between the two leaves an empty native collection the next creation of the same configuration adopts and never a row whose points have nowhere to go; the primary key arbitrates, so a racing creator on any process gets `AlreadyExists` (`open_or_create_collection` adopts the winner's row; a minted incarnation colliding with a live or queued one is re-minted, up to `_MAX_MINT_ATTEMPTS`, then `VectorStoreAttemptsExhaustedError`). Deletion is one registry transaction: the collection is unreachable when it commits, and its points wait on the queue. The registry's incarnation logic is the segment store's (#1661, the port of #1548) wherever it can be: the mint loop with its bound and its in-transaction locking re-check of the purge queue, the idempotent deletion that queues the tombstone, the claim under `FOR UPDATE SKIP LOCKED`, and the retried read-then-create of open-or-create. It differs where the backend is remote: deletion is `DELETE ... RETURNING` then the queue insert, since the registry has no write fence to pin the row with; a handle is fenced by a liveness read before the remote operation and, for a write, after it, rather than a row lock held across a write; the purge is one tombstone per call, reclaimed by one filter-delete rather than budgeted row batches; and the tombstone outlives its first clean round by the retention, described next, because a remote write can land after the purge.

The new `purge_deleted_collections` is part of the `VectorStore` contract: one bounded round per call on the tombstone that came due first. The round looks for one point under the incarnation in the native collection the tombstone names and deletes by filter when it finds one. The queue entry is the incarnation's tombstone, because a write the registry read as live can land after the purge that followed the deletion: a round that found points leaves the entry due, a round that found none stamps it clean, and a round that finds nothing once `tombstone_retention_seconds` (a day unless configured, per store) has passed since the stamp removes it; until then the incarnation is never re-minted. The queue records only the times of what happened, the deletion (`enqueued_at`) and the last clean round (`clean_at`), both written by the database's `now()`. The retention is applied by the database's own arithmetic when a claim or a removal is decided (`now() - interval` on PostgreSQL; `datetime('now', '-N seconds')` on SQLite, which writes the same text form as `CURRENT_TIMESTAMP`), so a changed retention reaches tombstones already stamped and no client clock enters a decision. A claim is two statements, each a range on its own partial index, named in the segment store's scheme: tombstones no round has found clean, oldest deletion first (`collection_registry_{vector_store_name}_gc__ea` on `enqueued_at WHERE clean_at IS NULL`), and, failing that, tombstones found clean past the retention, oldest round first (`collection_registry_{vector_store_name}_gc__ca` on `clean_at WHERE clean_at IS NOT NULL`), each under `FOR UPDATE SKIP LOCKED` on PostgreSQL so concurrent purgers split a backlog. A round that found nothing applies its outcome only to the entry as the database finds it then: stamped if still unstamped, removed if still past the retention (on SQLite, where the claim holds no row lock, two purgers can claim one tombstone). The retention decides nothing about validity, a stale write is refused by the check after the operation; it only has to exceed, by orders of magnitude, the longest a request can be in flight, which `request_timeout_seconds` bounds. The stores whose deletion reclaims physically (both SQLite stores) return `False`.

Measured on PostgreSQL 16 with 20,000 live collections and 20,000 tombstones waiting out their retention, every statement logged by `auto_explain` over twelve executions (past the prepared statement's switch to a generic plan): every registry statement is an index lookup; the first claim reads 3 buffers, the second 4, the removal 4, and the two claims when nothing is due 1 and 2. The claim this replaces, one statement over an index on `enqueued_at`, read and filtered out 19,950 rows (460 buffers) to reach the first due tombstone. With two sweepers running rounds back to back (about 78 per second) beside 16 interactive workers, interactive throughput and `is_live` p99 are unchanged within run-to-run noise on PostgreSQL. On SQLite, where the sweepers write in the same process, throughput drops 2.5-14% at that rate, as much as with sweepers that only commit a one-row write per round, and not measurably at the resource manager's pace.

The Milvus store lets creations of one native collection take turns within a process: Milvus Lite does not arbitrate two creations racing in one process (the loser fails with `File exists`, not an already-exists error), and every logical collection of a namespace and configuration shares its native collection, so two sessions created at once could fail. `validate_identifier` uses `fullmatch`, so a namespace, collection name or vector store name with a trailing newline is refused, and the event backend's service locator hashes such a session id (#1661 replaces that check with the segment store's validator).

The resource manager starts one sweeper per vector store the first time it hands the store out, calling the purge after a short pause while rounds find something and after the idle interval otherwise; `close()` cancels the sweepers. The event backend's service locator creates a session's collection strictly when it finds none; with creation arbitrated across processes that create can lose to another worker's, and the locator then opens the collection the winner registered. The semantic manager's `open_or_create_collection` needs nothing.

`QdrantConf` and `MilvusConf` gain `collection_registry`, the name of a relational database under `resources.databases`, required (named for the component it backs, as `segment_store` and `feature_store` are), `tombstone_retention_seconds`, and `request_timeout_seconds` (whole seconds, 30 unless configured), which bounds every request to the backend: the Qdrant client's timeout, and a timeout on every Milvus call. The timeout is #1630, folded into this PR as its own six commits, since the purge's safety rests on it; `registry_replication_factor` leaves `QdrantConf` with the registry collections it sized (never documented). The wizard points the registry at its SQLite database, the one relational database it configures for every SQL-backed component; the docs' parameter table and the Milvus examples show the keys, and the sample configurations point the registry at `profile_storage`, the relational database their other components use (renaming that id is #1705).

Deployment consequence: the per-namespace registry collections are no longer read, and existing points carry name-keyed partition values no incarnation resolves, so existing Qdrant and Milvus data is orphaned. No migration; pre-GA.

`collection_lifecycle_contract.py` holds the contract tests both stores mix in: stale handles on every operation (`get` included), empty re-creation, open-or-create adopting the live incarnation, idempotent deletion, purge reclaiming what deletion deferred and sparing live collections, and a write landing under a dead incarnation raising and being reclaimed; run against Qdrant in local, REST and gRPC modes and against Milvus Lite. `collection_registry/test_sqlalchemy_collection_registry.py` runs the registry on SQLite and PostgreSQL, concurrent creators and doubly claimed tombstones included; the cross-worker Qdrant tests use two clients over one registry and pin that a strict create is created once. Both carry the equivalents of the tests #1661 adds for the segment store wherever they share its behavior, each verified to fail with that behavior removed: a claim skipping a tombstone another purger holds, racing deletions queuing one tombstone, a mint re-checking the queue after its insert and re-minting on a live collision, one tombstone per round, a persistent database error surfacing with its cause, a read checking the registry once and a write twice, a failed native creation registering nothing, open-or-create bounded under lost races, and deletion leaving the records for the purge.

Reviewing note: this PR is the horizontal-scaling mechanism on `main`'s vector store as it is, built first on purpose; the redesign PRs that used to sit under it in the chain (one collection per store, declared properties, cosine-only scores, session-owned storage, and the rest) are re-stacked on top of it and are optional contract changes.

### Stack

22 PRs on `main`: 4 independent ones, and three consecutive stages numbered on their own: the vector store scale-out, the SQLite store fixes and the vector store contract changes. A stacked PR's diff on GitHub is cumulative until the PRs under it merge.

| # | PR | change |
|---|---|---|

| # | PR | change |
|---|---|---|
| [vector store scale-out 2/5] | #1631 **(this PR)** | Mint an incarnation per collection life in a SQL-arbitrated registry, so any process may serve any Qdrant or Milvus collection |

This PR's own change is its 45 commits, `b87a682eb`, `9fd93fc30`, `09329d263`, `23ccac06a`, `52413cdef`, `e46ae5173`, `92fd650df`, `0330586de`, `2ba210f5a`, `1ab105b32`, `1940b4554`, `de33f71eb`, `018154f8c`, `87d923b5b`, `16cd848a1`, `635573077`, `e1372f056`, `46ef06286`, `d2373939f`, `00ee89fc2`, `085f5c17e`, `bfa469ae5`, `a757862de`, `f007c9a3f`, `110d69a0c`, `a41b18e6b`, `8d0f7c3a8`, `d5b3bb85c`, `c4ea58572`, `7594e0c10`, `a7189450a`, `987469ce6`, `c5c0efc1d`, `e443dac65`, `aedcad290`, `914dcd78b`, `92a38429e`, `7de506272`, `ee941c431`, `062296c4c`, `4a40f85f0`, `109a5d626`, `a8d474b39`, `1cc7f9d5c`, `6e0e903fb` (the last 41 address the 2026-09-21 and 2026-09-22 reviews and a code review of the result; `c5c0efc1d` through `7de506272` are #1630's six, folded in); the rest of its diff is the PRs under it. Stacked on #1671 alone; #1670 is stacked on it.

### Verification

At every one of its 45 commits (on #1671), on 2026-09-21 and 2026-09-22: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`). The full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`) at `23ccac06a`, `0330586de`, `de33f71eb`, `d2373939f`, `00ee89fc2`, `085f5c17e`, `a757862de`, `f007c9a3f`, `7de506272`, `062296c4c` and its head `6e0e903fb`, 1984 tests at the head. The vector store, resource manager, configuration, installation and config-service tests pass at each of the review commits between them; from `bfa469ae5` on, the long-term memory tests too, with their integration tests: the registry against PostgreSQL, and the stores against Qdrant and Milvus Lite, in test containers. At the head the vector store and resource manager integration tests pass (192). Every test added from `110d69a0c` on was run with the behavior it pins removed, and fails there. The tests write every time they store in the queue by the database, as the code does, so on SQLite a test's queue holds only the text form the code writes. Of the code review's new tests, the two that remain (client not opened on a bad `collection_registry`, the SQLite double claim) fail on the commit before their fix; its test of the refusal after close went with the closed gate in `f007c9a3f`.


docs/open_source/configuration.mdx
```diff
@@ -345,6 +345,7 @@ To see a complete example of a potential `cfg.yml` file, check out [GPU-based Sa
               # token: $MILVUS_TOKEN
               # db_name: default
               consistency_level: Session
+              registry_database: profile_storage
```
edwinyyyu: registry_database should not take profile_storage (or the id "profile_storage" should be renamed), since profile storage refers to semantic memory (it's an old name) -- although if it's the pattern already to reuse profile_storage, then file a GitHub issue for it instead of fixing it in this PR.

Apply this reasoning to all occurrences.

packages/server/src/memmachine_server/installation/configuration_wizard.py
```diff
@@ -494,13 +494,18 @@ def database_conf(self) -> DatabasesConf:
             case self.QDRANT_VECTOR_STORE_ID:
                 # Localhost defaults assume `docker run -p 6333:6333 qdrant/qdrant`
                 # or similar; user can edit cfg.yml to point at a remote Qdrant.
-                databases.qdrant_confs = {self.QDRANT_VECTOR_STORE_ID: QdrantConf()}
+                databases.qdrant_confs = {
+                    self.QDRANT_VECTOR_STORE_ID: QdrantConf(
+                        registry_database=self.SQLITE_DB_ID
```
edwinyyyu: Doesn't make sense to be SQLite unless the configuration wizard is for a single-host deployment. Is Postgres at least configurable? Are the other defaults also SQLite, or Postgres? Go with what the other SQL databases already use.

packages/server/src/memmachine_server/common/errors.py
```diff
@@ -231,6 +231,10 @@ def __init__(self, message: str, resource_name: str | None = None) -> None:
         super().__init__(message)
 
 
+class ResourceManagerClosedError(MemMachineError):
```
edwinyyyu: Naming makes it sound like the manager itself is closed. This is confusing.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
```diff
@@ -148,9 +154,13 @@ class VectorStore(ABC):
...
-    A given logical collection identified by a (namespace, name) pair
-    must be managed by at most one process at a time.
-    The consumer is responsible for sharding names across processes.
+    A logical collection is identified to callers by a (namespace, name)
+    pair and inside the store by an incarnation minted per life of the
+    pair, so nothing written under one life of a name is ever seen by, or
+    reclaimed out from under, another. Which processes may share a store's
+    collections is the store's own contract: QdrantVectorStore and
+    MilvusVectorStore serve any process sharing the backend and the
+    registry database; the SQLite stores state their own bound.
```
edwinyyyu: State the contract without naming specific implementations.

packages/server/src/memmachine_server/common/vector_store/collection_registry/sqlalchemy_collection_registry.py
edwinyyyu, line 1: Make collection registry an ABC. The ABC and its implementations should live in a dedicated directory like the search engines for the SQLite vector store to avoid polluting the top-level vector store directory.

SQL collection registry should be named like the sqlalchemy segment store.

docs/open_source/configuration.mdx
edwinyyyu: Kept `profile_storage` here: reusing it is the existing pattern. In the docs and the samples, `episode_store.database`, `session_manager.database`, `semantic_memory.database`/`config_database` and the event backend's `segment_store` all name it, so the registry follows them, and the rename is filed as #1705 (docs, samples and the tests that assert the id).

The one place the registry did not follow the pattern was the sample configurations, which had it on `sqlite_test`, a SQLite file one node can reach, under the Qdrant they recommend for server deployments; 0330586de points them at `profile_storage` like everything else.

packages/server/src/memmachine_server/installation/configuration_wizard.py
edwinyyyu: It is what the other SQL databases use: the wizard configures one relational database, `sqlite_db` (SQLite, `memmachine.db`), and points every SQL-backed component at it: the episode store, the session manager, semantic memory's `config_database` and the segment store. Postgres is not a wizard choice for any of them; a user edits `cfg.yml` for that, and the registry key moves with the rest. No change here.

packages/server/src/memmachine_server/common/errors.py
edwinyyyu: Renamed to `ResourcesClosedError` in 92fd650df: what `close()` closes is the manager's resources, and the getter refuses after that because a store handed out then would run without its sweeper. The message names the store that was requested.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
edwinyyyu: Done in e46ae5173: the ABC says only that which processes may share a store's collections is the store's own contract, stated on the store.

packages/server/src/memmachine_server/common/vector_store/collection_registry/sqlalchemy_collection_registry.py
edwinyyyu: Done in 52413cdef. `CollectionRegistry` is an ABC in the new `vector_store/collection_registry/` package, laid out like `vector_search_engine/`: `collection_registry.py` holds the ABC with its `RegisteredCollection` and `PurgeClaim` types, `sqlalchemy_collection_registry.py` holds `SQLAlchemyCollectionRegistry`, named like `SQLAlchemySegmentStore`. The stores take `registry: CollectionRegistry` in their params in place of `backend`, `registry_engine` and `tombstone_retention_seconds`, and the database manager builds the SQLAlchemy registry for each configured backend, naming the table pair of the backend's kind there. The registry test moved with the module.

packages/server/src/memmachine_server/common/vector_store/collection_registry/collection_registry.py
```diff
@@ -0,0 +1,165 @@
...
+    when the backend still held points under the incarnation, False when
+    it held none.
+    """
+
+    incarnation: UUID
+    namespace: str
+    name: str
+    config: VectorStoreCollectionConfig
+    clean_at: datetime | None
+    found: bool | None = None
```
edwinyyyu, line 52: Explain the PurgeClaim dataclass. What's the purpose of each field? Why do they belong together?

packages/server/src/memmachine_server/common/vector_store/collection_registry/collection_registry.py
```diff
@@ -0,0 +1,165 @@
...
+    the purge that followed the deletion, so one purge cannot be the last:
+    the entry stays through purge rounds until a round finds nothing, then
+    through a retention, then through one more round that finds nothing
+    again. Only then is it removed, and until then the incarnation is
+    never re-minted.
+    """
+
+    @abstractmethod
+    async def startup(self) -> None:
+        """Ready the registry, idempotently."""
```
edwinyyyu: Is startup meant to be idempotent? That's not part of the contract for other startup methods, such as on SegmentStore.

packages/server/src/memmachine_server/common/vector_store/collection_registry/collection_registry.py
```diff
@@ -0,0 +1,165 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    async def create(
```
edwinyyyu: Instead of create and delete, why not register and unregister?

packages/server/src/memmachine_server/common/vector_store/collection_registry/collection_registry.py
```diff
@@ -0,0 +1,165 @@
...
+        The body purges and sets `found`. A round that found points keeps
+        the entry due, so rounds continue; a round that found none stamps
+        the entry clean the first time and removes it when it is the round
+        due after the retention. A body that raises leaves the entry as it
+        was, for a later claim.
+
+        Returns:
+            AbstractAsyncContextManager[PurgeClaim | None]:
+                The claim, held for the body of the context.
+        """
```
edwinyyyu, line 168: The concept of "due" and how `claim_due` is meant to be used need explanation.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
```diff
@@ -26,6 +26,14 @@ class VectorStoreCollection(ABC):
...
     All data operations are scoped to this logical collection.
 
+    A handle is bound to one life of the collection: once the collection
+    is deleted, every operation of the handle raises
+    VectorStoreCollectionHandleStaleError, and a collection created again
+    under the same (namespace, name) is a new life the handle cannot reach.
+    A read already in flight when the collection is deleted may instead
+    return what it read, the collection's content from before the deletion.
+    A store that cannot detect a stale handle says so in its own contract.
+
```
edwinyyyu: Can this be more concise? Would it be simpler to explain this in terms of interpreting it as a concurrent request? Use the appropriate vocabulary.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
```diff
@@ -267,8 +282,10 @@ async def delete_collection(self, *, namespace: str, name: str) -> None:
         """
         Delete a logical collection from the vector store.
 
-        This will delete all data in the collection.
-        It is idempotent.
+        The collection is unreachable when this returns: opening it finds
+        nothing, and handles bound to it are stale. Its records are gone
+        with it or, on a store that reclaims them afterward, left for
+        `purge_deleted_collections`. It is idempotent.
```
edwinyyyu: Surely this can be more concise?

packages/server/src/memmachine_server/common/configuration/database_conf.py
```diff
@@ -252,20 +252,33 @@ class QdrantConf(YamlSerializableMixin, ApiKeyMixin):
...
         description=(
-            "Whether the Qdrant cluster is running in distributed mode. "
-            "If True, native collections use custom sharding."
+            "The relational database (a name under resources.databases) that "
+            "holds this backend's collection registry. Required: Qdrant cannot "
+            "arbitrate collection creation or deletion across server processes; "
+            "the registry lives where a primary key and a transaction can. Each "
+            "backend has its own registry, named by the backend's key under "
+            "resources.databases, which must match [a-z0-9_]+ and be at most 32 "
+            "bytes."
```
edwinyyyu: Does not require explanation for why the registry is required. That belongs in the PR description or a design doc.

Also it's unclear what "Each backend has its own registry, named by the backend's key" means from this description. Need more context; or if the context does not belong here, that indicates this description is not the right place for this information.

Same feedback for Milvus config.

packages/server/src/memmachine_server/common/resource_manager/resource_manager.py
edwinyyyu, line 1: This may have merge conflict with #1661. Resolve correctly.

packages/server/src/memmachine_server/common/vector_store/collection_registry/collection_registry.py
edwinyyyu: Done in 109a5d626. A `PurgeClaim` is one purge round's hold on a tombstone. The registry fills in what the round needs to find the dead incarnation's points: `incarnation`, the value they carry, and `namespace` and `config`, which name the native collection they are in. The round reports what it found in `found`, which the registry records when the claim ends. The claim also carried the collection's `name`, which no round reads, and the clean-round stamp, which only the registry reads; both are gone from the dataclass, and the registry keeps the stamp from the row it claimed.

packages/server/src/memmachine_server/common/vector_store/collection_registry/collection_registry.py
edwinyyyu: No. Dropped in 109a5d626: the docstring is now `Startup.`, as on `SegmentStore`.

packages/server/src/memmachine_server/common/vector_store/collection_registry/collection_registry.py
edwinyyyu: Agreed, renamed in a8d474b39. The registry doesn't create or delete a collection: the store creates the native collection and the registry records it, and unregistering drops the record and queues the incarnation for purge. `create` and `delete` read as the store's operations.

packages/server/src/memmachine_server/common/vector_store/collection_registry/collection_registry.py
edwinyyyu: Rewritten in 109a5d626. A tombstone is due until a round finds no points under its incarnation, and due again once the retention has passed since that round. The docstring now says how a caller runs one round in the body (look under `claim.incarnation` in the native collection `claim.namespace` and `claim.config` name, delete what it finds, set `claim.found`) and what the registry records when the body ends.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
edwinyyyu: Shortened in 1cc7f9d5c: a read concurrent with the deletion may take effect before it, returning the content from before the deletion.

packages/server/src/memmachine_server/common/vector_store/vector_store.py
edwinyyyu: Two sentences in 1cc7f9d5c: when it returns, the collection is unreachable and its data is deleted or, on a store that reclaims it later, left for `purge_deleted_collections`; it is idempotent, as on main. That a handle goes stale is stated once, on `VectorStoreCollection`.

packages/server/src/memmachine_server/common/configuration/database_conf.py
edwinyyyu: Done in 6e0e903fb for both Qdrant and Milvus: the description now only says what the setting names. The key rule has its context in the Vector stores section of databases.mdx, which the configuration table row now links; why the registry is needed is in the PR description. Both Qdrant examples in configuration.mdx lacked this required key, so a copied configuration failed to load; they set it now.

packages/server/src/memmachine_server/common/resource_manager/resource_manager.py
edwinyyyu: Noted. I'll resolve it when one side merges, probably #1661 first. From a trial merge: four files conflict, `resource_manager.py` and `service_locator.py` and their tests. The correct resolution is one shared purge loop driving both sweepers, and #1661's `validate_partition_key` in the service locator; the stack from [vector store scale-out 3/5] up already carries that combination.

PR #1632, opened 2026-09-15
(Depends on #1611) Render session blocks and id markers (speedkick)

edwinyyyu:
Stacked on #1611 (context parts and block kinds), which is stacked on #1617 and #1597; this PR is the one commit on top of #1611's tip. `EventMemory.render_segments` gains session blocks and optional id markers, so an agent that searches and expands can name a session or a segment back to the memory from the text it was given.

## What changes

`EventMemory.render_segments(segments, *, datetime_format, parts=("author",), ids=())`:

- Takes any segments, in any order, and renders a segment given twice once (overlapping windows).
- Lays them out as a block per session, the sessions in the order of their latest timestamps with the latest last, a blank line between blocks; each block in the store's order `(timestamp, event_uuid, index, offset)`, one line per run of adjacent pieces of one event, as before. Nothing in the text says whether two lines are adjacent in the store: a filtered walk can omit a neighbor and no segment carries a position, so no rendering can promise contiguity and this one claims none.
- `ids` names what is marked, each independently: `"session"` heads every block with `[session:"<id>"]`, the id JSON-quoted since a session id is any string up to 255 bytes; `"segment"` starts every line with `[segment:<hex>]`, or `[segments:<first>..<last>]` when the line holds more than one segment. A uuid is 32 hex digits, no hyphens. The plural and the range say which id opens the event and which closes it without prompting; a one-segment event carries one id. An unknown kind raises `ValueError`.
- With no ids and one session the text is exactly what `render_segments` produced before, which `rerank` relies on (it renders one window per hit, ids off).

Rendered, with both kinds on:

```
[session:"planning"]
[segment:9531985d5d9d49f89818e811892f902b] [Thursday, January 15, 2026, 10:30 AM] Alice: "Can we move the launch to March?"
[segment:36f675cc81e74ef5a8e25d940ed90475] [Thursday, January 15, 2026, 10:31 AM] Bob: "March works if legal signs off by the 20th."

[session:"standup"]
[segment:6513270e269e4d37b2a74de452e6b438] [Thursday, January 15, 2026, 11:30 AM] Carol: "Legal signed off on the March launch."
[segments:6b0d549b6f03475a9600a35a099950d8..8d116ece173847d9bd9c172411e20b8f] [Thursday, January 15, 2026, 11:32 AM] Bob: "Roadmap draft is up; the launch row moves to March and pricing gets its own milestone."
```

The grammar a client can substitute short ids into is one marker per line: `\[segments?:([0-9a-f]{32})(?:\.\.([0-9a-f]{32}))?\]`, and `\[session:("(?:[^"\\]|\\.)*")\]` at a block's head.

## Decisions

1. **Flat input, not windows.** A window boundary would have been the only thing a windows-shaped input carried, and it cannot be honored: a window from a filtered walk may have events missing between two lines that look adjacent. So the renderer claims adjacency nowhere and needs no boundaries.
2. **Ids per event line, at the line's start.** Every event is addressable, not only a window's edges; the marker names what the line contains rather than where an id sits, so no suffix is needed (one line is one event) and a one-segment event is not marked twice.
3. **Short ids are the caller's.** The server cannot resolve an id abbreviated per conversation, so the client translates before calling `expand` regardless; emitting the abbreviation is then substitution on the markers above, and a returned or mutable mapping would add a type to the API and a request payload that grows with everything the model has seen. Minimal-unique prefixes (what claude-memory does) are the stateless alternative and are out of scope here.
4. **Hex, no hyphens**, per the review; ~4 tokens per id fewer than the canonical form, and the canonical form is still what every API surface speaks.

## Verification

- `uv run pytest packages/server/server_tests/memmachine_server/episodic_memory packages/server/server_tests/memmachine_server/server`: 642 passed. New in `TestRender`: block order by latest timestamp with the store's order inside a block, session headers with a quoted id, `segment`/`segments` markers, a segment given twice, an unknown kind rejected; the earlier render and rerank tests pass unchanged, which is the "same text as before" claim.
- `ruff check`, `ruff format --check`, `ty check`: clean (ty's two pre-existing `spacy` unresolved imports aside).


edwinyyyu:
Closed in favor of its main port, #1688, which carries this PR's content at its final tip; the review history stays here. See the "Source and review" section of #1688 for what was reviewed and what was adapted.

PR #1639, opened 2026-09-15
[vector store alt 4/9] Make no memory request create a project (speedkick)

edwinyyyu:

### Purpose of the change

Port of #1624: a memory request for an unknown project answers 404, the MCP add tool creates its own project once, and the default project `universal/universal` is created at startup. `docs/openapi.json` regenerated.

### Stack

Slice 4 of 9, every PR targeting `speedkick`; merge bottom-up.

| # | PR | change |
|---|---|---|
| 4 | #1639 **(this PR)** | Make no memory request create a project |
| 7 | #1642 | Remove custom sharding from the Qdrant store, and disable strict mode on its collections |
| 8 | #1643 | Mint an incarnation per collection life on Qdrant; delete logically, reclaim by purge |

This PR's own change is its commit `227c0cfef`; the rest of its diff is the slices under it, and drops out as they merge. Stacked on slice 3; slice 5 is stacked on it.

### Verification

`ruff check`, `ruff format --check`, `ty check`, `pytest packages/server/server_tests` (integration deselected): server 1934 passed, 3 skipped; run at this slice's own commit on 2026-09-15.


edwinyyyu:
Reopen if stack chosen.

edwinyyyu (commented):
Found while auditing the main version, #1624; both apply here.

1. Line comment below.
2. `memmachine-server --stdio` (app.py) calls `initialize_resource()` directly and never `init_global_memory`, so `ensure_default_project` does not run for that entrypoint; "at startup" covers the HTTP lifespan, mcp_http and mcp_stdio only. Same on #1624.

**packages/server/src/memmachine_server/main/memmachine.py:762** -- The 404 for an unknown project is raised only inside `if MemoryType.Episodic in target_memories`. A write with `types: ["semantic"]` (any request without episodic) to an unknown project is answered 200: `episode_storage.add_episodes` a few lines up has already inserted rows under the unknown session key (no FK to the sessions table), and the semantic manager ingests without a session lookup. Even with episodic included, the episode rows are persisted before this check runs, so the 404 arrives after the write. The description's "a write or a search opens the session and answers 404 for an unknown project" holds only for episodic requests. Same on #1624.

PR #1640, opened 2026-09-15
[vector store alt 5/9] Remove open-or-create and close from both stores (speedkick)

edwinyyyu:
Alternative to the `[vector store N/13]` chain (#1597, #1622..#1616): the two exclude each other. This chain keeps per-project `properties_schema` (user-defined metadata indexes) and conforms Qdrant to the incarnation lifecycle; SQLite and Milvus follow later. Every PR is a draft.

### Purpose of the change

Port of #1625, unchanged: `open_or_create_collection`, `close_collection` and `VectorStoreCollectionConfigMismatchError` go from the vector stores, `open_or_create_partition` from the segment store; the tests that need the pattern use a race-tolerant helper.

### Stack

Slice 5 of 9, every PR targeting `speedkick`; merge bottom-up.

| # | PR | change |
|---|---|---|
| 1 | #1636 | Restore per-project filterable properties |
| 2 | #1637 | Create a session's storage when the session is created, never on a request |
| 3 | #1638 | Make the scripts and examples that write to a project create it first |
| 4 | #1639 | Make no memory request create a project |
| 5 | #1640 **(this PR)** | Remove open-or-create and close from both stores |
| 6 | #1641 | Rename both stores' lookup to get_collection and get_partition |
| 7 | #1642 | Remove custom sharding from the Qdrant store, and disable strict mode on its collections |
| 8 | #1643 | Mint an incarnation per collection life on Qdrant; delete logically, reclaim by purge |
| 9 | #1644 | Bound every request to a remote vector store by a configured timeout |

This PR's own change is its commit `a5d3470eb`; the rest of its diff is the slices under it, and drops out as they merge. Stacked on slice 4; slice 6 is stacked on it.

### Verification

`ruff check`, `ruff format --check`, `ty check`, `pytest packages/server/server_tests` (integration deselected): server 1918 passed, 3 skipped; run at this slice's own commit on 2026-09-15.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1642, opened 2026-09-15
[vector store alt 7/9] Remove custom sharding from the Qdrant store, and disable strict mode on its collections (speedkick)

edwinyyyu:

### Purpose of the change

Two commits. `QdrantConf.is_distributed` put every logical collection in its own shard so that deleting one was a shard drop; with the next slice's lifecycle a deletion is a registry write and reclamation a filter on the incarnation, so the shard, with its own segments and graphs per logical collection, has nothing left to buy. The option goes from the configuration, the parameters, the database manager, the store and the tests that ran a one-node cluster for it.

The store then creates every collection it makes, physical and registry (and, in the next slice, the purge queue), with `strict_mode_config` disabled: a caller may filter on any property, declared or not, and Qdrant Cloud's default strict mode rejects a filter on an unindexed payload key. The chain that indexes every filterable key keeps strict mode on; this one turns it off. Collections created by an earlier version keep their setting. Both are breaking configuration changes on `speedkick`; `is_distributed` was never documented.

### Stack

Slice 7 of 9, every PR targeting `speedkick`; merge bottom-up.

| # | PR | change |
|---|---|---|
| 4 | #1639 | Make no memory request create a project |
| 7 | #1642 **(this PR)** | Remove custom sharding from the Qdrant store, and disable strict mode on its collections |
| 8 | #1643 | Mint an incarnation per collection life on Qdrant; delete logically, reclaim by purge |

This PR's own change is its commits `c8baf723a` and `9ae55c3c5`; the rest of its diff is the slices under it, and drops out as they merge. Stacked on slice 6; slice 8 is stacked on it.

### Verification

`ruff check`, `ruff format --check`, `ty check`, `pytest packages/server/server_tests` (integration deselected): server 1919 passed, 3 skipped; Qdrant integration lane (REST + gRPC) 122 passed; run at this slice's own commit on 2026-09-15.


edwinyyyu:
Reopen if stack chosen.

PR #1643, opened 2026-09-15
[vector store alt 8/9] Mint an incarnation per collection life on Qdrant; delete logically, reclaim by purge (speedkick)

edwinyyyu:

### Purpose of the change

A logical collection is identified to callers by its (namespace, name) pair and, inside the Qdrant store, by an incarnation minted when it is created. Points carry the incarnation, never the name, and a point's id is `uuid5(incarnation, record uuid)`, so a collection deleted and re-created under the same pair starts empty, its predecessor's points are never adopted by or reclaimed out from under the successor, and neither a stale handle nor a co-tenant of the shared physical collection can overwrite a point a live incarnation holds under the same record uuid (with bare record-uuid ids, as on `speedkick` and in #1631, they could). Handles are bound to one incarnation: once it is deleted, every operation of the handle raises `VectorStoreCollectionHandleStaleError`.

`delete_collection` becomes two registry writes: the live entry goes, conditionally on the incarnation the delete read, and an entry in the store's purge queue (`memmachine__purge_queue`, one per store, so one sweep serves every namespace) follows, stamped with the deletion time and the physical collection the incarnation lived in. `purge_deleted_collections` reclaims one dead incarnation per call, oldest first, by a filter on the incarnation; the resource manager runs it, one loop per store, as it runs the segment stores' purge. Qdrant has no transactions: a write that passed its fence before the deletion lands under the dead incarnation and is reclaimed with it; a crash between the two registry writes leaves the points unreachable and unreclaimed, a leak, never a collection the purge takes from under a live one.

The `VectorStore` contract states what every implementation guarantees (deletion makes the collection unreachable at once and is idempotent; a collection created under a deleted pair starts empty) and what it leaves to the implementation (immediate or deferred reclamation; what a handle held across a deletion does). Qdrant is the implementation that conforms here; the SQLite stores and the Milvus store keep their immediate reclamation, answer False from `purge_deleted_collections`, and state their handle behavior in their docstrings. Existing Qdrant data is not migrated.

`collection_lifecycle_contract.py` holds the tests a conforming store satisfies, run against Qdrant in local, REST and gRPC modes: stale handles, empty re-creation, idempotent deletion, purge reclaiming what deletion deferred and leaving every other collection alone (same configuration, another configuration, another namespace, and one re-created under another configuration while its predecessor awaits purge), and the same record uuid held by two collections of one physical collection.

### Tenant-created physical resources: limits and cross-tenant effects

Every project (tenant) declares its own `properties_schema`, fixed for the life of the project (to change it, delete the project and create it again; its memories start empty). Filters may name any `m.<key>`: a declared key is indexed where the backend supports it, an undeclared key is evaluated without an index. On Qdrant and Milvus the vector store maps every distinct `(namespace, vector dimensions, properties_schema)` to one **physical collection** that all projects declaring that schema share, created the first time a project declares it and never dropped by this code. One tenant's choices therefore change what other tenants get. Verified 2026-09-15 against the vendor docs and, for the Qdrant planner, Qdrant's `dev` source (`read_view/filtering.rs`, `read_view/dispatch.rs`, `sample_estimation.rs`).

- **Physical collections grow with distinct schemas.** A project declaring a schema nobody else uses creates a physical collection for the whole deployment. Qdrant Cloud allows 1000 collections per cluster by default and every collection carries its own resource overhead; Milvus allows 65,536 per instance. No reclaim of an empty physical collection is built.
- **Qdrant indexes every declared key** as a payload index on the shared physical collection, next to ~10 system keys MemMachine indexes itself. Each index costs memory for every tenant's points, and Qdrant (>= 1.16) degrades with large index counts. The store disables strict mode on every collection it creates, so Qdrant Cloud's default caps (`max_payload_index_count` 100, rejection of unindexed filters) do not apply; collections created by an earlier MemMachine keep their own setting.
- **Qdrant evaluates an undeclared key per point, within the project.** Every query carries the project's tenant condition on an `is_tenant` keyword index, which is the planner's primary clause, so the candidates are the project's own points, never the physical collection. An undeclared condition estimates as unknown (min 0), so the planner brute-forces the project's posting list with one payload read per point when the project's count in a segment is below `full_scan_threshold` (10,000 KB / vector bytes: ~1,700 points at 1536 dims), and otherwise samples 1,000 points to choose between that brute force and HNSW traversal with a payload read per visited node, where the missing `payload_m` links for the undeclared value fragment traversal under a restrictive condition (fewer results). Payload is in the `cold` tier (disk) by default. The cost lands on the querying project; other tenants feel it only as load on the shared node.
- **Milvus indexes nothing** (stated, not mitigated). The store creates no scalar index, so declared and undeclared keys cost the same: all keys are written as dynamic fields (`$meta`, 65,536 bytes per row, all keys together) and again into the `properties` JSON field, and the filter, tenant term included, is evaluated per row over the project's hash partition (1 of 16 by default), which holds every tenant hashing there. Milvus allows 64 fields and 1,024 partitions per collection, VARCHAR <= 65,535.
- **SQLite stores share nothing.** Each project gets its own tables (and, for the search-engine store, its own index file). `sqlite_vec` takes the `limit` nearest neighbors and applies the filter afterwards (`k = min(limit, 4096)`), so a selective filter returns fewer results than `limit`, declared or not; the search-engine store evaluates the filter per visited candidate with one SQL statement each and indexes no key.
- **Cross-tenant effects, plainly:** (1) the physical-collection count grows with every distinct schema ever declared and is bounded by the backend; (2) on Qdrant, index memory and query load are shared per physical collection, while undeclared filters cost the querying tenant, not its co-tenants' data; (3) on Milvus every filter scans the shared hash partition per row; (4) a schema cannot change in place.

### Stack

Slice 8 of 9, every PR targeting `speedkick`; merge bottom-up.

| # | PR | change |
|---|---|---|
| 4 | #1639 | Make no memory request create a project |
| 7 | #1642 | Remove custom sharding from the Qdrant store, and disable strict mode on its collections |
| 8 | #1643 **(this PR)** | Mint an incarnation per collection life on Qdrant; delete logically, reclaim by purge |

This PR's own change is its commits `ff8bcd62a` and `94cd5dfb5`; the rest of its diff is the slices under it, and drops out as they merge. Stacked on slice 7; slice 9 is stacked on it.

### Verification

`ruff check`, `ruff format --check`, `ty check`, `pytest packages/server/server_tests` (integration deselected): server 1931 passed at ff8bcd62a and 1932 passed at 94cd5dfb5, 3 skipped; Qdrant integration lane (REST + gRPC) 140 passed; run at this slice's own commit on 2026-09-15.


edwinyyyu:
Reopen if stack chosen.

Issue #1600, opened 2026-09-09
semantic_memory.enabled is honored only at startup: search, add and list still build and query the semantic stack whenever a request names the semantic type


edwinyyyu:
### What happened

`semantic_memory.enabled: false` only skips the semantic service at startup. Every request path that can name the semantic memory type still resolves the semantic stack lazily and uses it: `add_memories`, `search_memories` and `list_memories` on the REST API, and both MCP tools unconditionally. The flag does not make the memory type unavailable; it only changes who pays to bring it up.

Line numbers as of `21105d6e` (speedkick), `packages/server/src/memmachine_server/main/memmachine.py` unless stated:

- `start` (line 417) gates semantic service startup on `self._conf.semantic_memory.enabled` and logs "Semantic memory is disabled; skipping semantic service startup." (line 422). That is the only place the flag is consulted on the request-serving side.
- `add_memory` (line 777), `query_search` (line 1052) and `list_search` (line 1129) each branch on `MemoryType.Semantic in target_memories` and then call `self._resources.get_semantic_session_manager()`. None of them reads the flag. `query_search`'s docstring says "Search across enabled memory types", which is not what it does.
- `SemanticManager.get_semantic_session_manager` (`common/resource_manager/semantic_manager.py:255`) constructs `get_semantic_service` (line 218) on first use, which constructs `get_semantic_storage` (line 83) and `get_semantic_config_storage` (line 175) and resolves the semantic embedder and language model. Neither getter reads `enabled`. `get_semantic_storage` calls `storage.startup()`, which for the pgvector backend runs Alembic migrations (`semantic_memory/storage/sqlalchemy_pgvector_semantic.py:209-212`). So the first request that names the semantic type against a "disabled" semantic memory brings up the whole stack, migrations included, on the request path.
- The REST router (`server/api_v2/router.py`) defaults `types` to `[MemoryType.Episodic]` for add (line 297), search (line 315) and list (line 339) since #1555. That keeps an untyped REST request off the semantic stack, but an explicit `types: ["semantic"]` or `["episodic", "semantic"]` goes straight through.
- The MCP tools `mcp_add_memory` (`server/api_v2/mcp.py:468`) and `mcp_search_memory` (line 528) hardcode `target_memories=ALL_MEMORY_TYPES`, so every MCP add and search runs the semantic branch regardless of config.

What a caller sees depends on how complete the semantic config is, and both outcomes are wrong:

[...]

PR #1651, opened 2026-09-16
fix(config): let semantic memory be disabled with enabled: false alone

### LINKED ISSUE #1633: semantic_memory cannot be disabled with enabled: false alone: the section and config_database are required even when semantic memory is off
### What happened

Turning semantic memory off in the config file takes more than `enabled: false`. The configuration fails to load unless you also name a `config_database`, and it fails if the `semantic_memory` section is left out entirely.

Line numbers as of `98f3cde` (speedkick), `packages/server/src/memmachine_server/common/configuration/__init__.py`:

- `SemanticMemoryConf.config_database` is declared `Field(...)` (line 144), so it is required whatever `enabled` is.
- `Configuration.semantic_memory` has no default (line 373), so the section itself is required.

Checked against the model directly:

```
SemanticMemoryConf(enabled=False)
  -> 1 validation error for SemanticMemoryConf
     config_database  Field required [type=missing]

SemanticMemoryConf(enabled=False, config_database="pg")        -> OK
Configuration(... semantic_memory section omitted ...)
  -> 1 validation error for Configuration
     semantic_memory  Field required [type=missing]
```

This contradicts #1580 ("default to no short-term or semantic memory"). Semantic memory is off by default, but the operator still has to write a semantic section and point it at a database that will never be used. Every other semantic field is optional, and `_auto_disable_when_incomplete` (line 180) already treats a missing storage field as grounds to disable rather than to fail.

### Expected

With semantic memory disabled, `semantic_memory:` should be optional and so should `config_database`. Either of these should load:

```yaml
semantic_memory:
  enabled: false
```

or no `semantic_memory` section at all.

Concretely: make `config_database` optional (`str | None = None`), give `Configuration.semantic_memory` a default, and fold a missing `config_database` into `_auto_disable_when_incomplete` for configs that do set `enabled: true`. `SemanticManager.get_semantic_config_storage` (`common/resource_manager/semantic_manager.py:175`) would then need to raise `ResourceNotReadyError` when it is unset, as the storage getter already does for `database`.

### Related

- #1600 / #1601: the runtime side of the same flag (request paths ignore `enabled`). This issue is only about the config schema.
- #1575: project deletion with semantic memory disabled. Its reproduction config shows the workaround this issue asks to remove (`enabled: false` plus a `config_database`).

marvinyu-memverge:
### Purpose of the change

Let semantic memory be turned off with `enabled: false` alone, or by omitting the `semantic_memory` section. Today the configuration refuses to load unless the section is present and `config_database` names a database that will never be used.

### Description

`SemanticMemoryConf.config_database` was declared required (`Field(...)`) regardless of `enabled`, and `Configuration.semantic_memory` had no default, so an operator disabling semantic memory still had to write a semantic section and point it at a config database.

This change:

- makes `config_database` optional and folds a missing value into the existing auto-disable validator, so a config that sets `enabled: true` without it is auto-disabled with a warning naming the missing fields, the same way a missing storage, LLM, or embedder already is;
- gives `Configuration.semantic_memory` a default of a disabled section, and treats a null or empty `semantic_memory:` key the same way, so the section can be omitted;
- makes `SemanticResourceManager.get_semantic_config_storage` raise `ResourceNotReadyError` when no config database is set, matching what `get_semantic_storage` already does for `database`;
- re-runs both load-time semantic guards (required fields, OpenAI credentials) after a runtime config API update. The API assigns fields in place without re-validating, so with `config_database` now optional an enable request could otherwise persist `enabled: true` on a config the server cannot build. An update that leaves the config incomplete is auto-disabled and the change summary says why.
- updates the configuration docs so the four fields that are only needed when semantic memory is enabled are described the same way.

Configs that set `config_database` today are unaffected. The runtime request paths that ignore `semantic_memory.enabled` (#1600, #1601) and project deletion with semantic memory disabled (#1575) are separate issues and are not touched here.

### Fixes/Closes

Fixes #1633

### Type of change

### How Has This Been Tested?

- [x] Unit Test

New tests cover: `SemanticMemoryConf(enabled=False)` with no config database; auto-disable when `config_database` is missing; the sample config with the section removed, null, empty, and `{enabled: false}`; both disabled shapes round-tripping through `to_yaml()`; the config-storage getter refusing an unset database without touching the SQL engine; and the config API auto-disabling an enable request that lacks `config_database` or whose embedder has empty OpenAI credentials.

Each guard was mutation-checked: reverting the validator change, the section default, the null/empty mapping, the getter guard, or either API re-check fails exactly the test that names it.

**Test Results:** `uv run pytest packages/server/server_tests` — 1806 passed (the two `installation/test_install_in_windows*` failures reproduce on `main` on this machine and are unrelated); `packages/client/client_tests` 255 passed; `ruff check` / `ruff format --check` clean; `ty check packages` unchanged from `main`.

### Checklist

### Screenshots/Gifs

N/A

### Further comments

The same declarations exist on `speedkick`, where #1580 made semantic memory disabled by default, so this fix is worth carrying there too; the diff is confined to the config model, the two getters, and the config service.


packages/server/src/memmachine_server/common/configuration/__init__.py
```diff
@@ -179,28 +180,45 @@ class SemanticMemoryConf(YamlSerializableMixin):
...
+        elif not self.database:
+            missing.append("database")
+        missing.extend(
+            name
+            for name in ("config_database", "llm_model", "embedding_model")
+            if not getattr(self, name)
+        )
+        return missing
+
+    def auto_disable_when_incomplete(self) -> list[str]:
```
malatewang, line 203: The caller does not use the return data. It should be enough to log a message.

malatewang:
When semantic memory is disabled, the memmachine needs to handle the case when processing read/write request

packages/server/src/memmachine_server/common/configuration/__init__.py
marvinyu-memverge: The validator does discard it, but `Configuration.auto_disable_semantic_memory` uses the returned names — they become the `auto-disabled: missing required fields: config_database` text in the config API's response, so someone disabling semantic memory through the API is told which field is missing. That came out of an earlier review round asking for the reason to name a field.

Happy to drop it if you'd rather the API just reported that it was disabled.

marvinyu-memverge:
Good catch — fixed in 3692ae7 and fcdb516. Two things were reaching semantic memory without checking whether it was enabled:

- `add_episodes` / `query_search` / `list_search` defaulted `target_memories` to every memory type and dispatched on that alone, so with the section disabled the first request hit a backend with nothing configured and returned an unhandled 500.
- `delete_episodes` resolved the semantic service before building any of its work, so `POST /memories/episodic/delete` — which names no semantic memory at all — returned a 500 *and* left the episodes undeleted.

Both now skip semantic memory when it's off, the same way startup, shutdown and session deletion already did.

What I didn't touch: a request that *explicitly* names semantic memory on a disabled server. `{"types": ["semantic"]}` now comes back `200` with an empty result instead of erroring, and the `/memories/semantic/*` routes still go straight to the backend. I think those want a 4xx rather than a silent skip, but that's an API contract call rather than a bug fix, so I left it out of this PR — it belongs with #1575, which is the same area. That one also has `_cleanup_semantic_history` defined twice in `MemMachine`, so the guarded copy is dead code and the unguarded one runs.

edwinyyyu (approved):

malatewang (approved):

PR #1653, opened 2026-09-16
Run OpenAI-backed integration tests against an in-process fake API

edwinyyyu:
### Purpose of the change

`Run Python integration tests` has failed on every push to `main` since #1477 (2026-08-13): the `common/`, `main/` and `semantic_memory/` jobs all fail with `openai.AuthenticationError: 401 invalid_api_key` because the repo's `OPENAI_API_KEY` secret stopped being valid. That secret has now been removed, which would turn those failures into silent skips (`openai_integration_config` calls `pytest.skip`). This PR makes the OpenAI-backed integration tests run without any real API instead.

### Description

- `packages/server/server_tests/memmachine_server/fake_openai_api.py`: an in-process fake of the OpenAI HTTP API (FastAPI app served by uvicorn on an ephemeral loopback port), started once per test session. It serves:
  - `POST /v1/embeddings`: validates the request the way the real API does (input shape, empty strings rejected, at most 2048 inputs per request), honors `dimensions` and `encoding_format` (`float` and the SDK's default `base64`), and returns deterministic feature-hashed bag-of-words vectors (equal texts -> equal vectors, shared words -> positive cosine). The embedder's chunking/clustering is therefore still exercised: `test_many_inputs` still has to split 10000 inputs across requests.
  - `POST /v1/responses`: for structured requests, generates an instance of the request's JSON schema (every property filled, arrays get one element, `enum`/`anyOf` take their first alternative, strings are the property name); for plain requests returns a fixed sentence. Usage counts are populated so the metrics paths run.
- `conftest.py`: `openai_integration_config` serves the fake unless `OPENAI_API_KEY` is set, in which case the tests call the real API as before. `openai_client` and the `main/` embedder and language-model configs pass the fake's `base_url` through.
- `.github/workflows/pytest-integration.yml`: the credentials `env` block is removed. The repo has no `OPENAI_API_KEY`, `AWS_ACCESS_KEY_ID` or `AWS_SECRET_ACCESS_KEY` secret, so the block was already inert; the Bedrock/Cohere/Ollama-gated variants keep skipping exactly as they do today.
- `AGENTS.md`: one bullet on how the integration tests reach OpenAI.

What this changes about what the tests verify: the `main/` and `semantic_memory/` tests assert on MemMachine plumbing (CRUD, pagination, ingestion completing, search returning results), which a schema-conforming fake exercises fully. The two `slow` LongMemEval tests (`test_periodic_mem_eval`, `test_long_mem_eval_via_memmachine`) end with `"The relevant information is not found..." not in eval_resp`, where `eval_resp` is the `(text, tool_calls)` tuple, so that assertion can only fail if the whole answer equals that string exactly; it was not testing answer quality before this PR either and is left as is. The remaining value of those tests (full-snippet ingestion completes; semantic + episodic search return results) is kept and now runs in under a minute.

### Type of change

### How Has This Been Tested?

- [x] Integration Test

Local runs with `OPENAI_API_KEY` unset (Docker available for the pg/neo4j containers), `uv run --frozen --all-extras pytest <dir> -m integration` as the `main` job runs it:

| directory | before (this branch's parent) | after |
|---|---|---|
| `common/` | 284 passed, 109 skipped (14 for `OPENAI_API_KEY`) | 298 passed, 95 skipped |
| `main/` | CI: 1 passed, 20 errors | 21 passed, 46 skipped (Ollama/Bedrock variants) |
| `semantic_memory/test_semantic_llm_integration.py` + `test_semantic_memory_integration.py` | CI: job ran 5 h then failed | 26 passed, 52 skipped, 60 s |

Also `ruff check`, `ruff format --check` and `ty check --project packages/server` pass.

### Checklist


edwinyyyu:
Fake API requires maintenance.

PR #1656, opened 2026-09-16
[vector store 11/14] Keep the Qdrant and Milvus partition registries in SQL, so any process may create, delete and purge (speedkick)

edwinyyyu:
### Purpose of the change

Qdrant and Milvus have no transactions, unique constraints or conditional writes, so the registry each store kept inside its backend could not arbitrate two server processes creating, deleting or reclaiming the same partition: a create was read-then-write under a per-process lock, two creators both succeeded with the last registration winning (Milvus, whose insert does not enforce primary keys, kept both), and a delete was two unrelated writes. The stores now keep their registry in the deployment's relational database, `SqlPartitionRegistry`: one table pair per backend kind (`vector_store_qdrant_pt`/`_gc`, `vector_store_milvus_pt`/`_gc`), shared by every store on that kind of backend, a row per live partition keyed by collection and partition key with the incarnation its points carry, and a purge queue of dead incarnations. Creation is an insert the primary key arbitrates (a racing creator on any process gets AlreadyExists); deletion is one transaction, a DELETE that takes the row's write lock and the queue insert; a purge claim is a row lock held while the backend deletes the points, `FOR UPDATE SKIP LOCKED` on PostgreSQL so concurrent purgers split a backlog, retired when the delete returns and kept when it raises; the per-operation fence is a registry lookup by incarnation. The registry collections, the deterministic registry point ids, the deletion stamps and the per-process lock tables go, and with them `registry_replication_factor`.

`QdrantConf` and `MilvusConf` gain a required `registry_database`, the name of a relational database under `resources.databases`; DatabaseManager hands its engine to the store, and `provision()` creates the two tables on it, surviving a racing provisioner. The SQLite stores already keep their registry beside their data in one file and are unchanged.

The `VectorStore` contract no longer restricts a partition to one process: every operation is safe from any process sharing the backend, and a store that cannot give that says so itself (the engine-backed SQLite store holds a partition's index in the process that opened it; the sqlite-vec store is shared by the processes of one node).

### Stack

Slice 11 of 14, every PR targeting `speedkick`; merge bottom-up.

| # | PR | change |
|---|---|---|
| 2 | #1622 | Create a session's storage with the session, never on a request |
| 3 | #1623 | Make the examples that write to a project create it first |
| 4 | #1624 | Make no memory request create a project |
| 5 | #1625 | Remove open-or-create and close from both stores |
| 6 | #1626 | Rename logical collection to partition, and open to get, on both stores |
| 7 | #1654 | Remove custom sharding from the Qdrant store |
| 8 | #1630 | Bound every request to a remote vector store by a configured timeout |
| 9 | #1627 | Make a vector store one collection, with string-keyed partitions |
| 10 | #1631 | Mint an incarnation per partition life; delete logically, reclaim by purge |
| 11 | #1656 **(this PR)** | Keep the Qdrant and Milvus partition registries in SQL, so any process may create, delete and purge |
| 12 | #1618 | Let a deployment tune a Qdrant collection's HNSW, optimizers and quantization |
| | #1597 | EventMemory session, source and expansion: not a slice of this chain; 13 and 14 need it |
| 13 | #1628 | Make a vector store filter only on the properties it declares |
| 14 | #1616 | Close the filter union, and make negation the complement on every backend |

This PR's own change is its last commit, `be0ddabb` (19 files changed, 781 insertions(+), 598 deletions(-)); the rest of its diff is the slices under it, and drops out as they merge. Stacked on #1631; #1618 is stacked on it.

### Verification

`ruff check`, `ruff format --check`, `ty check` (two pre-existing spacy diagnostics), `pytest packages/server/server_tests packages/client/client_tests`: 2212 passed, 3 skipped, on the slice-12 tree (`speedkick` `973aad52` plus slices 2-12), 2026-09-16; every slice passed its own affected suites when built.


edwinyyyu:
Closed: folded into #1631 on Edwin's call. Incarnations and the SQL-arbitrated registry are one design; the intermediate state with registries inside Qdrant and Milvus had no value of its own, so #1631 now carries both (its stores commit) and this PR's content lives there.

PR #1657, opened 2026-09-16
[vector store 11/14] Check liveness after every remote operation; keep a deleted partition's tombstone until a clean sweep and a retention

edwinyyyu:
### Purpose of the change

A Qdrant or Milvus handle read its incarnation as live and then made the remote call; a deletion committing between the two left the write under a dead incarnation and the operation reporting success. The handle now reads the row again after the call and raises the stale error if the incarnation died meanwhile, so a completed operation never reports success on a dead partition. No lock spans the remote call and the logical delete waits for nothing; a write that landed under a dead incarnation is the purge's to reclaim.

One purge could not be the last: a Qdrant or Milvus write the registry read as live can land after the purge that followed the deletion, and the queue entry was retired at that first purge, so nothing would ever find those points. The queue entry is now the incarnation's tombstone. A purge round looks for one point under the incarnation and deletes by filter if there is one; a round that found points keeps the entry due; a round that found none stamps it clean; and the entry is removed only by a round that finds nothing again once `tombstone_retention_seconds` (a day unless configured, per store, on the database clock) has passed since the clean stamp. A round that finds a late write clears the stamp and the rounds start over. Until removal the incarnation is never re-minted. `purge_deleted_partitions` returns True when the round found points, so the sweeper comes back for another round, and False when it found none or nothing was due.

The retention is the one clock in the design and it decides nothing about validity: a stale write is refused by the liveness check after the operation, and the retention only has to exceed, by orders of magnitude, the longest a request can be in flight, which request_timeout_seconds bounds.

A retention is set in days and compared on the database clock; a fraction of a second has no use, and the field then matches request_timeout_seconds and max_retry_interval_seconds.

### Stack

| # | PR | change |
|---|---|---|
| [speedkick port 12/17] | #1666 | Serialize a collection's writes so the engine sees them in order (port of #1607) |
| [speedkick port 14/17] | #1668 | Take SQLite's write lock at BEGIN, not at the first write (port of #1609) |
| [vector store 11/14] | #1657 **(this PR)** | Check liveness after every remote operation; keep a deleted partition's tombstone until a clean sweep and a retention |

This PR's own change is its last 3 commits, `5d6212199`, `f5e256685`, `b40872e1f`; the rest of its diff is the PRs under it. Stacked on #1631; #1618 is stacked on it.

### Verification


edwinyyyu:
Folded into #1631 as its last three commits (post-operation liveness check, tombstone retention, whole-second retention); the chain on `main` is indexed in every PR body.

Issue #1658, opened 2026-09-16
Event memory: bound the vector upsert one encode transaction spans


edwinyyyu:
## Context

The follow-up to #1597 moves the vector upsert inside the segment store's write transaction, so segments commit only after their records are acknowledged by the vector store (a failed upsert rolls the segments back; a forget can only ever see links whose records exist). A bulk `encode_events` batch then holds one SQL transaction, and on SQLite the database writer lock, for the whole batch's upsert.

Costs that grow with batch size while the upsert is in flight:

- PostgreSQL: one pooled connection stays busy for the batch's upsert, and partition deletion (which takes the registry row for update) waits that long.
- SQLite: every writer on the database file waits for the batch's upsert, so a large import stalls other partitions' writes for the upsert's duration.

## Ask

Bound the work one encode transaction spans, so that a single transaction never covers a multi-second vector upsert. Either:

- chunk bulk batches inside `encode_events`, each chunk its own transaction and upsert (the caller still sees one call), or
- reject oversized batches at the API with a documented request limit.

Not decided yet: chunk size versus request limit, and whether it is a knob or a constant. This issue exists so the bound is not forgotten; the follow-up PR ships without one.

PR #1659, opened 2026-09-16
(Depends on #1597) Encode an event once, with its records inside the store's transaction (speedkick)

edwinyyyu:
Stacked on #1597: its commits come first, and this PR is the commits after 80f0c713f. Follows the orphaned-record analysis on #1597's review: with the vector upsert outside the segment store's transaction, a forget interleaving between an encode's commit and its upsert orphans the encode's records, and nothing ever names them again; #1597's forget-first encode had widened that race to concurrent re-encodes before 80f0c713f removed it. This change prevents the race instead of repairing after it, and repairs the residue a crash can leave.

## What changes

**Segment store contract** (`segment_store.py`)
- `write(*, exclusive=False)`: a write transaction as an async context manager. Entry checks the handle and pins the partition against deletion; normal exit commits; an exception rolls the block back, so a caller can make a write conditional on work of its own inside the block. `exclusive=True` takes the registry row `FOR UPDATE`: it waits for every write in flight and excludes new ones until the block exits, for a reader that must see the partition settled.
- `SegmentStorePartitionWriter`, the handle inside the block: `add_events(events)`, keyed by event uuid, each with its segments and their derivative uuids; and `get_segment_uuids_by_derivative_uuids`, the same read inside the transaction.
- **The partition holds an event at most once.** A new event table, `segment_store_ev` with primary key `(incarnation, uuid)`, is inserted first with `ON CONFLICT DO NOTHING RETURNING`; any uuid missing from the result rejects the whole batch with `SegmentStoreEventAlreadyStoredError`, naming them, before anything is stored. Exact under concurrency on both dialects: PostgreSQL makes the second inserter wait on the first transaction and then report the conflict; SQLite serializes on the writer lock. Segments carry a cascading foreign key to the event row.
- `delete_events(event_uuids)` replaces the by-event path of forget: event rows locked in order, then their segment rows, then deleted, cascading to segments and links. `delete_segments` stays for eviction and leaves the event held, so an evicted event cannot come back through a replay; only `delete_events` frees the uuid.
- `get_derivative_uuids_by_event_uuids` replaces `get_segment_uuids_by_event_uuids` and `get_derivative_uuids_by_segment_uuids`, whose only caller was forget; a held event with no derivatives answers an empty list.
- `add_segments` is gone; the purge reclaims event rows after the segments, on the same budget.

**EventMemory** (`event_memory.py`)
- `encode_events`: segment, derive, embed, then one `write()` block: `add_events`, then the vector upsert inside the transaction, so segments commit only once the vector store has acknowledged their records, and a failed upsert rolls them back. On an upsert error the same ids are deleted before the error propagates, since the upsert may have been applied before it failed; a delete that fails too is added as a note on the upsert's error. Encoding an event twice is rejected whole; forget frees it.
- `forget_events`: derivative uuids by event, vector delete, `delete_events`. Records before events: a failed record delete leaves the event whole for a retry.
- Read repair in `query`: a hit whose link is missing is re-checked under `write(exclusive=True)`, which waits for every encode in flight; a link found then is kept, a record still without one is an orphan, deleted after the fence is released. Exact because derivative uuids are never reused. Counter `event_memory_orphan_records_deleted_total`.

**LongTermMemory**: tenant deletion removes the segment partition before the vector collection, so the deletion drains and then blocks writers before the collection deletion removes every record that could have landed.

## What it guarantees, and what it does not

- No orphan from any interleaving of encode, forget and eviction: a forget can only see links whose records exist, and its delete is issued after our upsert's acknowledgment. No duplicate copies: an event is held once.
- The residue is a process dying between the upsert's acknowledgment and the commit, or a request delivered after its compensating delete: crash-rate, an embedding plus reserved ids with no text, reclaimed when retrieved and by tenant deletion. A ledger with row states, a sweeper and retention was designed and rejected as not worth its cost for that residue.
- A weak-ordered Qdrant replica set can still resurrect a deleted point; that is Qdrant's documented default for any client, and read repair converges it on sight.

## Cost

- PostgreSQL: a pooled connection stays busy for the vector call; writers do not conflict with each other, only partition deletion waits longer. Per-request latency is unchanged: the same statements and the same vector call, reordered.
- SQLite: the writer lock is held across the vector call, so every writer on the file serializes on segment time plus vector time. A sqlite-vec store on the same file as the segment store cannot be used with this design.
- Bulk batches hold one transaction for the whole batch's upsert; bounding that is #1658.
- The check that a PostgreSQL transaction timeout is not shorter than the vector request timeout waits for the vector-store chain's timeout fields.

## Decisions

- Same no-migration stance as #1597: the event table and the foreign key are new DDL, and speedkick tables are recreated.
- The store is renamed `EventMemoryStore`, config key included, in the last commit, since it now holds events, segments and links.

## Tests

`test_sqlalchemy_segment_store.py`, both dialects: rejection whole with the uuids named, a stored segment uuid under another event still rejected, a segment under the wrong event, rollback when the block raises, `delete_events` cascade and idempotence, `delete_segments` keeps the event held, stale handle at `write()` entry, purge budget with event rows; PostgreSQL: the exclusive write blocks on a write in flight and then sees it, a concurrent add of one event blocks on the uncommitted row and is rejected after it commits; SQLite: the exclusive write waits on the writer lock. `test_event_memory.py`: rejection then forget, an applied-then-failed upsert leaves nothing and the retry succeeds, a failed compensating delete is noted, an orphan is deleted on retrieval and not returned, a record whose encode is in flight is kept.


edwinyyyu:
Closed in favor of its main port, #1686, which carries this PR's content at its final tip; the review history stays here. See the "Source and review" section of #1686 for what was reviewed and what was adapted.

Issue #1568, opened 2026-09-01
SQLite per-connection state (foreign keys, sqlite-vec extension) is registered per-store on caller-supplied engines


edwinyyyu:
SQLite requires per-connection state for two things this codebase depends on: `PRAGMA foreign_keys=ON` (FK enforcement, including `ON DELETE CASCADE`) and loadable extensions (sqlite-vec). Several stores register that state via a `connect` listener in their own `__init__`, on an engine supplied through their params. That placement has two structural faults, independent of the store:

1. **Connections pooled before the store exists never receive the state.** A `connect` listener fires only for new DBAPI connections; anything the (possibly shared) engine pooled earlier -- e.g. a validation `SELECT 1` -- serves later statements with foreign keys off or the extension unloaded. For FK, that means cascade deletes silently leave orphaned child rows; nothing errors.
2. **Registering listeners on a live pool races its event dispatch.** SQLAlchemy dispatches events by iterating a plain deque; a concurrent `event.listen` (a second store constructed while the engine serves traffic) mutates it mid-iteration and raises `RuntimeError: deque mutated during iteration`. Reproduced deterministically in #1545's concurrency tests.

#1545 hit exactly this in the segment store (a live bug there: the shared relational engine is validated -- and therefore pooled -- before the store is constructed) and established the fix convention:

- per-connection state is registered at **engine creation**, before anything is pooled: `enable_sqlite_foreign_keys(engine)` in `common/resource_manager/database_manager.py`, applied to the engines the `DatabaseManager` creates;
- the store **verifies instead of mutates**: `startup()` checks `PRAGMA foreign_keys` and refuses an unenforced engine with a directive naming the helper.

Remaining instances of the old pattern, out of #1545's scope:

- `common/vector_store/sqlite_vector_store.py` (~line 691): `SQLiteVectorStore.__init__` registers the FK pragma on its caller-supplied `params.sqlalchemy_engine` (the second decorator, on the sync engine it creates itself two lines above, is fine). Latent rather than live today: the `DatabaseManager` creates that engine immediately before constructing the store, so nothing is pooled ahead of registration on the production path -- but the params API accepts any engine, and the dispatch race applies whenever a store is constructed while the engine serves traffic.
[...]

PR #1661, opened 2026-09-16
Overhaul segment store: shared tables with incarnation-scoped tenant keys (port of #1548)

### LINKED ISSUE #1557: [Bug]: Segment store corrupts non-UTC timezone-aware timestamps on SQLite (write path + filter bounds)

edwinyyyu:
### Port

Copy of #1548, merged into `speedkick` as `7752e4cb7`, on `main`: the same commit cherry-picked. Its added and removed lines equal 7752e4cb7's in every file but `uv.lock`, whose hunk is dropped (it also dropped stale `datasets`, `dill`, `multiprocess` and `fsspec[http]` entries that main's lock already lacks). Context differs in one file, `test_event_backend_wiring.py`; `database_manager.py`'s surroundings match again since #1679 (the port of #1552) reached main. One line differs in `test_resource_manager.py`: the origin adds `from unittest.mock import create_autospec` where main's line already imports `AsyncMock` and `patch` (#1683), so the port extends that line. A second commit drops a duplicated `@override` on `purge_deleted_partitions` that the speedkick merge carried. #1548's Testing section is speedkick's run; the Verification below is this branch's.

### Purpose of the change

Addresses #1544, #1546, and #1549 by construction: the segment store's per-tenant partitioned layout is replaced with shared tables and an incarnation-carrying tenant registry, on every dialect. Design record: `design/segment_store_shared_tables.md`.

Also fixes #1557, #1558, and #1559: #1462's UTC-normalization fixes are folded in for every SQL store (segment, episode, cluster), so no store ships with the SQLite timestamp corruption.

### Why an overhaul instead of the earlier targeted fixes

This PR began as three targeted repairs to the partitioned layout (detach-before-drop for #1544, a management lock for the create/delete deadlock, child-table DML for #1546), each verified on its own. Further investigation showed the layout itself was the root cause and could not be fully repaired:

- PostgreSQL's FK integrity triggers still addressed the partitioned parents internally, so writes and deletes kept a per-backend lock spike proportional to partition count even after every client statement was fixed (10 -> 166 -> 10 relation locks across the trigger's generic-plan attempt).
- Writer-vs-lifecycle-DDL deadlocks through the shared parents survived every lock discipline: 41-83 deadlocks per 20-second churn on upstream and on every intermediate fix, in two cycle shapes captured from the server log, one of them a lock-upgrade cycle inherent to dropping a child that carries a cloned FK.
- Scaling requirements ruled out every per-tenant-table variant: tenant creation must stay cheap at 1e5-1e6 tenants (a per-tenant table costs ~5 catalog relations plus disk files), with 1e4-1e7 rows per tenant.

### Description

- **Shared tables everywhere.** The ORM models are the physical schema on every dialect. PostgreSQL partitioning, per-tenant DDL, detach machinery, the store-wide management lock, and the per-partition entity cache are removed. With no partitions there is no generic-plan lock fan-out, and with no lifecycle DDL there is nothing for the deadlock class to live in: the churn smoke that measured 41-83 deadlocks/20s on every partitioned build measures zero, with ~60x more write throughput and ~200x more lifecycle throughput.
- **Incarnation-scoped tenant identity.** `segment_store_pt` is the tenant registry: logical key (primary key) plus an `incarnation`, a random unique-constrained UUID. Data rows carry the incarnation alone, so a data query cannot be built without resolving the registry, and referencing the wrong tenant is structurally impossible. A deleted-and-recreated tenant never sees its predecessor's rows, even mid-purge. A colliding mint is rejected rather than left to probability: the unique constraint rejects a collision with a live incarnation, and the mint transaction re-checks the purge queue after its insert so an incarnation whose garbage is still awaiting purge is re-minted (race-free with the existing tables: the check runs after the registry insert, and no new queue entry for the minted value can appear before commit, since the only registry row carrying it is uncommitted). Any integrity rejection with no committed row under the key is likewise retried with a fresh incarnation, up to `_MAX_MINT_ATTEMPTS`; a persistent cause surfaces as `SegmentStoreAttemptsExhaustedError` with the database error chained. Rejections are warning-logged at the detection site, since a genuine collision is astronomically unlikely and the log is the signal for broken randomness or a persistent database error even when the re-mint heals it.
- **O(1) deletion plus a purge queue.** `create_partition` is a row insert (microseconds, no DDL). `delete_partition` takes `FOR UPDATE` on the registry row (waiting out writers' `FOR SHARE` pins), enqueues the incarnation on `segment_store_gc` (logical key kept for forensics), and deletes the row: O(1) at any tenant size, the pool-model contract of "unreachable immediately, erased asynchronously".
- **Purge as an ABC capability.** `purge_deleted_partitions() -> bool` is part of the `SegmentStore` interface. The caller's whole protocol is "call until False"; sizing is not a call argument, because callers cannot know engine-appropriate bounds. Deployments set them once at construction: `purge_max_segments` (default 10000) and `purge_max_partitions` (default 100, its own bound because retiring an empty entry costs ~0.95 ms, ~200x a segment row, and empty partitions are cheap to mass-create-and-delete). Each call is one transaction (measured ~147k-312k rows/s), committing its progress or nothing. Within an entry, a call continues from the entry's cursor, `purged_through` (the last segment key purged): it reads the next keys after it in primary-key order, deletes that key range, and records its last key in the same transaction, so no batch rereads rows earlier batches deleted (PostgreSQL keeps them in the table and its indexes until vacuum), and the call that finds the entry empty costs one index lookup. The cursor relies on no row being written into an incarnation once it is enqueued, which the write fence guarantees. Links follow by `ON DELETE CASCADE`, 50-68% faster than manual link deletion at one and four links per segment; cascade deletion saturates ~3M link rows/s (380k segments/s at 1 link/segment, 46k at 64), so heavily linked partitions purge in sub-second calls. Link fan-out is set by the deriver, which the store cannot reject after derivation. Link rows go by the cascade alone: the store requires foreign-key enforcement of its engine and does not verify it. Entries are claimed one at a time, oldest first (enqueue stamped by the database clock, indexed), with `FOR UPDATE SKIP LOCKED`, so a call neither materializes nor locks the rest of the backlog and concurrent purgers from any process share the queue; only the claiming call touches a dead incarnation's rows, so reclamation is deadlock-free by construction. On SQLite, which drops locking clauses and defers BEGIN to the first write, the claim is an UPDATE of the queue row, so purgers serialize at the claim and each reads the cursor its predecessor committed. The store never schedules purging; implementations whose deletes reclaim physically return `False`. The delete path does not purge: an inline drain was tried, first of the global queue and then scoped to the deleted key, and removed, since prompt physical erasure is not a promise the store can keep on every dialect (on SQLite any writer past the busy timeout fails) and the sweeper reclaims within its interval; a deployment must run the sweeper, and the ABC says so.
- **Fencing (#1549).** Writes pin the registry row under `FOR SHARE` with an incarnation predicate. Reads add the same predicate to their data statement as an `EXISTS`, so one statement checks liveness and reads; a read that finds no rows issues the registry check on its own, to tell an empty partition from a stale handle, and a windowed read ends with one registry read because its statements take separate snapshots. On SQLite the driver defers BEGIN until the first write, so a SELECT-only fence checks nothing; BEGIN IMMEDIATE would mean taking over transaction management of the caller's shared engine, so the fence is a self-checking registry-row UPDATE (same write lock, scoped to the transaction, match count as the staleness check), and deletion opens its transaction the same way. A handle that outlives its partition or its incarnation raises `SegmentStorePartitionHandleStaleError` on every dialect; SQLite previously had no stale-handle detection at all. Measured cost: none on reads that find rows, one registry round trip on reads that find none.
- **Foreign keys.** The segment table's FK to the registry is removed on purpose (registry and data rows decouple so deletion is O(1)); the link table's FK to the segment table and its cascade remain, enforced by a single-row indexed check with no partition fan-out. On SQLite, enforcement is per-connection state registered at ENGINE creation (`enable_sqlite_foreign_keys`, applied by the DatabaseManager to its engines): a listener added later misses connections the shared engine already pooled, a pre-existing bug surfaced in review round 4. The store states the requirement on its `engine` parameter and does not verify it, which is the reference practice for SQLite foreign keys; an engine without the pragma leaves link rows that outlive their segments until the partition is dropped, when the purge reclaims them with a warning. The SQLite vector stores carry the same latent pattern; filed as #1568.
- **Folded in: #1462's SQL-store fixes (fixes #1557, #1558, #1559).** App-supplied timezone-aware datetimes are normalized to UTC before persisting in every SQL store: segment timestamps, episode `created_at` and its `start_time`/`end_time` bounds, cluster `last_ts` and pending `created_at`. Without this, SQLite (whose `DateTime(timezone=True)` discards tzinfo) read non-UTC values back shifted by their offset. Datetime filter values are normalized to UTC-aware instants at `Comparison`/`In` construction, the filter language's contract stated on `FilterExpr` (a value denotes an instant; naive means UTC), so every consumer receives instants and compilers only choose a representation; the SQL column leaf binds values as-is, and remaining per-backend normalizations are idempotent defenses, removable separately. Two consumers change by design: the Neo4j compiler's `datetime.timestamp()` read a naive value in the server's local zone and now receives instants, with its ISO-string branch parsing under the same rule (pinned under a non-UTC process zone); the in-memory short-term evaluator tags a naive stored metadata datetime as UTC at comparison time, since stored user data is not rewritten. The episode store's bounds spell the convention inline as two explicit steps, `ensure_tz_aware(...).astimezone(UTC)`, kept unwrapped so the naive-means-UTC decision stays visible at each site. PostgreSQL is unaffected. On SQLite, rows the episode and cluster stores previously wrote from non-UTC-offset datetimes hold the local wall clock as if it were UTC and read back shifted before and after this change alike; the offset was never stored, so only new writes are correct, and a range query spanning the upgrade mixes the two conventions.

### Measurements

Layout comparison and scaled runs are in the design doc. Headlines (pgvector:pg16, one harness for all arms): tenant creation 0.006 ms vs 6-12 ms of per-tenant DDL; ingest and read parity or better at 40x2k and 1M-pair scales including a 500k-row tenant; O(1) delete at any size; churn 0 deadlocks vs 41-83.

Store-API ABAB against upstream main (3 interleaved rounds, one PG instance, medians, at the revision before the read fence was folded into the data statement): ingest +42% (11.1k vs 7.8k pairs/s), `delete_segments` +20%, event/derivative lookups 10-17% faster, windowed context expansion 8% faster (5.9 vs 6.4 ms), lifecycle create+open+delete 4.4x faster (2.9 vs 12.8 ms/cycle); seed context reads +0.19 ms (1.30 vs 1.11 ms) from the fence's separate registry round trip, since removed. Read-path ABAB of that fold (5 interleaved rounds, medians): seed context reads 1.17 vs 1.42 ms, event lookups 1.04 vs 1.61 ms, derivative lookups 1.05 vs 1.33 ms, windowed context expansion 8.74 vs 9.67 ms (8 rounds x 600 reps, paired median -0.91 ms); reads that find nothing unchanged. Server-side (`EXPLAIN ANALYZE`, best of 30) the `EXISTS` conjunct plans as a one-time InitPlan costing ~3 us per statement, so the client-side gain is the round trip. Purge drains ~237k rows/s when a partition's rows lie together on disk (12,000-row partitions, as the benchmark loaded them). With tenants' rows interleaved on disk, as concurrent ingestion lays them out, each deleted segment and its links sit on different pages, and the link cascade is 60-80% of a batch at one link per segment, about what deleting the links in bulk costs without a foreign key. Against the purge before the cursor (batches of `uuid IN (SELECT ... LIMIT n)` from the incarnation's first key, with a link check before retiring), on PostgreSQL 16 (128 MB shared buffers) with 200,000-row dead incarnations interleaved in a 3.8M-row table, the two alternating over 4 rounds under an interactive load, VACUUM and CHECKPOINT before each phase: a purge takes 9.1 vs 14.9 s (medians) and the call that retires the entry 4 vs 1,042 ms; concurrent reads run at 292 vs 308/s and writes at 67 vs 66/s during it, both within the rounds' spread (398 and 108/s idle); WAL per purge is 1.9 vs 1.8 GB; with an older snapshot held open, 5.3 vs 10.9 s. On a contiguous 12,000-row partition the cursor's extra statements cost ~8% (237k vs 257k rows/s). The store-API, read-path, and EventMemory end-to-end (ingest, queries, context expansion, mixed load) benchmarks show no difference between the two beyond run-to-run noise. Claim and mint-check statements measure 215/156 us.

### Compatibility

No migration from the partitioned layout is provided: the event backend is opt-in and pre-GA, and existing databases recreate their schema (see the design doc). The partition-key contract (`[a-z0-9_]`, max 32 bytes) is unchanged. The store supports PostgreSQL and SQLite, the two dialects the configuration can build and CI tests: the engine validator refuses any other dialect with the `ValueError` its other engine checks raise. `main`'s store sent other dialects down a generic path; this one does not work there (on MySQL 8.4 every purge fails with error 1235, since MySQL rejects `DELETE ... IN (SELECT ... LIMIT n)`, and create/delete churn deadlocks with error 1213, as it already did on `main`; on SQL Server, SQLAlchemy drops every locking clause and renders `JOIN LATERAL`, which T-SQL lacks; on Oracle, SQLAlchemy cannot render the JSON columns).

### Stack

| # | PR | change |
|---|---|---|
| — | #1661 **(this PR)** | Overhaul segment store: shared tables with incarnation-scoped tenant keys (port of #1548) |

| # | PR | change |
|---|---|---|

This PR is its 12 commits, `5a6559ef3`, `78b5b36de`, `273763688`, `c2503baf2`, `2c1305066`, `6845ad690`, `2bf775fc7`, `bc03bdb7d`, `01572175a`, `09ba90592`, `14ae7b5c4`, `7b014ac34` (the third drops the resource manager's closed gate, which guarded a get racing `close()` that the server's shutdown order never produces; the fourth refuses engines of dialects other than PostgreSQL and SQLite; the fifth has the tests stamp queue entries with the database's clock and arithmetic, as the store does, instead of Python times; the sixth purges in primary-key ranges, and the seventh reverts it: with fewer than a batch of rows left, its DELETE had no key bound and read every already-deleted row of the incarnation, making a purge's final call slower, 5.0-7.5 s against 0.84-1.5 s at 250,000 rows; the eighth purges from a cursor on the queue entry and makes the SQLite claim a write; the ninth leaves links to the foreign key's cascade, dropping the link check; the tenth adds tests that segment writes are atomic and that mixed operations strand nothing; the eleventh has the SQLite claim touch the queue row's incarnation, as the store's other SQLite write claims touch their row's key; the twelfth keeps an entry whenever a batch deleted at least the budget, so a batch that ever deleted more than it read could not retire an entry with rows left), directly on `main`, independent of the others: review and merge it in any order among the independent PRs (a later one touching the same file rebases). The stages' history carries the same change from [vector store scale-out 3/5] up, so their diffs include it until it merges.

### Verification

At its first 2 commits, on 2026-09-21, and at `273763688`, `c2503baf2`, `2c1305066` and `6845ad690`, on 2026-09-22 and 2026-09-23: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`), 1971 tests at `78b5b36de`, 1970 at `273763688` (the removed test is the one that pinned the gate) and 1971 at `c2503baf2`, `2c1305066` and `6845ad690` (the added test is the dialect check's, verified to fail without the check; the FIFO test, re-stamped by the database, still fails with the claim order reversed); the resource manager and segment store integration tests pass against PostgreSQL at `6845ad690`. At `bc03bdb7d`, `01572175a` and `09ba90592`, on 2026-09-23: `ruff check`, `ruff format --check` and `ty check` (as CI runs it) clean; the full server suite without integration tests passes, 1978, 1977 (the removed test is the link check's) and 1984 tests; the segment store tests pass on SQLite and against PostgreSQL at `09ba90592`, 100 each. At `14ae7b5c4`: `ruff check`, `ruff format --check` and `ty check` clean, 1984 tests without integration, and the SQLite segment store tests (100) pass. At `7b014ac34`: `ruff check`, `ruff format --check` and `ty check` clean; the segment store tests pass on SQLite and against PostgreSQL, 100 each. The new tests were checked against deliberately broken stores: the racing-purgers test fails with the SELECT claim restored on SQLite (4 of 6 rows left, 3 of 3 runs), the partial-add test fails when an add's segments commit apart from its links, and the model test fails when `delete_segments` deletes only one of its segments.


packages/server/src/memmachine_server/episodic_memory/event_memory/segment_store/sqlalchemy_segment_store.py
```diff
@@ -924,45 +1155,168 @@ async def close_partition(
...
+                    # entry.
+                    enqueued_at=func.now(),
+                )
+            )
+            await session.execute(
                 delete(PartitionRow).where(PartitionRow.partition_key == partition_key)
             )
 
-    # Helpers
+    @override
```
edwinyyyu: Duplicate `@override` (this line and the next). Same on the origin, #1548.

edwinyyyu (commented):
1. Duplicate decorator (line comment); inherited from #1548, flagged there too.
2. The commit trailer says "cherry picked from 91969730f", an intermediate commit on no pushed ref; #1548's merge commit is 7752e4cb7.

marvinyu-memverge (commented):
Reviewed at `dcf4ee35` as a port. I checked fidelity first, since that's what a port is worth: diff-of-diffs between `7752e4cb7^1..7752e4cb7` and `origin/main...dcf4ee35`, compared per file on added/removed line sets. **Thirty common files, zero differences.** Origin-only: `test_metrics_factory_wiring.py` and `uv.lock`, both matching your stated adaptations. Nothing appears in the port that isn't in the origin. The port is faithful.

Two asks, both about the body rather than the code.

**1. "The episode store gains only the `ensure_tz_aware` import" understates what that file does.**

`episode_sqlalchemy_store.py` also normalizes `created_at` to a UTC instant on the **insert** path, and normalizes **both** the `start_time` and `end_time` filter bounds. Those are behavioural changes to datetime handling in a store that isn't this PR's subject, and the adaptation list is where a reviewer decides which of the 30 files they can skip. The changes themselves look right to me and the design doc's "Datetime convention" consequence explains why the per-site spelling is deliberate -- it's the one-line summary I'd fix, not the code.

**2. "Purpose of the change" describes the one commit this diff deletes.**

The section is detach-before-drop and the `SHARE ROW EXCLUSIVE` table lock. Both are gone in what ships: `grep -i 'detach|management_lock|PARTITION OF'` on this branch's store returns nothing, while `origin/main`'s has `LOCK TABLE segment_store_pt IN SHARE ROW EXCLUSIVE MODE` at :802 and `CREATE TABLE ... PARTITION OF` at :1053 and :1059. The overhaul removes the partitioned layout and the machinery those earlier fixes added, which is exactly what the design doc's Consequences say. The **commit message** is fine and shouldn't change -- it's the origin's squash message and carries all four commits, which is what a faithful port should keep. It's the body that's been truncated to the first bullet, and that bullet is the most thoroughly superseded one. A reader of a 21-PR stack navigates by these bodies.

**On your two self-flagged items** -- not restating them, just adding a fact to each. The duplicate `@override` is inherited: the same pair sits at the same lines 1209-1210 in `7752e4cb7`, so fixing it here is a deliberate divergence from the origin rather than a port slip, which I think is worth doing but is your call as porter. And the trailer SHA `91969730f31c06ad08a4f880025a4f77850d6040` isn't merely non-canonical -- it doesn't exist in the repository at all, and isn't reachable from `speedkick` or this branch, so nobody can resolve it. `7752e4cb7` would be resolvable.

Verified while reading, so it's on record rather than re-derived next time: the fence is on **every** public path, with no gaps -- writes through `_lock_partition_for_write` (`add_segments` :346, `delete_segments` :784), reads carrying `_registry_row_query().exists()` in the data statement (:420, :509, :665, :690, :728, :757) plus the standalone registry read for the empty-result and windowed cases (:437, :474, :732, :761). The purge budget can't retire an entry on a spent budget: both decrements are guarded by an equality return, so `remaining >= 1` at the top of every iteration. The SQLite RETURNING dependency is enforced rather than assumed (`_MIN_SQLITE_VERSION = (3, 35)` in a validator, alongside the StaticPool and `:memory:` rejections). And the test surface is the part I'd point other people at -- 88 tests, with a named test for each of the three invariants I'd otherwise have asked you to pin, including `test_empty_incarnations_do_not_consume_row_budget` and `test_purge_batches_links_that_escaped_integrity`.

marvinyu-memverge (approved):
Re-reviewed at `78b5b36` (two commits on `main` `3809561`). Both body asks from my `dcf4ee35` review are addressed, and I have nothing new.

Since `main` moved four commits underneath, I compared the two ranges rather than the two heads. Commit 1 differs from `dcf4ee35` only where the rebase required it: `test_metrics_factory_wiring.py` is now part of the port, and its hunk is line-identical to `7752e4cb7`'s (the two attributes it sets, `_closed` and `_segment_store_purge_tasks`, are the ones `get_segment_store` touches at head); and the `test_resource_manager.py` import line is the union with #1683's. The three files both `main` and this PR touch (`database_manager.py`, `resource_manager.py`, `test_resource_manager.py`) have removed-line sets equal to the origin's, so the conflict resolution dropped nothing from #1679 or #1683. Your Verification claim holds as stated. The trailer now resolves, and commit 2 is the one decorator line.

Verified rather than re-read: round 1 read every file whole at `dcf4ee35`; this round proves the code at `78b5b36` differs from that by the lines above and nothing else. CI 48/48 at head, no bot findings at any head.

malatewang (approved):

edwinyyyu:
Doing final review rounds before merging.

edwinyyyu:
Claude found a potential performance issue. Reviewing.

edwinyyyu:
Two performance issues:
1. Postgres chooses a full scan for filtered expand_context (also a problem on main). Can be done by walking expansion in chunks.
2. Postgres can choose a full scan for purge. Can be fixed by refactoring queries.

Will implement proposed fixes and benchmark.

edwinyyyu:
Known performance issue, not addressed in this PR: on PostgreSQL, `get_segment_contexts` with a `property_filter` can read every row on the seed's side of its partition and sort them, taking seconds on large partitions (7.2-8.6 s per 20-seed request on a 1M-segment partition, 0.38-0.45 s on 20k-segment ones). `main`'s store sends the same query and gets the same plan, so this PR ports the problem rather than introducing it; the two have not been compared on identical data. The fix bounds the walk in SQL, which changes what a filtered read can return, so it belongs in a follow-up: #1712.

edwinyyyu:
**Purge changes in this PR since the approvals** (`bc03bdb7d`..`7b014ac34`)

**1. Each purge batch reread everything earlier batches had deleted.**
- **What happened:** a batch took the first n keys of the dead incarnation, starting from its first key every time. PostgreSQL keeps deleted rows, in the table and in its indexes, until vacuum. So each batch stepped over all rows already purged, visiting them to learn they were dead, and batches got more expensive as the purge went on. How much depended on the query plan and on whether an older transaction was open.
- **Fix:** the queue row carries a cursor, `purged_through`: the last segment key purged. A batch reads the next ≤ n keys after it, deletes that key range, and records its last key in the same transaction. No batch sees rows an earlier batch deleted.

**2. The call that retires an incarnation read all of it, plus the link table.**
- **What happened:** finding the incarnation empty meant reading all its deleted rows. The link check before retiring read every deleted link and, in the plan PostgreSQL chose, scanned the entire link table. At 1M rows that call took 9.5 s.
- **Fix:**
  - With the cursor, finding the incarnation empty is one index lookup.
  - The link check is removed: links go by the foreign key's cascade. The store already requires foreign-key enforcement of its engine and doesn't verify it.

**3. A SQLite race that the cursor would have introduced.**
- **What happened:** SQLite ignores `FOR UPDATE`, and its driver runs reads before the first write outside the transaction. Two purgers could therefore claim the same entry, and one could plan its batch from a stale cursor, delete nothing, and retire the entry with rows left.
- **Fix:** on SQLite the claim is an `UPDATE … RETURNING` of the queue row, the same primitive as the write fence and `delete_partition`. Purgers serialize at the claim. PostgreSQL keeps `FOR UPDATE SKIP LOCKED`.

**Also in this range:**
- **A defensive check:** a batch counts as full when `deleted >= remaining`.
- **An abandoned attempt:** batches bounded by a primary-key range (`6845ad690`), reverted in `2bf775fc7` because its last call had no key bound and read every deleted row.

**What the cursor relies on:** no row is written into an incarnation after it's enqueued. The write fence guarantees that, and forced tests on both backends cover it.

**Schema:** a new nullable column, `segment_store_gc.purged_through`. No migration is needed, since the queue table is new in this PR relative to main.

**Tests added:**
- purgers racing with the first one held mid-batch (fails with the old SELECT claim on SQLite);
- a purge call whose commit fails;
- incarnation sizes around the batch bound;
- a key re-created while its predecessor is purged;
- seeded runs of concurrent purgers and writers with exact end counts;
- link counts in every end-state purge test;
- data operations: failed or partial writes leave nothing, a stale handle's `delete_segments` raises, random operation sequences agree with a model, and mixed-operation churn completes with nothing stranded.

**Measured against the previous tip.** Setup: PostgreSQL 16; 200k-row dead incarnations with rows interleaved in a 3.8M-row table; the two alternating over 4 rounds under interactive load.

| | before | after |
|---|---|---|
| purge | 14.9 s | 9.1 s |
| retiring call | 1,042 ms | 4 ms |
| purge with an older transaction open | 10.9 s | 5.3 s |

- **Interactive load:** concurrent reads and writes during a purge stayed within the rounds' spread.
- **Other benchmarks:** store-API, read-path and EventMemory end to end show no difference.
- **One small cost:** a tiny purge whose rows are stored together is about 8% slower, from the cursor's extra statements.

**Where to look:**
- `purge_deleted_partitions` in `sqlalchemy_segment_store.py`: the claim branches, the cursor, and the retire condition;
- `PurgeQueueRow`;
- the purge section of `design/segment_store_shared_tables.md`.

**Known and not fixed here:** filtered context expansion (#1712, pre-existing on main).

PR #1663, opened 2026-09-16
[vector store contract 1/6] Answer with cosine scores and uuids, not vectors and stale properties, and require a vector on the record (port of #1598 and #1603)

edwinyyyu:
### Port

Copy of #1598 and #1603, merged into `speedkick` as `abf92a3a4` and `cdeef885f`, on `main`: the same two commits, re-derived on the one-collection store of [vector store scale-out 5/5] rather than cherry-picked, since the stores' data paths there are keyed by partition rather than by incarnation. The source intent is the same file by file, and the port's test bodies are spliced in by test name. Base-driven differences: the semantic manager declares no indexed properties, since the semantic storage writes vectors only; the search engines lose `get_vectors` and the gated test engines their delegations to it; the store tests observe a record's presence through `query`, since `get` is gone; #1603 adds the `Field` import its default needs and drops the assertion #1598 left dead. Context differs where the chain's base differs from speedkick's (#1671's sharding removal, #1630's `timeout=`, the `chunk_text_balanced` rename).

### Purpose of the change

Three things the vector store was doing that it is not the authority for.

**It carried four similarity metrics and shipped one.** Every embedder here produces vectors meant to be compared by cosine — OpenAI hard-coded it, Bedrock defaulted to it, SentenceTransformer only reported what the model declared — while every layer that touched a score paid for the other three in direction flags, threshold directions, and per-backend tables mapping the enum onto native metric names.

**It handed vectors back.** `VectorStoreCollection.get` had one production caller: the semantic storage read a feature's stored embedding back so it could write the same embedding again with fresh properties, because `upsert` demands a whole record. `return_vector=True` was passed there and nowhere else, and `VectorSearchEngine.get_vectors` existed to serve it.

**It handed properties back.** Both consumers were reading that copy to get from a search hit to a domain id, and what the copy cost them differed: semantic memory's was of mutable columns, only as fresh as the last write to it, while event memory's was an immutable uuid pairing that stayed correct but duplicated a mapping the segment store already served from an index.

### Description

`SimilarityMetric` is gone. Scores are cosine similarities in `[-1, 1]`, and the names say so: `QueryMatch.score` → `cosine_similarity`, `query(score_threshold=)` → `query(min_cosine_similarity=)`, which no longer needs a direction to be meaningful. The Bedrock embedder's `similarity_metric` config key goes with it, and the install and configuration docs drop it.

`get` is **removed**, with nothing offered in its place, and `VectorSearchEngine.get_vectors` goes with it. Semantic memory's read-modify-write is gone (see below), so nothing reads a stored vector back at all.

No scoring-by-id entry point takes its place. One would be needed if a caller assembled a candidate set outside the store and asked for those ids to be scored — what a selective-filter plan running *above* the store would do. Property filtering stays inside the store instead, so the candidate set stays there too, and a store-side regime reaches engine keys directly without a public method addressed by record UUID.

`QueryMatch` answers a `record_uuid` and a score. Properties stay **stored and filterable** — a filter is evaluated against the copy rather than trusted as the record — and are never returned. That ends `return_properties`, `Record` on the read path, and `set_properties`.

Each consumer now owns its own mapping:

- **Event memory** used a `_segment_uuid` property. The segment store already holds that mapping on the derivative's own row, non-null, under a primary key leading on exactly the columns the lookup filters — so the copy bought nothing an indexed read does not. `get_segment_uuids_by_derivative_uuids` mirrors the forward lookup that was already there.
- **Semantic memory** had nothing to reach for: its record uuid was `uuid5(namespace, feature_id)`, one-way, so the feature id rode along in the properties. It now owns a unique `vector_uuid` column and resolves hits through it. Every property that collection carried was a copy of a column on the feature row, never filtered on, so the payload goes entirely. The delete paths read the column *before* deleting the rows, since the row is what says which vector record a feature owns.

### Two consequences beyond the vector store

The vector graph stores carried a metric per stored embedding, as a companion property beside every vector. That is gone; `Node.embeddings` holds plain vectors.

NebulaGraph's `cosine()` cannot take `APPROXIMATE` and its vector indexes offer only L2 and IP. Cosine similarity between unit vectors *is* their inner product, so embeddings are normalized on the way in and compared with `inner_product()` against an IP index: cosine ranking that, unlike `cosine()`, can be approximate. Normalization happens in `_vector_to_gql_literal`, the one place any vector becomes a GQL literal, so the stored side and the query side cannot disagree. **Its tests skip everywhere, CI included** — they need a real NebulaGraph server rather than a testcontainer — so this is the least-exercised part of the change and wants a reviewer who can run against a licensed 5.x server.

### Breaking

Data does not survive this: existing semantic features have no `vector_uuid`, and existing collections carry properties nothing reads.

#### Purpose

`Record` is input-only since #1598: a collection stores vectors to search them and properties to filter on them, and answers a query with `QueryMatch`. Nothing reads a record back.

Its type never caught up. `vector: list[float] | None = None` says a caller may omit the vector, which no caller may do — so every store re-stated the rule at its own `upsert`, four copies of the same message:

```python
if record.vector is None:
    raise ValueError(f"Record {record.uuid} has vector=None, which is not allowed on input.")
```

`properties` was optional the same way, leaving Qdrant with `record.properties if record.properties is not None else {}`.

#### Description

- `vector: list[float]` — required.
- `properties: dict[str, PropertyValue] = Field(default_factory=dict)` — never `None`.
- The four runtime checks go, in sqlite, sqlite-vec, Qdrant and Milvus, along with Qdrant's `None` coalescing.
- Rejection moves to the model, so a caller meets it at construction rather than at a write.

#### Tests

The two `test_upsert_rejects_none_vector` tests asserted that a *store* raises. They now assert the model does — `pytest.raises(ValidationError)` on a `Record` built without a vector — plus a new test that `properties` defaults to `{}`. The four `_make_record` helpers require a vector.

### Stack

| # | PR | change |
|---|---|---|

| # | PR | change |
|---|---|---|
| [vector store contract 1/6] | #1663 **(this PR)** | Answer with cosine scores and uuids, not vectors and stale properties, and require a vector on the record (port of #1598 and #1603) |

This PR's own change is its 2 commits, `d31de1d53`, `67268cd10`; the rest of its diff is the PRs under it. Stacked on #1676; #1622 is stacked on it.

### Verification

At every commit from [sqlite store fixes 1/7] (`d7a6042b2`) to the tip `545ec401e`, on 2026-09-21: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`); 1971 tests at this PR's head `67268cd10`. Those commits were verified before the re-order that moved #1460 into the stage, which left their trees unchanged (`git diff` against the verified commits is empty). The 22 commits below `d7a6042b2` (the scale-out stage and the independent PRs' copies in its history), whose trees the re-order changed only by the absence of #1460's files, are being re-verified as this is written; this paragraph is replaced with the result.


packages/server/server_tests/memmachine_server/common/vector_store/test_sqlite_vector_store.py
```diff
@@ -1101,6 +989,19 @@ async def _fresh_store(db_path, tmp_path, *, save_threshold=1000):
     return store, engine
 
 
+async def _stored_record_uuids(engine, store) -> set[UUID]:
```
edwinyyyu, line 992: `_stored_record_uuids` is added with no caller: the origin's hunk on `test_a_reverted_publication_costs_search_not_records` (a #1589 test not yet on main) was dropped in the port, so the helper waits for #1469 above.

packages/server/src/memmachine_server/semantic_memory/storage/vector_store_semantic_storage.py
```diff
@@ -86,6 +82,10 @@ class VectorSemanticFeature(BaseVectorSemanticStorage):
     __tablename__ = "vector_semantic_feature"
 
     id = mapped_column(Integer, primary_key=True)
+    # The feature's record in the vector store. This is the mapping that lets a
+    # search hit be resolved back to a feature; the vector store holds no copy
+    # of anything this table owns.
+    vector_uuid = mapped_column(Uuid, nullable=False, unique=True, default=uuid4)
```
edwinyyyu, line 88: NOT NULL `vector_uuid` with a uuid4 default replaces the uuid5 `feature_vector_uuid`; existing semantic feature rows have no value and no migration is provided. #1598's body stated this under "Breaking"; this PR's body does not.

edwinyyyu (commented):
Two notes on the ported content (line comments): a helper with no caller, and a NOT NULL column with no migration that #1598's body disclosed under "Breaking" and this body does not.

PR #1664, opened 2026-09-16
[sqlite store fixes 2/8] Require a vector on the record type nothing reads back (port of #1603)

edwinyyyu:
### Port

Copy of #1603, merged into `speedkick` as `cdeef885f`, on `main`: the same commit cherry-picked, with one adaptation for merging: also removes the optional-vector handling the chain below kept in the Qdrant, Milvus and sqlite-vec stores and the in-memory test partition, which this change makes unnecessary.

### Purpose of the change

Require a vector on the record type nothing reads back

`Record` is written and never returned, so `vector` being optional described nothing a caller could do. Every store rejected `None` at its own `upsert` with the same message, four copies of one rule that the type could state once. `properties` was optional in the same way, so Qdrant carried `record.properties if record.properties is not None else {}`.

The vector is required and the properties default to `{}`. The four checks go, and rejection moves to the model, where a caller finds it at construction rather than at a write.

### Stack

One stack of 21 PRs on `main`, merged bottom-up: the vector store chain, whose first five already merged into `speedkick`, then the SQLite store fixes already merged into `speedkick`, re-derived on top of the chain. Every PR's diff on GitHub is cumulative until the PRs under it merge.

| # | PR | change |
|---|---|---|
| [vector store 1/15] | #1541 | fix(event-backend): make expand_context return timeline-neighbor episodes |
| [vector store 2/15] | #1661 | Overhaul segment store: shared tables with incarnation-scoped tenant keys (port of #1548) |
| [vector store 3/15] | #1663 | Answer with cosine scores and uuids, not vectors and stale properties (port of #1598) |
| [vector store 4/15] | #1670 | Remove per-project filterable properties (port of #1606) |
| [vector store 5/15] | #1671 | Remove custom sharding from the Qdrant store (port of #1654) |
| [vector store 6/15] | #1630 | Bound every request to a remote vector store by a configured timeout |
| [vector store 7/15] | #1622 | Create a session's storage with the session, never on a request |
| [vector store 8/15] | #1624 | Make no memory request create a project |
| [vector store 9/15] | #1625 | Remove open-or-create and close from both stores |
| [vector store 10/15] | #1626 | Rename logical collection to partition, and open to get, on both stores |
| [vector store 11/15] | #1627 | Make a vector store one collection, with string-keyed partitions |
| [vector store 12/15] | #1631 | Mint an incarnation per partition life in a SQL-arbitrated registry; delete logically, reclaim by purge, and keep the tombstone until a clean sweep and a retention |
| [vector store 13/15] | #1618 | Let a deployment tune a Qdrant collection's HNSW, optimizers and quantization |
| | | *The SQLite store fixes, stacked on [vector store 13/15]:* |
| [sqlite store fixes 1/8] | #1460 | Publish vector index files atomically (but not durably) (as merged into `speedkick`, #1588) |
| [sqlite store fixes 2/8] | #1664 **(this PR)** | Require a vector on the record type nothing reads back (port of #1603) |
| [sqlite store fixes 3/8] | #1469 | Never reuse a row id in SQLiteVectorStore (as merged into `speedkick`, #1589) |
| [sqlite store fixes 4/8] | #1672 | Own the search engine's concurrency in the store, not in each engine (port of #1612) |
| [sqlite store fixes 5/8] | #1673 | Serialize a collection's writes so the engine sees them in order (port of #1607) |
| [sqlite store fixes 6/8] | #1674 | Refuse a pending row replay cannot honor, instead of dropping it (port of #1608) |
| [sqlite store fixes 7/8] | #1675 | Take SQLite's write lock at BEGIN, not at the first write (port of #1609) |
| [sqlite store fixes 8/8] | #1676 | Give every write a fresh row id, so a key names one version (port of #1610) |
| | | *Parked on `speedkick` until #1597 (EventMemory session, source and expansion) reaches `main`:* |
| [vector store 14/15] | #1628 | Make a vector store filter only on the properties it declares |
| [vector store 15/15] | #1616 | Close the filter union, and make negation the complement on every backend |

This PR's own change is its last commit, `aab326d86`; the rest of its diff is the PRs under it. Stacked on #1460; #1469 is stacked on it.

### Verification

`ruff check`, `ruff format --check` and `ty check` clean at every commit; `pytest packages/server/server_tests -m 'not integration'`: 1927 passed, 3 skipped on the stack's tip (`595707eb` on `main` `8101f141`) and 1887 passed on the chain's tip (`a631b173`), with the vector store and resource manager suites run at every commit; `pytest packages/client/client_tests`: 255 passed. 2026-09-17.


edwinyyyu:
Folded into #1663 as its second commit: requiring the vector follows from #1598 making a record write-only, and every file it touches is in #1598's diff. The chain is now built on the required-vector Record, as speedkick was.

PR #1449, opened 2026-06-11
Improve embedding fidelity

edwinyyyu:
### Purpose of the change

Chunk-merged embeddings (texts longer than the model input limit are split, embedded per chunk, and averaged) reproduce the whole-text embedding more faithfully with **greedy max-size chunks and a length-weighted average** than with the current balanced chunks and unweighted mean. Measured on three encoders; numbers below.

The special-token sanitizer in the OpenAI embedder is removed. It was a workaround for a server-side bug at api.openai.com (text-embedding-3-small returned 500 on inputs containing tokens such as `<|endoftext|>`), and that bug is gone: 7 tokens x 3 models x 3 input forms all return 200 with valid embeddings (probe script and results attached as a PR comment below). The workaround was not free: it deleted the literal token from the text before embedding, so a stored transcript that quotes `<|endoftext|>` was embedded as if it did not. It is also OpenAI-specific by design. `OpenAIEmbedder` is also used with OpenAI-compatible endpoints (Ollama, DashScope), which the probe does not cover; an endpoint that fails on a plain string containing such a token is broken at the endpoint, and the remedy is to not use it, not for MemMachine to pre-edit input for every provider that might mis-tokenize.

`max_input_length` above the OpenAI embedder's per-request cap (75,000 code points) is now clamped to the cap. A limit above the cap can never be honored (no chunk larger than the cap fits in any request), and before this change it reached `cluster_texts` with an oversized chunk and raised `ValueError` (HTTP 500), the #1298 failure for that configuration. Greedy chunking widened which input lengths hit it: at `max_input_length=80_000`, balanced chunking crashed on 76,000-80,000 and 160,000-char inputs; greedy crashed on everything above 75,000. No shipped sample config sets a limit above the cap.

### Description

- Use greedy max-size chunks (`chunk_text`) instead of balanced chunks (`chunk_text_balanced`) in the OpenAI, SentenceTransformer, and Amazon Bedrock embedders.
- Merge chunk embeddings with a length-weighted `np.average` instead of unweighted `np.mean`.
- Remove the special-token sanitizer from the OpenAI embedder; keep the empty-input coercion (`or "."`), which is independently load-bearing: the API rejects empty strings and `chunk_text("")` yields no chunks to average.
- Clamp `max_input_length` to `max_total_input_length_per_request` in the OpenAI embedder, with a regression test that fails before the clamp (`ValueError` from `cluster_texts`) and passes after.

### Measurements

[...]

PR #1666, opened 2026-09-16
[speedkick port 12/17] Serialize a collection's writes so the engine sees them in order (port of #1607)

### LINKED ISSUE #1468: SQLiteVectorStore: row_id reuse + concurrent async writes silently lose a record's vector

edwinyyyu:
### Port

Copy of #1607, merged into `speedkick` as `3e816841d`, on `main`: the same commit cherry-picked, with no adaptation.

### Purpose of the change

Serialize a collection's writes so the engine sees them in order

A write commits to SQLite and only then applies to the search engine, so two writers to one uuid could reach the engine in the opposite order to the one they committed in: an upsert overtaking a delete re-adds a vector for a record that is gone, and two upserts inverting leave the engine serving the older vector. Never reusing a row id does not cover this, because an upsert of an existing uuid keeps its row id.

A save in that window costs a write outright: it publishes the index and trims every applied log row, and a write that applied after the index was written is then in neither.

A per-collection asyncio.Lock now spans a write from SQL commit through engine apply, mark-applied, and any save it triggers; shutdown's save takes it too. The lock belongs to the store, not to a collection handle: a handle is constructed per open_collection call, so several can address one collection, and only a shared lock serializes them. Readers are untouched.

Three tests fail without the lock, each interleaving made deterministic by gating the engine: an upsert overtaking a delete of its uuid, a save trimming a write it did not publish, and that overtake across two handles, which a per-handle lock passes. The rest pin behavior the lock must preserve: an upsert surviving a delete of another record, disjoint concurrent upserts and deletes, writes racing a checkpoint, and batches that name one uuid twice.

Fixes #1468.

Claude-Session: https://claude.ai/code/session_01ESpWYTmCR7X3bJEpoA8SAn

Co-authored-by: Claude Fable 5.1 <noreply@anthropic.com>

### Stack

| # | PR | change |
|---|---|---|
| [speedkick port 12/17] | #1666 **(this PR)** | Serialize a collection's writes so the engine sees them in order (port of #1607) |
| [speedkick port 14/17] | #1668 | Take SQLite's write lock at BEGIN, not at the first write (port of #1609) |
| [vector store 11/14] | #1657 | Check liveness after every remote operation; keep a deleted partition's tombstone until a clean sweep and a retention |

This PR's own change is its last commit, `890b05a14`; the rest of its diff is the PRs under it. Stacked on #1665; #1667 is stacked on it.

### Verification


edwinyyyu:
Closing: the SQLite vector stores will be removed from main and re-added at their speedkick state once the vector store chain has landed, so this port is not needed.

PR #1668, opened 2026-09-16
[speedkick port 14/17] Take SQLite's write lock at BEGIN, not at the first write (port of #1609)

edwinyyyu:
### Port

Copy of #1609, merged into `speedkick` as `6f3957753`, on `main`: the same commit cherry-picked, with no adaptation.

### Purpose of the change

Take SQLite's write lock at BEGIN, not at the first write

`delete` resolves row ids and then writes, in one transaction. Under SQLite's default deferred BEGIN the write lock is taken at the first write, so another writer can take it between that read and the write, and the transaction that read first is the one that loses: its own write cannot upgrade, and SQLite reports that immediately rather than waiting out busy_timeout, because waiting could only deadlock. Reproduced in both journal modes:

journal_mode  BEGIN      other writer  our write delete        deferred   shut out      fails delete        IMMEDIATE  shut out      ok wal           deferred   commits       fails wal           IMMEDIATE  shut out      ok

The store now emits BEGIN explicitly and lets a transaction ask for BEGIN IMMEDIATE, which every write path does. The mode is chosen per transaction, not per engine: a hook that asked for IMMEDIATE unconditionally would make every read take the write lock, and two readers would then serialize against each other.

Two tests fail without it. One holds a competing lock across a read-then-write transaction and asserts that transaction completes; asserting instead that the other writer is excluded passes either way, because a deferred BEGIN shuts it out too, later and by a different lock. The other races four creates of one name: under a deferred BEGIN the losers read no stored config, go on to CREATE TABLE, and fail there with "table already exists" instead of VectorStoreCollectionAlreadyExistsError.

Claude-Session: https://claude.ai/code/session_01ESpWYTmCR7X3bJEpoA8SAn

Co-authored-by: Claude Fable 5.1 <noreply@anthropic.com>

### Stack

| # | PR | change |
|---|---|---|
| [speedkick port 12/17] | #1666 | Serialize a collection's writes so the engine sees them in order (port of #1607) |
| [speedkick port 14/17] | #1668 **(this PR)** | Take SQLite's write lock at BEGIN, not at the first write (port of #1609) |
| [vector store 11/14] | #1657 | Check liveness after every remote operation; keep a deleted partition's tombstone until a clean sweep and a retention |

This PR's own change is its last commit, `141ccc148`; the rest of its diff is the PRs under it. Stacked on #1667; #1669 is stacked on it.

### Verification


edwinyyyu:
Closing: the SQLite vector stores will be removed from main and re-added at their speedkick state once the vector store chain has landed, so this port is not needed.

Issue #1564, opened 2026-09-01
Commit Qdrant to payload-partitioned multitenancy: drop per-collection shard keys, retire the in-process lock table, recover fencing and reclamation


edwinyyyu:
Qdrant offers two ways to separate tenants, and they have opposite cost profiles. This issue tracks committing to the scalable one and finishing it, rather than leaving both half-supported.

Measured against Qdrant 1.19.0, one hundred tenants each way:

| | shard key per collection | payload partitioning |
| --- | --- | --- |
| admitting a tenant | 450 ms, explicit API call | no call at all |
| 100 tenants | 45.0 s | 0.3 s including 10,000 points |
| segments | 505 | 5, and still 5 after delete and reuse |
| deleting one tenant | O(1) drop | 9 ms filter-delete for 100 points |
| fencing of stale writes | yes, `Shard key "t1" not found` | none |
| cluster mode | required, plus a bootstrap `--uri` | not required |

Payload partitioning is the only one that scales, because a tenant is a value rather than a physical structure. Everything below follows from choosing it.

### 1. Stop creating a shard key per logical collection

`QdrantConf.is_distributed` switches to custom sharding and gives every logical collection its own shard key:

```python
sharding_method=models.ShardingMethod.CUSTOM if self._is_distributed else None
...
if self._is_distributed:
    await self._ensure_shard_key(entry.native_collection_name, entry.partition_key)
```

The cost is per logical collection. A deployment serving thousands of tenants — a tenant being a user, or a group of users — has at least that many logical collections and in practice more, since a tenant's memory is typically partitioned further. At a thousand logical collections that is ~7.5 minutes of shard-key creation and ~5,000 segments; at ten thousand, ~75 minutes and ~50,000 segments, before any data is written.

Qdrant's own guidance recommends against per-tenant physical structures: "Creating a separate collection for each tenant is rarely the most efficient approach. Each collection carries its own resource overhead."

The flag's docstring sells the O(1) delete, which is genuine, but it is paid for per tenant. Note the two mechanisms are layered correctly — the payload partition key is written and filtered in both modes, with the shard key passed additionally — so this is a policy problem, not a correctness one.

If a container ever needs splitting, the right lever is `shard_number > 1` with automatic sharding: a provisioning knob, no custom sharding, no per-tenant cost.

### 2. Retire the per-process lock table

`_name_locks` serialises lifecycle operations within one process, which was the only arbitration available before a cross-process registry existed. It also leaks:

```python
[...]

PR #1671, opened 2026-09-16
[vector store scale-out 1/5] Remove custom sharding from the Qdrant store (port of #1654)

edwinyyyu:
### Port

Copy of #1654, merged into `speedkick` as `779362c61`, on `main`: the same commit cherry-picked, with two differences for main: `get`, which `main` still has, also loses its shard-key selector, and the removed sharding tests are removed in `main`'s form.

### Purpose of the change

The Qdrant store could shard its native collection by logical collection (`QdrantConf.is_distributed`, CUSTOM sharding, one shard key per logical collection, a shard-key selector on every operation) so that a logical collection could be deleted by dropping its shard. The payload partition key was written and filtered in both modes, with the shard key on top, so no query changes here; what changes is the cost of a tenant.

A shard key is a physical structure, and its cost is per tenant. Measured in #1564 (Qdrant 1.19.0, one hundred tenants): admitting a tenant is an explicit `create_shard_key` call of about 450 ms, 45 s for the hundred against 0.3 s with payload partitioning alone, including 10,000 points; 505 segments against 5, and still 5 after deleting and reusing tenants; and cluster mode is required, with a bootstrap `--uri`. At ten thousand tenants that is over an hour of shard-key creation and some 50,000 segments before a point is written. Qdrant's own guidance says the same: a physical structure per tenant is for the few oversized ones, the long tail is a payload value.

What the shard bought was the O(1) delete, and with it a kind of fencing (a write to a dropped shard fails). Deletion is about to become a registry write that is O(1) and atomic as seen by every reader, a stale handle is fenced by that registry, and the points are reclaimed afterward by a filter-delete off the request path, so a shard per logical collection would only add its cost; it goes now, on the current shape, so the later changes do not carry it. Promoting a single oversized tenant to its own custom-sharded collection later, Qdrant's tiered arrangement, stays open: it is a separate collection, not a flag on this one.

`is_distributed` was never documented. A configuration that still names it is ignored, as any unknown key under a database `config` is (`QdrantConf` does not forbid extras); nothing reads it.

### Stack

| # | PR | change |
|---|---|---|

| # | PR | change |
|---|---|---|
| [vector store scale-out 1/5] | #1671 **(this PR)** | Remove custom sharding from the Qdrant store (port of #1654) |

This PR is its one commit, `f2436a700`, directly on `main`; #1631 is stacked on it.

### Verification

At its one commit, on 2026-09-21: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`), 1914 tests at its head `f2436a700`.


edwinyyyu (commented):
1. "`is_distributed` was never documented; a configuration naming it is rejected": the second half does not hold. `SupportedDB.build_config` does `conf_cls(**conf)`, no configuration model sets `extra="forbid"`, and pydantic's default is `extra="ignore"`, so `SupportedDB.QDRANT.build_config({"host": "h", "registry_database": "r", "is_distributed": True})` returns a `QdrantConf` with no error: the key is silently ignored. Same sentence on the origin, #1654.
2. The commit trailer "cherry picked from 408b0a9d" names a local chain-main3 commit; #1654's merge commit is 779362c61.

marvinyu-memverge (approved):
Approve at f2436a70. One non-blocking ask, one observation.

1. The commit message still carries the origin's sentence "is_distributed was never documented; a configuration naming it is rejected", which your own self-review on 09-17 corrected: no conf model sets extra="forbid" (0 hits for extra= or model_config under common/configuration/ at head), so build_config's conf_cls(**conf) silently drops the key. The PR body already says "ignored"; the commit is what git log keeps. Worth amending to the body's wording on the next push, and the title too, since the commit still says "[vector store 2/13]" where the PR is "[vector store scale-out 1/5]".

Observation, not an ask: a collection created under is_distributed=True keeps CUSTOM sharding on the Qdrant side, and Qdrant requires a shard-key selector on an upsert to a custom-sharded collection, so a deployment that had set the flag would fail every write after this lands until it recreates its collections. The flag was undocumented and needed cluster mode with a bootstrap --uri, so I do not expect a producer; noting it in case an upgrade note exists for the speedkick deployment.

Verified: diff-of-diffs against 779362c6 (git diff 779362c6^ 779362c6 vs origin/main..refs/pr/1671, changed lines sorted) differs in exactly the two stated adaptations: get's shard_key_selector, and the removed sharding tests in main's form (coll.get / record.uuid / the COSINE lines where the origin removed _present_uuids). git grep at head for is_distributed|shard_key|_ensure_shard_key|ShardingMethod|distributed_qdrant: 0 hits repo-wide, docs and sample configs included (main: 37 hits, all inside this PR's seven files). delete_collection now takes the pre-existing FilterSelector(_partition_filter(name)) path unconditionally. CI green at head, no threads.

malatewang (approved):

PR #1675, opened 2026-09-17
[sqlite store fixes 6/7] Take SQLite's write lock at BEGIN, not at the first write (port of #1609)

edwinyyyu:
### Port

Copy of #1609, merged into `speedkick` as `6f3957753`, on `main`: re-derived onto the one-collection store of #1627. The begin hooks are module-level and registered once per engine, since #1627 gives every SQLite store of one backend one engine; every write transaction of the store takes the immediate `BEGIN`, the partition registry's included.

### Purpose of the change

`delete` resolves row ids and then writes, in one transaction. Under SQLite's default deferred `BEGIN` the write lock is taken at the *first write*, so another writer can take it between that read and the write, and the transaction that read first is the one that loses: its own write cannot upgrade, and SQLite reports that immediately rather than waiting out `busy_timeout`, because waiting could only deadlock.

Measured in both journal modes, so this is not a WAL-only concern:

| `journal_mode` | `BEGIN` | other writer | our write |
|---|---|---|---|
| `delete` | deferred | shut out | **fails** |
| `delete` | `IMMEDIATE` | shut out | ok |
| `wal` | deferred | commits | **fails** |
| `wal` | `IMMEDIATE` | shut out | ok |

The store now emits `BEGIN` explicitly and lets a transaction ask for `BEGIN IMMEDIATE`, which every write path does. The mode is per transaction, not per engine: a hook that asked for `IMMEDIATE` unconditionally would make every read take the write lock, and two readers would then serialize against each other. This makes read-then-write atomic by construction rather than by every write transaction happening to write first.

### Tests

`test_a_write_transaction_can_still_write_after_it_has_read` fails without it: it holds a competing lock across a read-then-write transaction and asserts that transaction completes. Two earlier versions passed either way: asserting that the *other* writer is excluded passes because a deferred `BEGIN` shuts it out too, later and by a different lock; letting the other writer's context manager exit before our write passes because rollback releases the lock first. `test_racing_creates_admit_exactly_one` races four creates of one name and asserts one wins and the rest are refused with `VectorStorePartitionAlreadyExistsError`. On this store it passes either way, because `create_partition`'s first statement is the registry insert and a deferred `BEGIN` takes the write lock there; it stays as the contract's pin. (Its origin discriminated on `speedkick`'s `create_collection`, whose losers read no stored config and reached `CREATE TABLE`.)

### Stack

| # | PR | change |
|---|---|---|

| # | PR | change |
|---|---|---|
| [sqlite store fixes 6/7] | #1675 **(this PR)** | Take SQLite's write lock at BEGIN, not at the first write (port of #1609) |

This PR's own change is its one commit, `c2c65b70c`; the rest of its diff is the PRs under it. Stacked on #1674; #1676 is stacked on it.

### Verification

At every commit from [sqlite store fixes 1/7] (`d7a6042b2`) to the tip `545ec401e`, on 2026-09-21: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`); 2057 tests at this PR's head `c2c65b70c`. Those commits were verified before the re-order that moved #1460 into the stage, which left their trees unchanged (`git diff` against the verified commits is empty). The 22 commits below `d7a6042b2` (the scale-out stage and the independent PRs' copies in its history), whose trees the re-order changed only by the absence of #1460's files, are being re-verified as this is written; this paragraph is replaced with the result.


packages/server/server_tests/memmachine_server/common/vector_store/test_sqlite_vector_store.py
```diff
@@ -1119,52 +1178,493 @@ async def _set_all_pending_operations_unapplied(engine) -> None:
...
+        coll2 = await store2.get_partition(NAME)
+        assert coll2 is not None
+        found = set(await _present_uuids(coll2, [record.uuid for record in records]))
+        assert found == {record.uuid for record in records}
+
+        await store2.shutdown()
+        await engine2.dispose()
+
+    @pytest.mark.asyncio
+    async def test_racing_creates_admit_exactly_one(self, store):
```
edwinyyyu, line 1516: On main this test passes without the change: the INSERT into the registry is the transaction's first statement, so a deferred BEGIN already takes the write lock there, and the losers' IntegrityError maps to `VectorStorePartitionAlreadyExistsError` either way. Verified at b7d2bb461 with `_begin` patched to always emit a plain `BEGIN`: this test passes and `test_a_write_transaction_can_still_write_after_it_has_read` fails. The body's "two tests fail without it" is one on main, and the CREATE TABLE race it describes is speedkick's `create_collection`, not `create_partition`.

edwinyyyu (commented):
1. `test_racing_creates_admit_exactly_one` does not discriminate on main (line comment; verified).
2. Adaptation the body does not list: `_register_sqlite_begin` is made idempotent across stores sharing an engine (`_disable_implicit_begin`/`_begin` hoisted to module level, `event.contains` guard, docstring "Stores sharing an engine register once"). It is required by #1627's one engine per backend and is absent from 6f3957753.

PR #1678, opened 2026-09-17
fix(metrics): make Prometheus metrics correct with multiple workers, and wire the trackers up (port of #1523 to main)

wanghy73:
> **Port of #1523 from `speedkick` to `main`.**
>
> Cherry-pick of `ace7752184a85d45ad5b34d6f3cf7c1c20f0c948`, the squash commit #1523 merged as on
> byte-identical to the one reviewed on #1523 (checked with `git patch-id`), so the
> review there still stands.
>
> - `ruff check` and `ruff format --check` clean repo-wide on the pinned ruff 0.16.8.
> - Server unit suite: **1887 passed**, 3 skipped, against **1869 passed** on `main` —
>   the 18 added are this PR's two new test modules, and nothing regressed.
>
> One extra commit on top, test-only: `main` runs `ty check` over the server
> package (`speedkick` did not when #1523 merged), and it rejects the
> `get_sql_engine` stub in the new test. It now patches the class via
> `monkeypatch` with the real method's signature instead of assigning to the
> instance. `ty` passes on 3.12 and 3.14. I confirmed the test still fails when
> the segment-store wiring is reverted, so it keeps its teeth.
>
> The body below is #1523's, unchanged.

---

Two fixes to metrics that are wrong in ways nothing reports. Both were found while
trying to attribute request latency on a capacity run and discovering the numbers
could not be trusted.

## 1. Metrics are not aggregated across uvicorn workers

`/metrics` called bare `generate_latest()`. With `MEMMACHINE_WORKERS > 1` uvicorn
forks a worker per process and each keeps its own registry, so a scrape returned
whichever worker happened to answer — about 1/N of the traffic, a different 1/N
each time. Consecutive scrapes look like counter resets, and every rate, histogram
quantile and calls-per-request figure derived from them is wrong in an unbounded
direction.

Build a fresh registry per scrape and attach `MultiProcessCollector` when
`PROMETHEUS_MULTIPROC_DIR` is set. The single-worker path keeps the default
registry, so nothing changes for the default deployment.

**Measured**, on an 8-tenant capacity run over one rate ladder:

| workers | counter resets before | after |
|---|---|---|
| 4 | 70 | **0** |
| 8 | 87 | **0** |

Five consecutive scrapes of a loaded server returned identical values where they
previously diverged.

This needs `PROMETHEUS_MULTIPROC_DIR` set and a writable directory mounted at it.
The chart change that does so is separate.

## 2. OperationTracker silently discards timings when given no factory

`OperationTracker` accepts `metrics_factory=None` and then throws away every
timing it takes — no error, no warning, no series. A component that is fully
instrumented but never handed a factory is indistinguishable from one that was
never instrumented at all, and the only way to notice is to go looking for a
metric that should exist and find nothing.

That is not hypothetical: the Neo4j store, the episode store and the session
store each shipped instrumented and unwired, which is why database latency
appeared to be unmeasurable. Every call was being timed and discarded. This wires
the factory through the resource manager, the database manager and the event
backend's params.

### On the tests

`test_metrics_factory_wiring.py` asserts at the **call sites**, not on the
components. An earlier version tested that each store honours a factory it is
given — which passes whether or not anything passes one, and still passed with
the fix reverted. The tests here fail when the wiring is removed.

## Verification

- All 9 changed Python files parse; the new test module was written against the
  running server.
- Deployed to a test cluster as `memmachine/memmachine:instrumented-mp` and
  exercised with 1, 4 and 8 workers under load.

The two changes are kept as separate commits so either can be reverted alone.


edwinyyyu (approved):
Does not touch core/storage so any changes are easily undoable. Not reviewed for better approaches.

packages/server/src/memmachine_server/server/app.py
```diff
@@ -144,6 +128,79 @@ def start_http() -> None:
...
+    With more than one worker and no directory configured, one is chosen here
+    rather than left unset: without it every worker keeps its own unaggregated
+    registry and a scrape returns whichever worker answered, which reads as a
+    plausible number rather than an error. The path is fixed so all workers
+    share it; two servers on one host must set the variable themselves.
+    """
+    path = os.environ.get("PROMETHEUS_MULTIPROC_DIR")
+    chosen = False
+    if not path:
+        if _worker_count() <= 1:
```
malatewang, line 183: Should move this check out of the if code block? If there is only one worker, there is no sense to setup the directory.

wanghy73: Checked this one rather than moved it, because hoisting the check turns out to break a case.

`prometheus_client` picks its multiprocess value class from the presence of `PROMETHEUS_MULTIPROC_DIR` **alone** — the worker count never enters into it — and then mmaps a file into that directory on the first metric. So a one-worker deployment where the operator has set the variable still needs the directory to exist and to be clear of the previous run’s files.

Verified directly against the library, pointing the variable at a missing directory in a single process:

```
ValueClass chosen: MmapedValue
raised FileNotFoundError: [Errno 2] No such file or directory:
  /tmp/definitely-not-created-probe/counter_4084679.db
```

That is the failure the docstring warns about, and moving `if _worker_count() <= 1: return` to the top of the function would reintroduce it for exactly that configuration.

Your underlying point is right though, and it is already what the code does: the worker count gates *choosing a default* directory, which is why the check sits inside `if not path`. `test_an_explicit_directory_is_honoured_for_one_worker` pins the other half.

I have added a comment at that line in be659b8 recording the reasoning, so the next reader does not have to rediscover it. Happy to restructure if you would rather see it expressed differently.

malatewang (approved):

PR #1679, opened 2026-09-17
fix(db): bound asyncpg statement and connect time so a dead socket recovers (port of #1552 to main)

wanghy73:
> **Port of #1552 from `speedkick` to `main`.**
>
> Cherry-pick of `10b3706d5574080153a62b678196989dbc24f85f`, the squash commit #1552 merged as on
> byte-identical to the one reviewed on #1552 (checked with `git patch-id`), so the
> review there still stands.
>
> - `ruff check` and `ruff format --check` clean repo-wide on the pinned ruff 0.16.8.
> - Server unit suite: 1869 passed, 3 skipped — identical to `main`. This PR adds no
>   tests; the fault-injection evidence below is from #1552 and was not re-run here.
>
> The body below is #1552's, unchanged.

---

A pooled connection whose peer stops responding is not an error, it is a wait.
The kernel retransmits with exponential backoff — observed at `tcp_retries2`
attempt 13, `Send-Q` stuck at 34 bytes and unchanged over 20 seconds — and
asyncpg has no deadline of its own, so the request blocks for the whole of it.

`pool_pre_ping` does not help: its liveness `SELECT` is written into the same
dead socket and waits with everything else.

**The failure this produces is worse than a slow request.** The pool never
discards the dead connections, so the process stays wedged *after the network
recovers* and only a restart clears it. Seen twice on a benchmark deployment:
searches timing out indefinitely while `/api/v2/health` answered in about a
millisecond, four uvicorn workers idle in `select()`, no lock contention in
Postgres, and a fresh connection from the same pod completing in **18 ms**.

`create_async_engine` was passed no `connect_args`, so nothing bounded either the
statement or the connect. This adds `command_timeout` and `connect_timeout`,
defaulted to 60 s and 10 s and settable to null for the old behaviour. Both are
gated on `driver == "asyncpg"`: they reach `asyncpg.connect()` and mean nothing to
aiosqlite or aiomysql.

The keyword assembly moves into a helper, because adding a branch to
`async_get_sql_engine` pushed it past ruff’s complexity limit and the repeated
"if not None" lines were asking for it.

## Verified by fault injection

`iptables DROP` on the pod’s traffic to Postgres:5432, so pooled connections wedge
exactly as they did in the incident:

| | under the fault | **after the fault is removed** |
|---|---|---|
| stock | no response at 150 s (cap) | **still hung, no response at 90 s** |
| with this change | 500 after 194 s | **200 in 1.5 s** |

**The recovery is the point.** The stock build stays wedged once the network is
healthy again, which is what forced the restarts.

## What this does not claim

Under a *total* outage the request still takes ~194 s to fail rather than ~60 s.
A search touches three separate stores, so it pays a timeout per store: the bound
is roughly `command_timeout` × stores touched, not `command_timeout`. Lowering the
default would shorten it proportionally.

**60 s is chosen, not tuned.** It is generous against real query times here, which
are milliseconds, and is left configurable rather than optimised. Reviewers who
know the workload better should push back on it.

## Caveats for review

- **Not built and tested from this branch alone** — the image used for the fault
  injection carried this change and the startup-warming fix together, on top of
  the Qdrant wiring from #1532.
- `ruff check` and `ruff format --check` pass.


packages/server/src/memmachine_server/common/resource_manager/database_manager.py
```diff
@@ -45,6 +46,33 @@
...
+            kwargs[attr] = value
+
+    # Gate on the driver, not the dialect: these reach asyncpg.connect() and
+    # mean nothing to aiosqlite or aiomysql.
+    if conf.driver == "asyncpg":
+        connect_args: dict[str, float] = {}
+        if conf.command_timeout is not None:
+            connect_args["command_timeout"] = conf.command_timeout
+        if conf.connect_timeout is not None:
+            connect_args["timeout"] = conf.connect_timeout
```
edwinyyyu, line 73: Unqualified key "timeout". Please rename.

packages/server/src/memmachine_server/common/configuration/database_conf.py
```diff
@@ -447,6 +447,25 @@ class SqlAlchemyConf(YamlSerializableMixin, PasswordMixin):
             "that were reset server-side. Internal default is False."
         ),
     )
+    command_timeout: float | None = Field(
```
malatewang, line 450: Add these two new fields to the sample configuration

packages/server/src/memmachine_server/common/resource_manager/database_manager.py
wanghy73: This one I could not rename: `timeout` is asyncpg’s own parameter name, not a name this PR chose.

`connect_args` is handed to `asyncpg.connect()`, whose signature has `timeout` for the connection-establishment deadline and no `connect_timeout` at all:

```
timeout:         PRESENT
command_timeout: PRESENT
connect_timeout: absent
```

So renaming the dict key would just mean asyncpg never receives the setting.

The configuration field *is* qualified — it is `connect_timeout` in `SqlAlchemyConf`, and only the wire name is bare, at the single point where we translate to their API. I have added a comment there in e02d300 so the next reader does not have to go and check asyncpg’s signature to learn that.

If what you would prefer is a local alias making the translation explicit at the call site, say so and I will add it.

packages/server/src/memmachine_server/common/configuration/database_conf.py
wanghy73: Done in e02d300. Both fields now sit beside the `pool_*` settings in `deployments/helm/values.yaml` and are rendered into the server config by `memmachine-configmaps.yaml`, following the existing pattern:

```yaml
  pool_pre_ping: false
  command_timeout: 60   # asyncpg only: seconds a statement may run before it is cancelled; null for unbounded
  connect_timeout: 10   # asyncpg only: seconds to wait for a new connection; null to wait indefinitely
```

One thing I checked rather than assumed: that the documented “set to null to restore the unbounded behaviour” is actually reachable *from the chart*. Setting either value to null renders an empty value, which YAML reads as null, which the config model accepts and which then omits `connect_args` entirely. Confirmed by rendering the chart both ways and feeding the output through `SqlAlchemyConf` and `_sql_engine_kwargs`:

```
null  -> command_timeout=None connect_timeout=None -> NO connect_args (unbounded)
60/10 -> {"command_timeout": 60.0, "timeout": 10.0}
```

`helm lint` is clean.

edwinyyyu (approved):

packages/server/src/memmachine_server/common/configuration/database_conf.py
malatewang: add them to the sample config too.

wanghy73: You are right — I had only added them to the Helm chart, which is a different file. Now added to the sample configs proper in f9b7268: the `profile_storage` block of all three of `episodic_memory_config.cpu.sample`, `.gpu.sample` and `.nebula.sample`, beside the `pool_*` settings and in the same comment style.

```yaml
        pool_pre_ping: false  # test connections for liveness before use (default: false)
        command_timeout: 60.0  # max seconds a statement may run; ~ = unbounded (asyncpg only, default: 60)
        connect_timeout: 10.0  # max seconds to establish a connection; ~ = wait forever (asyncpg only, default: 10)
```

Checked rather than assumed: each sample still parses, its `profile_storage` config still validates against `SqlAlchemyConf`, and the `~ = unbounded` in the comment resolves to null and drops `connect_args` entirely, so the documented escape hatch works from the sample as written.

malatewang (approved):

PR #1682, opened 2026-09-17
Hand the Qdrant vector store a metrics factory, so its tracker emits (port of #1532)

edwinyyyu:
### Port

Copy of #1532, merged into `speedkick` as `b6c90ab40`, on `main`: the same commit cherry-picked, keeping its author, wanghy73, with two differences: the Dockerfile change (an `EXTRAS` build arg) is left out as unrelated to the wiring, and the `metrics_factory_id` key is added to the database configuration table in the docs. A second commit rewords that row to name Neo4j alongside Qdrant, since `Neo4jConf` takes the same key on `main` (#1678) and has no row of its own. The other three stores #1532 names, Neo4j, episode and session, were wired on `main` by #1678 (the port of #1523); the Qdrant store is the one left.

### Purpose of the change

`OperationTracker` accepts `factory=None` and then discards every timing without an error, a warning, or a series. A component that is fully instrumented but never handed a factory looks identical to one that was never instrumented; the only way to notice is to go looking for a metric that should exist and find nothing.

That is what `QdrantVectorStore` was doing: `QdrantVectorStoreParams.metrics_factory` defaults to `None` and nothing set it. `QdrantConf` gains `MetricsFactoryIdMixin` so it can resolve one, and `database_manager` passes it through.

`test_qdrant_creates_vector_store` pinned the exact `QdrantVectorStoreParams` kwargs and had to change. It now asserts `metrics_factory is not None` rather than pinning a value: passing the keyword is not the property worth guarding, since `None` is accepted and silently discards everything. Removing the wiring fails it.

### Stack

| # | PR | change |
|---|---|---|
| — | #1682 **(this PR)** | Hand the Qdrant vector store a metrics factory, so its tracker emits (port of #1532) |

| # | PR | change |
|---|---|---|

This PR is two commits, `f51702f15` (the port) and `1a0dc145a` (the docs-row wording), directly on `main`, independent of the others: review and merge it in any order among the independent PRs (a later one touching the same file rebases). The stages' history carries the same change from [vector store scale-out 3/5] up, so their diffs include it until it merges.

### Verification

At `f51702f15`, on 2026-09-21: `ruff check` and `ruff format --check` clean; `ty check` clean as CI runs it (`uv run --frozen --all-extras ty check --project packages/server`); the full server suite without integration tests passes (`pytest packages/server/server_tests -m "not integration"`), 1914 tests at `f51702f15`. `1a0dc145a` changes one line of `docs/open_source/configuration.mdx` and nothing else.


edwinyyyu (commented):
1. The commit trailer "(cherry picked from commit b6c90ab4070e2a5f3fb1d6c4fa2b7a7a4fd2e4ad)" names a sha that exists neither locally nor on GitHub; #1532's merge commit is b6c90ab400bc0dd92105b8b94c4205cf66459f61 (same 9-character prefix).
2. The port also adds `MetricsFactoryIdMixin` to database_conf.py's import block (speedkick had it via #1523); the adaptation list omits it.

marvinyu-memverge (approved):
Approve at 1a0dc145. No asks; one observation.

Observation, not an ask: MilvusConf carries no MetricsFactoryIdMixin on main or on speedkick, and database_manager passes nothing to MilvusVectorStoreParams.metrics_factory (the field exists), so the Milvus store is the fifth component in the shape #1532 describes. Out of this PR's scope; flagging so it is on the list.

Verified: diff-of-diffs of f51702f1 against b6c90ab4 differs only in the stated omission (the Dockerfile ARG EXTRAS block) and the stated addition (the docs row), plus the import line your self-review already named; 1a0dc145 on top is the one docs line naming Neo4j on that row, which matches Neo4jConf on main. The wiring mirrors the Neo4j precedent at database_manager.py:267; QdrantVectorStoreParams.metrics_factory and the OperationTracker(params.metrics_factory, prefix="vector_store_qdrant") at :637 were already on main, so the one kwarg at :611 is the whole fix. get_metrics_factory() resolves the class-level "prometheus" factory every wired component shares, and it dedupes histograms by name, so two Qdrant confs in one process share one vector_store_qdrant_latency_seconds series with no duplicate registration. The test sees the real factory (QdrantVectorStore is patched, so no tracker is built in the unit test) and fails on KeyError if the kwarg is removed. CI green at head, no threads.

malatewang (approved):

PR #1687, opened 2026-09-17
Context parts, block kinds and kind-keyed segmenter and deriver tables (port of #1611)

edwinyyyu:
## Source and review

- Source: #1611 on `speedkick`, closed in favor of this PR; its tip f861a75fd is the content ported here.
- Review so far: Iterated with Edwin in session on speedkick (naming, the `Context` class, the kind tables, block typing); #1611's GitHub thread carries one review, today's comment on the passthrough module; not reviewed by the team.
- Adaptation for main: None for the four ported commits, rebased clean; a fifth commit removes the passthrough segmenter (below).

Port of #1611 to `main`, stacked on #1685 (the port of #1617) as #1611 is stacked on #1617 on speedkick: four commits rebased clean, and one more. Content, design and tests are described on #1611, with these corrections to its body:

- "PassthroughSegmenter is gone" was false on both branches: the kind-keyed tables removed its config, its service-locator branch and its test, and the module survived with no consumer. The fifth commit here removes it and rewords the long-term memory's over-fetch comment that named it.
- `EventMemoryParams.format_options` is not removed here; it was already gone at this branch's parent (#1597's history).
- Rendering takes `datetime_format: DateTimeFormat`, not `options`.
- Decision 6, filter translation in `event_memory/utils.py`, belongs to #1597's history; that module does not exist at this head.
- A behavior change #1611 does not state: `_episode_to_event` attaches `Context(Author(name=producer_id))` to every episode, where before only message episodes carried a producer context and the rest a null one, so a non-message episode now renders with the `producer: ` prefix too.

#1632's port follows on this branch.

Verified on this tree: 362 episodic memory tests, ruff, and ty clean on `packages/server` as CI runs it; the segmenter and long-term memory suites again on the fifth commit.


edwinyyyu (commented):
`packages/server/src/memmachine_server/episodic_memory/event_memory/segmenter/passthrough_segmenter.py` survives this change with no consumer: the diff removes `PassthroughSegmenterConf`, the `type: passthrough` branch in service_locator.py and test_passthrough_segmenter.py, but not the module, so `git grep PassthroughSegmenter` at this head hits only its own definition. The description says "`PassthroughSegmenter` is gone"; the module should go with it. long_term_memory.py (`_episode_to_event` neighborhood) also still says "under non-passthrough segmenters". Same finding on the origin, #1611.

PR #1684, opened 2026-09-17
Add session, source and expansion to EventMemory (port of #1597)

edwinyyyu:
## Source and review

- Source: #1597 on `speedkick`, closed in favor of this PR; its tip 80f0c713f is the content ported here.
- Review so far: Reviewed manually by Edwin on speedkick over many iterations (14 reviews and 13 inline comments on #1597, 66 commits of review rounds); not reviewed by the team.
- Adaptation for main: None. #1523 merged to main as #1678 before this base, so `LongTermMemory` passes the metrics factory into `EventMemoryParams` exactly as #1597 does; an earlier base without it had this port drop that line.

Port of #1597 to `main`: the same content as #1597's tip 80f0c713f, squashed into one commit and applied onto #1670's tip 4d93821b8 (the port of #1606, on `upstream/main` 4ad28ded3), so until #1663 and #1670 merge the diff shows their commits too.

A second commit removes a comment left over from the handoff branch's first commit: a line above `EventMemoryParams` naming a rendering model this branch does not have (review comment on this PR and on #1597).

Design and rationale: #1597's body and `design/event_memory_handoff.md` on #1579, with these corrections to #1597's body, which names some things by their pre-review names:

- The store writes `block_kind` from `segment.block.block_type`; there is no `.kind` attribute on a block at this head (the kind model is #1611's, ported as #1687).
- `seed_cosine_similarities` became `cosine_similarity_by_seed_uuid`; `system_predicates` is module-private, `_system_predicates`; `property_keys.py` is not taken as-is, its `validate_caller_property_key` is `validate_user_property_key` with the docstrings rewritten.
- "Nothing is renamed" does not hold: `FormatOptions` is `DateTimeFormat`, `get_segment_contexts` is `get_segments` and `get_segment_neighborhoods`, `ScoredSegmentContext` and `QueryResult` are `QueryHit`, `string_from_segment_context(s)` is `render_segments`, and `EVENT_BACKEND_SYSTEM_FIELDS` is module-private and no longer exported.
- Its verification dates (review rounds 09-11 to 09-14) predate the last 19 commits (09-15 and 09-16); the content described reflects them.

Dependent ports follow the same way: #1659 (write transaction), #1617 (eviction), #1611 (blocks), #1632 (render ids), stacked on this branch: #1686 directly, then #1685, #1687 and #1688 on top of #1686.

[...]

PR #1690, opened 2026-09-17
[coding agents 1/3] Serve the v1 event-memory API: tenants, query, expand, events

edwinyyyu:
Slice 1 of `design/coding_agent_integration.md` (#1579): the tenant-scoped v1 HTTP API over the event memory, with the bodies `design/server_redesign.md` ("Server API") specifies, on today's server. Stacked on #1688 (the main port of the event-memory stack, with the write transaction #1686 under it); its diff shows that stack until it merges. It goes through neither the v2 router nor `Episode`.

One verb from `EventMemory.query` up: the operation is a query, its route `query`, its hits query hits; "search" names only the vector store's technique.

## Routes (prefix `/v1`)

- `PUT /tenants/{tenant}` 201 on creation, 200 when it existed; `GET` 200 / 404; `DELETE` 204, idempotent. A tenant is an opaque name, at most 255 bytes.
- `POST /tenants/{tenant}/episodic-memory/query`: `query`, `limit`, `min_cosine_similarity`, `expand_context`, `rerank` (`candidates`, `min_score`, or null), `since`, `until`, `session_ids`, `source_ids`, `block_kinds`, `filter`, `datetime_format`, `parts`. Hits are `score`, `seed`, `segments` (`uuid`, `event_uuid`, `index`, `offset`, `timestamp` with offset, `session_id`, `source_id`, `context`, `block`, `properties`) and `text`, the window rendered with `render_segments(ids=("session", "segment"))`, so every hit carries the `[session:"..."]` and `[segment:<hex>]` markers a client continues from.
- `POST .../episodic-memory/expand`: `anchor`, `before`, `after`, the filters, `datetime_format`, `parts`; answers `before` and `after` as segment lists plus `before_text` and `after_text`.
- `POST /tenants/{tenant}/events`: a list of events (`id` optional, minted if absent; `timestamp` optional, server time if absent; `session_id` required; `source_id`; `context`; `blocks`; `properties`) → `encode_events`, 200 `{stored: [...]}`. Ingest is synchronous, so `?wait=` is accepted and ignored.
- `POST .../events/delete`: `{ids}` → `forget_events`.

Errors answer `{"error": {"code", "message"}}` on the v1 routes only: `tenant_not_found` 404, `component_not_enabled` 404, `event_exists` 409, `invalid_request` 422, `internal` 500 with the traceback logged and never returned. The 409 is the last commit: it adds the code to `ErrorCode` and maps `EventMemoryStoreEventAlreadyStoredError` from #1686 onto it, with a route test; a review comment on this PR describes the tree before that commit.

## The seam

`TenantEventMemories` (`create`, `exists`, `delete`, `resolve -> event memory, reranker, defaults`) is all a route's logic depends on; the router's default dependency constructs the one implementation, `EpisodicSessionTenantEventMemories`, which makes a tenant one episodic-memory session under the key `v1/tenants/<name>`, created with the event backend and no short-term memory, every resource the deployment's. It cannot collide with the v2 API: a v2 session key is `<org>/<project>` with ids that cannot contain a slash, so it holds exactly one, and a v1 key holds at least two. Delete drops the session and with it the partition and the collection. `LongTermMemory` gains two accessors, `event_memory` and `reranker`, with no episode translation on this path.

## Departures from the redesign, stated in the route descriptions

- Tenant by name, not UUID (the design's decision).
- No `position` on segments and no watermark route: there is no event store yet. Segments carry `event_uuid` where the design writes `event_id`.
- A batch naming an event the tenant already holds is rejected whole with 409 `event_exists`, naming the uuids; ids are the client's dedupe key.
- `session_id` is required on an event where the design has it optional: `EventMemory` requires one, and this path supplies no default the way the episode translation does.
- `expand` takes a segment uuid as its anchor, not a segment or event uuid, and answers `before_text` and `after_text` rather than one `text` per side.
- `filter` is the existing string grammar; no JSON-tree parser exists on this tree and none was written.
- `rerank.reranker` is rejected with `invalid_request` rather than ignored: naming a reranker would put the resource manager into a route, and the tenant's reranker is the deployment's.
- Defaults live in one place, `EpisodicMemoryDefaults` (query limit 10, expansion 2, rerank candidates 40, expand 5 before and 5 after); no episodic-memory setting existed for them.

## Choices to review

An anchor the tenant does not hold is 422 `invalid_request` (the closed set has no segment-level 404); `events/delete` answers 200 with a JSON `null` body, since the route returns nothing and the redesign's head position does not exist here, where the tenant delete answers 204; a delete racing a resolve that still holds the manager's reference surfaces as 500 rather than a 409 outside the closed set.

## Verified

`docs/openapi.json` regenerated by `docs/tools/generate_openapi.py`, which now loads the v1 router as well: five paths and their schemas added, nothing of v2 changed. 691 server and episodic-memory tests, 59 of them new (38 on the routes through the test client over the real `EventMemory` on the in-memory fakes, 21 on the resolver), ruff, and ty clean on `packages/server` as CI runs it.

### Stack: coding agents

Slices of `design/coding_agent_integration.md` (#1579), in dependency order. Prerequisites in the event-memory line: the port stack #1684 → #1686 → #1685 → #1687 → #1688 under 1, and #1692 (the block kinds capture writes: `thinking`, `tool_call`, `tool_result`, `injected`) between 1 and 2. 3 shares no code with the server, so its branch is on main, but it depends on 2 and #1692 functionally: it configures the endpoint 2 serves and writes the kinds #1692 registers, so it merges after both.

| Slice | PR | Base | Content |
| --- | --- | --- | --- |
| 1/3 | #1690 | #1688 | the v1 event-memory API: tenants, query, expand, events |
| 2/3 | #1691 | #1692 | `memory_query` and `memory_expand` served at `/v1/mcp` |
| 3/3 | #1693 | main | Claude Code and Codex in the client: the MCP entry, the `Stop` hook, and capture |


packages/server/src/memmachine_server/server/api_v1/router.py
```diff
@@ -0,0 +1,393 @@
...
+    )
+
+
+ADD_EVENTS_DESCRIPTION: Final[str] = """
+Remember a batch of events.
+
+An event's `id` is the client's dedupe key: a client that retries a
+batch supplies the ids it used, and a batch without them is stored
+again. A tenant holds an event once: a batch naming an id the tenant
+already holds is rejected whole with `event_exists`, and nothing in it
```
edwinyyyu, line 332: `ErrorCode` (errors.py) is the closed set `tenant_not_found | component_not_enabled | invalid_request | internal`, and #1686's `EventMemoryStoreEventAlreadyStoredError` is a plain `Exception`, so once #1686 is absorbed a repeated id is answered 500 `internal` by the `except Exception` arm, not 409 `event_exists`. Either the code and the mapping land here (or with #1686's absorption), or this text and the PR body should say 500.

edwinyyyu (commented):
1. The 409 `event_exists` promise (line comment) has no path to be kept by the PRs named.
2. `events/delete` returns `None` at 200, which FastAPI serializes as the JSON literal `null`, and the OpenAPI 200 response advertises `application/json`; the description says "no body". 204 would make that true.

packages/server/src/memmachine_server/server/api_v1/router.py
edwinyyyu: This describes the branch before its last commit, "Answer a batch naming a held event with 409 event_exists" (7c43319c5 after today's rebase): it adds `event_exists` to `ErrorCode`, maps `EventMemoryStoreEventAlreadyStoredError` onto 409 in the envelope route, and tests the route. The body now names the commit.

PR #1691, opened 2026-09-17
[coding agents 2/3] Serve memory_query and memory_expand over MCP at /v1/mcp

edwinyyyu:
Slice 2 of `design/coding_agent_integration.md` (#1579): the two recall tools a coding agent calls, served by the server over MCP at `/v1/mcp`. Stacked on #1692 (the block kinds), over #1690 (the v1 API), whose seam it uses; its diff shows #1690 and the port stack below it until they merge.

## What it serves

A second FastMCP app, mounted at `/v1/mcp` beside the untouched v2 app at `/mcp`, with the tenant read from the `X-MemMachine-Tenant` header into a context variable of its own, so the two apps' contexts never mix. Tenants resolve through the same `EpisodicSessionTenantEventMemories` the routes use, against the one `MemMachine` instance the process holds, so a tenant reached through a tool and through a route is one memory. Both apps' lifespans are chained by the application's, since a mounted app's lifespan is not run by its mounter.

Tools, with no count or size parameter on either, as the design requires:

- `memory_query(cue, within=None, kinds=None, since=None, until=None)`: the tenant's `EventMemory.query` with the tenant's defaults (`query_limit`, `expand_context`), reranked through the tenant's reranker over `rerank_candidates` when it has one, each hit's window rendered with the session and segment markers, best first, one blank line between hits. `within` is a session id, `kinds` block kinds, `since` and `until` ISO 8601 timestamps that must carry an offset.
- `memory_expand(id, direction="around")`: `id` is `[segment:<hex32>]` or the bare uuid; a `[segments:a..b]` run is refused with the instruction to give one end. The step is `expand_before + expand_after` from the defaults: `around` spends a quarter backward and the rest forward, `earlier` and `later` all of it one way. Each side renders with the markers, so its edges are the next handles; a side that ran out says so in one line.

The descriptions are the only prompting: what a good cue is (the context a memory was encoded in, the user's own wording, query the surroundings and expand when the target cannot be pinned, a lead is another query), how to read the markers, and that a window is an append read once. The rendered timestamp format is one constant, `medium` date and `long` time in the zone the timestamp carries.

Errors (no header, unknown tenant, no event backend, a bad id or timestamp, an anchor the tenant does not hold) are tool errors with a plain message.

## Verified

28 tests through fastmcp's in-memory client with the v1 fake resolver: the tool surface (names, descriptions, schema properties exactly the five and two parameters), query rendering and ordering under `AngleEmbedder`, every filter reaching the memory, reranking, the three directions with their counts asserted against the defaults, every id form, every error, and one end-to-end streamable-HTTP handshake through the mounted app that reads the header. 352 server tests, ruff, and ty clean on `packages/server` as CI runs it. `docs/openapi.json` is unaffected: the generator builds from the routers only, and regenerating it compares equal.

## Choices to review

- `memory_expand`'s wire parameter is `id` through a pydantic alias, since ruff forbids a Python parameter named `id` and the repo forbids per-file ignores.
- fastmcp's default stateful transport is kept, as in the v2 app; the header is read on the initialize request, which is how a static agent config sends it.
- The tenant is resolved before arguments are parsed, so a bad argument sent to an unknown tenant reports the tenant.

### Stack: coding agents

Slices of `design/coding_agent_integration.md` (#1579), in dependency order. Prerequisites in the event-memory line: the port stack #1684 → #1686 → #1685 → #1687 → #1688 under 1, and #1692 (the block kinds capture writes: `thinking`, `tool_call`, `tool_result`, `injected`) between 1 and 2. 3 shares no code with the server, so its branch is on main, but it depends on 2 and #1692 functionally: it configures the endpoint 2 serves and writes the kinds #1692 registers, so it merges after both.

| Slice | PR | Base | Content |
| --- | --- | --- | --- |
| 1/3 | #1690 | #1688 | the v1 event-memory API: tenants, query, expand, events |
| 2/3 | #1691 | #1692 | `memory_query` and `memory_expand` served at `/v1/mcp` |
| 3/3 | #1693 | main | Claude Code and Codex in the client: the MCP entry, the `Stop` hook, and capture |


packages/server/server_tests/memmachine_server/server/api_v1/test_mcp.py
```diff
@@ -0,0 +1,546 @@
+"""Tests for the v1 MCP tools: what they render, and what they refuse."""
+
+import re
+from datetime import UTC, datetime, timedelta
+from uuid import uuid4
+
+import httpx2
```
edwinyyyu, line 7: `httpx2` is declared by no pyproject in the workspace; it reaches this test only as fastmcp-slim's dependency, so a fastmcp-slim bump that drops it breaks the test rather than an import in src.

edwinyyyu (commented):
One dependency note (line comment).

packages/server/server_tests/memmachine_server/server/api_v1/test_mcp.py
edwinyyyu: Declared in 1220ad4bf: `httpx2>=2.0.0` in the root dev group, the floor being the first release with the ASGI transport, the client, Timeout and Auth the test names (checked in the 2.0.0 wheel); locked with uv 0.12.15, the release CI installs, so the lock gains one line.

PR #1698, opened 2026-09-18
Add Qdrant to docker-compose and the Helm chart, and upgrade PostgreSQL to 18

wanghy73:
## Purpose of the change

Two deployment gaps, plus a version bump.

**Qdrant had no container anywhere.** The sample configs call it *"the recommended default for server deployments (concurrent access, scaling)"* and then tell the reader to start it by hand:

```
docker run -p 6333:6333 -p 6334:6334 qdrant/qdrant
```

which leaves it off `memmachine-network` and outside the cluster, so the app cannot reach it by service name. Postgres and Neo4j are proper services in both compose and the Helm chart; Qdrant was in neither. The config schema has supported it all along (`QdrantConf`) — only the deployment plumbing was missing.

**PostgreSQL moves 16 → 18.**

## Description

**1. `docker-compose.yml`** — a `qdrant` service pinned to `v1.19.1` on `memmachine-network`, ports 6333 (REST) / 6334 (gRPC), a named volume at `/qdrant/storage`, a healthcheck, `QDRANT_HOST`/`QDRANT_PORT`/`QDRANT_GRPC_PORT` for the app, a `depends_on` gate, and the matching block in `sample_configs/env.dockercompose`.

The healthcheck is an explicit `bash -c` rather than the usual `CMD-SHELL` + curl. The image carries `bash` but no `curl`, `wget` or `nc`, and its `/bin/sh` is dash, which has no `/dev/tcp` — so the normal idiom cannot work here.

**2. Helm chart** — `qdrant-deployment.yaml` and `qdrant-service.yaml` (both gated on `qdrant.enabled`), `qdrant-pvc` at `/qdrant/storage`, a `wait-for-qdrant` initContainer beside the existing two, an `event_vector_store` entry under `resources.databases` in the ConfigMap, `qdrant.*` values, and the README's component / PVC / template / values tables. Probes are `tcpSocket`, matching neo4j and postgres.

The ConfigMap entry is deliberately **not** gated on `qdrant.enabled`, matching `db_neo4j`. `enabled: false` means *"skip the in-cluster objects and point at an external host"*, so gating the config too would have broken exactly that case. I had it wrong first and caught it by rendering with `enabled=false`.

**3. PostgreSQL 16 → 18** — `pgvector/pgvector:pg18` (PostgreSQL 18.6, pgvector 0.8.6, up from 16.15).

The tag bump alone is not enough. PG18 images store data in a major-version subdirectory, so the mount point has to move with it:

| | pg16 | pg18 |
|---|---|---|
| `PGDATA` | `/var/lib/postgresql/data` | `/var/lib/postgresql/18/docker` |
| declared `VOLUME` | `/var/lib/postgresql/data` | `/var/lib/postgresql` |

Both the compose volume and the chart's `postgres-pvc` mountPath are updated.

## Type of change

## How Has This Been Tested?

- [x] Manual verification

**Qdrant, compose** — against v1.19.1:

| Check | Result |
|---|---|
| `docker compose config` | valid |
| `docker compose up -d qdrant` | healthy in ~9s |
| Healthcheck on the live port | exit `0` |
| Healthcheck on a dead port | exit `1` — it genuinely fails |
| Create collection → upsert → search → delete | all succeed, `score: 1.0` |
| Collection survives `compose restart` | volume wired correctly |
| `http://qdrant:6333` from another container on the network | resolves and responds |

The negative test mattered: a healthcheck that always returns 0 would be worse than none.

**Qdrant, Helm** — `helm lint` clean, and `kubeconform -strict` against Kubernetes 1.29 schemas:

```
default            -> 17 resources, Valid: 17, Invalid: 0
qdrant.enabled=false -> 14 resources, Valid: 14, Invalid: 0
external host      -> 14 resources, Valid: 14, Invalid: 0
apiKey + https     -> 17 resources, Valid: 17, Invalid: 0
all deps disabled  ->  8 resources, Valid:  8, Invalid: 0
```

Exactly three objects drop out when Qdrant is disabled, and nothing else changes. The rendered `event_vector_store` block was also fed through the real `QdrantConf` model for the default, `apiKey`, `prefer_grpc` and external-host renders — valid in each. `api_key` is emitted only when `qdrant.apiKey` is set.

**PostgreSQL 18:**

- compose: healthy in ~9s; `server_version` 18.6, `data_directory` `/var/lib/postgresql/18/docker`, pgvector 0.8.6.
- A table with a `vector` column survives `compose up --force-recreate`, so the moved mount really does persist.
- Checked what happens if the mount is *not* moved, rather than assuming: pg18 **refuses to start** and prints an explanation, even on an empty volume. It does not start up and quietly write to an unmounted path, so a missed mount cannot silently lose data.

## Further comments

**On the PostgreSQL upgrade and existing data.** A PG16 data directory cannot be read by PG18 — normal for any PostgreSQL major upgrade. I simulated it: populate a PG16 volume, then point pg18 at it, and the container exits with the layout error instead of starting, in both the old and new mount positions. Nothing is lost or overwritten, but nothing starts either until the data is migrated.

This was raised and the answer was that there is no existing data to preserve — these are dev deployments — so no migration is needed here. The dump/restore commands are documented in the chart README for anyone who does have data.

**Scope.** This adds the container and makes it addressable. It does not switch anything over: using Qdrant still means pointing `configuration.yml` at `backend: event` with `vector_store` wired to the qdrant entry. Neo4j serves the declarative backend, Qdrant the event one.

**Not verified.** No Kubernetes cluster was reachable from here, so the Helm side is render and schema validation, not a deployed rollout.


edwinyyyu (approved):

wanghy73:
Good catch — you are right, and the PR as it stood had no answer. It pinned `episodic_memory.long_term_memory.vector_graph_store: db_neo4j` and deployed Qdrant beside it, so Qdrant ran **unused** and switching meant hand-editing the rendered config. Compose started both too. Fixed in 60e46a6.

The server treats the two as a discriminated union on `backend` — `DeclarativeLongTermMemoryConf` takes `vector_graph_store`, `EventLongTermMemoryConf` takes `vector_store` + `segment_store` — so exactly one applies. The deployment now follows that.

**Helm** — one value picks the backend, and only the store it needs is deployed:

```bash
helm upgrade --install memmachine . --set episodicMemory.longTermMemory.backend=event
```

| `backend` | Store deployed | `long_term_memory` wiring |
|---|---|---|
| `declarative` (default) | Neo4j | `vector_graph_store: db_neo4j` |
| `event` | Qdrant | `vector_store: event_vector_store`, `segment_store: db_postgres` |

The unused store’s Deployment, Service and PVC are skipped even if its `enabled` flag is left `true`. `neo4j.enabled` / `qdrant.enabled` keep their existing meaning — in-cluster vs external host — and only apply to the store the backend actually uses. The `wait-for-*` initContainers are gated on the backend alone, since an external host still needs waiting for.

**Compose** — the two stores are now in `declarative` and `event` profiles:

```bash
docker compose --profile event up      # qdrant, no neo4j
```

The env sample sets `COMPOSE_PROFILES=declarative`, so the default stays exactly what it is today.

Verified:

- Helm renders exactly **one** store Deployment per backend, with the matching initContainer.
- Both rendered `configuration.yml` files load through the real `Configuration.load_yml_file` — declarative resolves to `db_neo4j`, event to `event_vector_store` + `db_postgres`.
- `helm lint` clean; `kubeconform -strict` on Kubernetes 1.29 valid for declarative, event, and each with its store external.
- Brought the event profile up for real: postgres and qdrant healthy, **no neo4j container started**.

One thing worth recording, since it is not obvious: compose profiles alone were not enough. A plain `depends_on` on a profiled service that is not active fails the whole project with `depends on undefined service`. The `required: false` on those two `depends_on` entries is what makes the switch work.

deployments/helm/values.yaml
```diff
@@ -9,13 +9,24 @@
...
 
+# Long-term memory backend. This picks which vector store is deployed and wired:
+#   declarative -> Neo4j (vector_graph_store). The legacy backend.
+#   event       -> Qdrant (vector_store) + Postgres (segment_store).
+# The two are mutually exclusive in the server config, so only the store the
+# chosen backend needs is deployed; the other is skipped even if its
+# `enabled` flag is true.
+episodicMemory:
+  longTermMemory:
+    backend: declarative     # declarative | event
```
malatewang: As we discussed, switch event memory as default.

deployments/helm/templates/memmachine-configmaps.yaml
```diff
@@ -31,7 +31,14 @@ data:
...
         embedder: default_embedder
         reranker: my_reranker_id
+{{- if eq .Values.episodicMemory.longTermMemory.backend "event" }}
+        backend: event
+        vector_store: event_vector_store
+        segment_store: db_postgres
+{{- else }}
+        backend: declarative
         vector_graph_store: db_neo4j
+{{- end }}
```
malatewang, line 42: Check invalid value here?

deployments/helm/templates/qdrant-service.yaml
```diff
@@ -0,0 +1,23 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: memmachine-qdrant
+  namespace: {{ .Release.Namespace }}
+spec:
+  selector:
+    app: qdrant
+  ports:
+    - name: rest
```
malatewang, line 17: use the port number from values.yaml

deployments/helm/values.yaml
```diff
@@ -61,6 +72,22 @@ postgres:
...
 
+qdrant:
+  enabled: true              # false = skip in-cluster Deployment/Service/PVC, use external host
+  host: memmachine-qdrant    # internal service name (default); override when enabled: false
+  port: 6333                 # REST port
+  grpcPort: 6334             # gRPC port
+  image: qdrant/qdrant:v1.19.1
+  prefer_grpc: false         # true = talk gRPC (grpcPort) instead of REST
+  https: false               # true when fronted by TLS, or for Qdrant Cloud
+  # apiKey: <QDRANT_API_KEY> # required for Qdrant Cloud; omit for in-cluster
```
malatewang: Should move the api key out of yaml?

deployments/helm/values.yaml
wanghy73: Done in 46b3b01 — `backend` now defaults to `event`.

One thing to be aware of: releases installed before chart 0.2.0 ran Neo4j, so a plain `helm upgrade` would now switch them to `event` and Helm would remove the Neo4j Deployment, Service and `neo4j-pvc`. The README calls this out and says to pin `--set episodicMemory.longTermMemory.backend=declarative` when upgrading. Docker compose still defaults to `declarative`; happy to switch that too if we want them aligned.

deployments/helm/templates/memmachine-configmaps.yaml
wanghy73: Done in 46b3b01. Every template now reads the backend through a `memmachine.ltmBackend` helper (`templates/_helpers.tpl`) that fails the render on anything other than `declarative` or `event`, e.g.:

```
episodicMemory.longTermMemory.backend must be "declarative" or "event", got "Event"
```

The `else` here is also now an explicit `else if eq $backend "declarative"`, so nothing falls through silently.

deployments/helm/templates/qdrant-service.yaml
wanghy73: Done in 46b3b01. The Service ports are `qdrant.port` / `qdrant.grpcPort`, and they target the container's named ports (`rest` / `grpc`), so changing the values moves the Service port without touching what Qdrant listens on. Rendered with 7000/7001: the Service, the configmap and the `wait-for-qdrant` initContainer all pick up the new ports.

deployments/helm/values.yaml
wanghy73: Done in 46b3b01. The key no longer lands in the ConfigMap — `configuration.yml` renders `api_key: $QDRANT_API_KEY`, which the server resolves from the env var. The key comes from a Secret:

- `qdrant.existingSecret`: a Secret you create out-of-band with a `QDRANT_API_KEY` key (preferred; wins if both are set), or
- `qdrant.apiKey`: the chart creates `qdrant-secret` from it.

With either set, the in-cluster Qdrant also enforces the key via `QDRANT__SERVICE__API_KEY`, so it isn't left open. Verified the rendered config loads through `Configuration.load_yml_file` with the key resolved from the env.

deployments/helm/values.yaml
wanghy73: Follow-up: docker compose now defaults to `event` too (b71b8d9), so the two deployments match.

- `COMPOSE_PROFILES=event` in `sample_configs/env.dockercompose`, so a default `docker compose up` runs Postgres + Qdrant and no Neo4j.
- `memmachine-compose.sh` generates a `configuration.yml` wired for the selected profile (`vector_store` + `segment_store` for event), and points the Qdrant host at the `qdrant` service.
- An existing `.env` without `COMPOSE_PROFILES` would have started no vector store at all; the script now adds one, inferring `declarative` from an existing config that uses Neo4j, so current setups keep their backend.

Note: I rewrote the last two commits to add the DCO sign-off, so the Helm changes are now 8beccba (was 46b3b01).

wanghy73:
@malatewang @edwinyyyu could you take another look?

@malatewang, all four of your comments are addressed, with a reply on each thread:

| Comment | Commit |
|---|---|
| Make event memory the default | 8beccba (Helm), b71b8d9 (compose, so the two match) |
| Check for an invalid backend value | 8beccba: a `memmachine.ltmBackend` helper fails the render on anything but `declarative` / `event` |
| Take the Qdrant ports from `values.yaml` | 8beccba |
| Keep the API key out of the YAML | 8beccba: read from a Secret, via `qdrant.existingSecret` |

The thread replies cite 46b3b01. That is the same commit before a rebase, and it is 8beccba on the branch now.

@edwinyyyu, a lot has changed since your approval. 60e46a6 deploys only the vector store the chosen backend uses, and 8beccba and b71b8d9 make `event` the default in Helm and in compose. One change needs a look on upgrade: releases installed before chart 0.2.0 must pin `backend: declarative`. Otherwise Helm removes the Neo4j Deployment, Service and `neo4j-pvc`. The README covers this.

All 44 checks pass, and the branch merges cleanly.

malatewang (approved):

PR #1707, opened 2026-09-22
feat: persist UUID episode identifiers

malatewang:
Assign and persist UUIDs for episode entries while preserving batch and semantic ingestion chronology.

### Purpose of the change

In current implementation, episode acquires database generated auto increment UUID. when adding a new episode, the episode must be saved to the episode store first to get the id, then episode is added to semantic and episodic memory sequentially. This increases the latency. This PR uses python to generate uuid for episode and episodes are added to the episode store, episodic memory and semantic memory in parrallel. 
This change breaks the episode store schema. It is not backward compatible.

### Description

Please include a summary of the change and the issue it addresses. Please also include relevant motivation and context. List any dependencies that are required for this change.

### Fixes/Closes

Fixes #(issue number)

### Type of change

- [ ] Documentation update

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test
- [ ] End-to-end Test

### Checklist

### Maintainer Checklist

- [ ] Reviewed the code

### Screenshots/Gifs

### Further comments


marvinyu-memverge (commented):
Reviewed at 10e0ae25. The parallel write and the grace period on missing
episodes both look right to me; three things I'd want settled before this
goes in, all reproduced locally:

1. Upgrading an existing deployment. The episode store only runs create_all,
   so an existing episodestore keeps its INTEGER id column, and inserting a
   UUID into it fails (SQLite: "IntegrityError: datatype mismatch"; Postgres
   rejects it the same way). Because the store write now runs alongside the
   episodic and semantic writes, each of those failed adds still lands in
   episodic memory and the semantic queue. So on an un-migrated database
   every add errors *and* leaves an orphan. Could we either ship a migration
   (semantic storage already has one for its history_id -> string change) or
   check the column type at startup and refuse to start? Either way it fails
   loudly instead of half-writing.

2. Orphans can't be deleted by id. Same shape outside the upgrade case: if
   episode_storage.add_episodes fails (DB timeout, dropped connection), the
   episode is still in episodic memory and shows up in search, but the new
   existence check in delete_episodes raises ResourceNotFoundError for it,
   so it can only be cleared by deleting the whole session. Either
   delete_episodes should go ahead for ids that aren't in the store, or a
   failed store write should undo the episodic write.

3. Ordering within a batch. History now sorts by (created_at, history_id),
   and every message in one add shares a created_at whenever the client sends
   the same timestamp for all of them (e.g. importing a transcript stamped
   per session). Ties then fall back to the random UUID, so ingestion can
   process the messages out of order and the older fact wins. I ran
   test_ingestion_keeps_latest_value_across_uuid_ordered_batches with every
   entry at the same created_at: it ends on "violet" instead of "red". On
   main the int id kept insertion order. Something that breaks ties by
   position in the batch would restore it.

Verified and fine: created_at reaches semantic history on all four backends,
the 30 s grace period covers ingestion reading an episode before its store
row commits, and CI is green at head.

malatewang:
> Reviewed at [10e0ae2](https://github.com/MemMachine/MemMachine/commit/10e0ae2588cc988b9de16094fc5d70f8cbc162a9). The parallel write and the grace period on missing episodes both look right to me; three things I'd want settled before this goes in, all reproduced locally:
> 
> 1. Upgrading an existing deployment. The episode store only runs create_all,
>    so an existing episodestore keeps its INTEGER id column, and inserting a
>    UUID into it fails (SQLite: "IntegrityError: datatype mismatch"; Postgres
>    rejects it the same way). Because the store write now runs alongside the
>    episodic and semantic writes, each of those failed adds still lands in
>    episodic memory and the semantic queue. So on an un-migrated database
>    every add errors _and_ leaves an orphan. Could we either ship a migration
>    (semantic storage already has one for its history_id -> string change) or
>    check the column type at startup and refuse to start? Either way it fails
>    loudly instead of half-writing.
> 2. Orphans can't be deleted by id. Same shape outside the upgrade case: if
>    episode_storage.add_episodes fails (DB timeout, dropped connection), the
>    episode is still in episodic memory and shows up in search, but the new
>    existence check in delete_episodes raises ResourceNotFoundError for it,
>    so it can only be cleared by deleting the whole session. Either
>    delete_episodes should go ahead for ids that aren't in the store, or a
>    failed store write should undo the episodic write.
> 3. Ordering within a batch. History now sorts by (created_at, history_id),
>    and every message in one add shares a created_at whenever the client sends
>    the same timestamp for all of them (e.g. importing a transcript stamped
>    per session). Ties then fall back to the random UUID, so ingestion can
>    process the messages out of order and the older fact wins. I ran
>    test_ingestion_keeps_latest_value_across_uuid_ordered_batches with every
>    entry at the same created_at: it ends on "violet" instead of "red". On
>    main the int id kept insertion order. Something that breaks ties by
>    position in the batch would restore it.
> 
> Verified and fine: created_at reaches semantic history on all four backends, the 30 s grace period covers ingestion reading an episode before its store row commits, and CI is green at head.

We do not support upgrade or backward compatibility.

PR #1711, opened 2026-09-23
Delete the fixed-sleep semantic background ingestion test

edwinyyyu:
### Purpose of the change

Delete `test_background_ingestion_processes_messages_on_message_limit`. It keeps failing the semantic-memory integration job on unrelated PRs with `Expected 0 uningested messages, got 2`. Another test in the same file already covers everything it checks, and that test doesn't flake.

### Description

The test added two messages to one set, slept a fixed 0.6 s, and then asserted that both had been ingested. It isn't an ingestion bug: the pass finishes, just after the sleep ends.

On Neo4j, the first ingestion pass for a new set creates the set's vector index and blocks on `CALL db.awaitIndexes()`. Timestamped debug logs from a local run of the `neo4j_semantic_storage-sql_db_episode_storage-sqlalchemy_pg_engine` variant show:

| Event | Time |
|---|---|
| second message added | 48.912 |
| `Processing 2 messages for set user-123` | 49.240 |
| sleep ends, assertion runs (`Stopping semantic memory services`) | 49.657 |
| `Finished processing 2 messages out of 2` | 50.359 |

When the assertion ran, Neo4j reported that the `Feature` label did not exist yet. Before this change, the test failed on 3 of its 4 Neo4j variants locally.

`test_multiple_sets_processed_independently` uses the same service configuration (`feature_update_interval_sec=0.05`, `uningested_message_limit=0`). It runs the same flow: one `add_messages` call per message, two messages per set. It makes the same assertions, zero uningested messages and at least one feature, for each of two sets. It polls with a 15 s timeout instead of sleeping. In the same local run, its pass finished 0.73 s after its last message was added. A fixed 0.6 s sleep would have failed it too, but polling makes it pass.

Despite its name, the deleted test did not exercise the message limit. A limit of 0 is clamped to 1, so any message triggers ingestion. The threshold itself is covered by the `get_history_set_ids(min_uningested_messages=...)` cases in `storage/test_semantic_storage.py`.

### Type of change

### How Has This Been Tested?

- [x] Unit Test
- [x] Integration Test

**Test Results:** `test_semantic_memory_background.py` on all storage variants (`-m "integration or not integration"`, with PostgreSQL and Neo4j test containers): 72 passed. `ruff check` and `ruff format --check` are clean on the file.


marvinyu-memverge (approved):
Approving at 6bb8ad41.

Checked the three claims the deletion rests on:

- test_multiple_sets_processed_independently covers everything the deleted
  test did: same Params (feature_update_interval_sec=0.05,
  uningested_message_limit=0), same one-add_messages-per-message flow, same
  two assertions per set, and it polls with a 15 s timeout instead of a fixed
  sleep.
- The limit really is clamped (semantic_memory.py: max(limit, 1)), so the
  deleted test never exercised a threshold. The threshold cases (0/1/2) are in
  storage/test_semantic_storage.py.
- The flake is this test. Of the recent failed integration runs I looked at
  (09-18 to 09-22, seven different branches), all but one that failed on a
  test failed on this one with "Expected 0 uningested messages, got 2".

One observation, no action needed: it isn't Neo4j-only. Run 35409873276
failed it on pgvector_semantic_storage-sql_db_episode_storage-
sqlalchemy_sqlite_engine, so the fixed sleep also loses on a non-Neo4j
backend. The awaitIndexes delay is one way to miss the 0.6 s window, not the
only one. Deleting the test fixes both.

malatewang (approved):

